PR #972, opened 2014-08-19
hide pagination for screen readers

bmaddy:
This was confusing some people during user tests at PSU. They suggested we hide it.


mtribone:
Oh boy. So sorry to have missed this for so long. I'm not sure why this fell to the bottom for me, but many thanks @cbeer for closing the issue. Yeah, as others have pointed out, we should not deny access to functional items for any user. It appears that the confusion was that because "previous" was listed on page 1 when the user wasn't anywhere else in the set of results.

When viewing results, listing, etc. and it involves pagination, don't display "previous" when I'm on page 1. Make sense?

PR #982, opened 2014-08-29
Search results default to the users last used view type

jcoyne:


ebenenglish:
Is there a configuration option to disable the pull-view-params-from-the-session behavior? This kinda messes with the geographic searches in Blacklight-Maps, in that a new search from an item on the map always results in a map view, rather than a normal search view.

I'm in the process of working on a PR to fix it, but would rather not have to pass in yet another parameter via the data attributes if possible...

cbeer:
What do you mean by "a new search from an item on the map always results in a map view"?

ebenenglish:
Sorry, confusing wording. On the map views, clicking a marker results in a popup that has a link to a new search that allows the user to find more items from the same location (or that intersect the clicked-on coordinate location). That search could be for a geographic facet field string, or it could be a set of coordinates, etc.

By design, those search links don't have a 'view' parameter. Prior to BL 5.8, the resulting catalog#index view (after clicking the search) would simply use the default index view type as set in the blacklight_config.

In Blacklight >=5.8, if there's no view param specified, the most recent view from the user's session will be used. Which, in the case of a map view, will be 'map'. So the user never actually gets to see the search results, just another map, as catalog#index will render the documents using the map view type rather than whatever the default was (usually 'list').

Hopefully this makes more sense?

cbeer:
I guess I don't understand why what you describe is not the expected behavior. If I'm exploring data in one mode, I think I'd like to stay in that mode. If what you describe is really desirable, I think passing an explicit `view` parameter is the right thing to do.

ebenenglish:
I agree that maintaining the view is probably the desired default behavior in most cases. It just doesn't work well with the 'geographic data as facet' map views that blacklight-maps offers.

If there was a config option to turn off the 'use the most recent view param from the session if no value is explicitly passed in the params hash', as @mejackreed suggested above, that would make things easier.

However, it's true that an implementer might not want to change this as the default behavior for their entire site, even if it is desirable in the blacklight-maps views, so I'll just work on passing an explicit ```view``` parameter.

PR #1047, opened 2014-12-16
Refactoring to allow repositories other than Solr

jcoyne:


cbeer:
@jcoyne rebase?

lib/blacklight.rb
```diff
@@ -43,26 +48,70 @@ class << self
...
+    require 'erb'
+    require 'yaml'
+
+    return @blacklight_yml if @blacklight_yml
+    unless File.exists?(blacklight_config_file)
+      if File.exists?(solr_file)
+        Deprecation.warn Blacklight, "Configuration is now done via blacklight.yml. Suppport for solr.yml will be removed in blacklight 6.0.0"
+        return solr_yml
+      else
+        raise "You are missing a solr configuration file: #{solr_file}. Have you run \"rails generate blacklight:install\"?"
```
cbeer: This error message needs to be updated, right?

lib/railties/blacklight.rake
```diff
@@ -14,8 +14,9 @@ namespace :blacklight do
     desc "Put sample data into solr"
     task :seed do
       docs = YAML::load(File.open(File.join(Blacklight.root, 'solr', 'sample_solr_documents.yml')))
-      Blacklight.solr.add docs
-      Blacklight.solr.commit
+      conn = RSolr.connect(Blacklight.connection_config)
```
cbeer: Shouldn't we use a `Blacklight` method to make this connection?

lib/blacklight.rb
```diff
@@ -1,6 +1,7 @@
 # -*- encoding : utf-8 -*-
 require 'kaminari'
 require 'rsolr'
```
cbeer: Should we generate `gem "rsolr"` into the application Gemfile?

PR #1054, opened 2014-12-25
Run travis with ruby 2.2.0 + rails 4.2.0

cbeer:


jronallo:
It looks as if this build is passing but is very noisy with duplicated key warnings.

PR #1063, opened 2015-01-26
Allow blacklight configuration to provide multiple title field options

cbeer:
and have presenter methods select the first, present field.

Fixes sul-dlss/spotlight#901


jcoyne:
There's no test for document_show_link_field. Can you add one?

PR #1064, opened 2015-01-27
Test against blacklight_config.solr_document_model, not SolrDocument.

rwd:
If a custom Solr document model is specified, `render_document_heading` does not detect it as its first argument because it tests for a `SolrDocument` instance.

It should instead check for the class specified in `blacklight_config.solr_document_model`


jcoyne:
Thanks for the contribution @rwd. Good catch.

PR #1066, opened 2015-01-28
current_sort_field should return sort from solr response if present

cjcolvar:
When making sort and per page sicky for a session, I noticed that current_sort_field does not reflect the result of the current search like current_per_page.  This happens when the solr query includes a sort different than the request params.


app/helpers/blacklight/catalog_helper_behavior.rb
```diff
@@ -78,7 +78,7 @@ def search_field_label(params)
   #
   # @return [Blacklight::Configuration::SortField]
   def current_sort_field
-    blacklight_config.sort_fields[params[:sort]] || default_sort_field
+    (blacklight_config.sort_fields[@response.sort] if @response and @response.sort.present?) || blacklight_config.sort_fields[params[:sort]] || default_sort_field
```
cbeer: I think we need to be trickier here. sort fields are not necessarily keyed on the list of fields. I think we'll need something like e.g.:

```
blacklight_config.sort_fields.find { |f| f.sort == @response.sort }
```

cjcolvar:
@cbeer how does that look?

cbeer:
:+1:

PR #1071, opened 2015-02-03
SolrDocument.primary_key should be the same as Solr's unique key.

cbeer:
h/t @bcail


bcail:
sounds good to match unique key and primary key.

PR #1072, opened 2015-02-03
Rename get_search_results to search_results

jcoyne:
`search_results` lets you pass in the filter methods.  This is helpful
for when you have different methods in the same class (e.g. index, show)
both calling `search_results`, but with different filters to run.


lib/blacklight/request_builders.rb
```diff
@@ -45,14 +45,22 @@ module RequestBuilders
     # specified otherwise. 
     #
     # Incoming parameter :f is mapped to :fq solr parameter.
-    def solr_search_params(user_params = params || {})
+    def solr_search_params(user_params = nil, processor_chain = nil)
+      if !user_params
```
cbeer: `unless`?

lib/blacklight/solr_helper.rb
```diff
@@ -78,7 +78,17 @@ def solr_doc_params(id=nil)
...
+    Deprecation.warn(self, "get_search_results is deprecated and will be removed in blacklight-6.0. Use `search_results' instead")
+    search_results(user_params, extra_controller_params, solr_search_params_logic)
+  end
+
+  # a solr query method
+  # @param [Hash,HashWithIndifferentAccess] user_params ({}) the user provided parameters (e.g. query, facets, sort, etc)
+  # @param [Hash,HashWithIndifferentAccess] extra_controller_params ({}) extra parameters to add to the search
+  # @param [List<Symbol] processor_chain a list of filter methods to run
+  # @return [Blacklight::SolrResponse] the solr response object
+  def search_results(user_params, extra_controller_params, solr_search_parms_logic)
```
cbeer: While we're changing this, I wonder if we can do away with `extra_controller_params`.. Maybe a new type of `solr_search_params_logic` that appends the attributes?

cbeer: typo: `solr_search_parms_logic`

jcoyne: Yeah, it was tedious passing an empty hash around there.

jcoyne:
@cbeer I refactored to avoid having to pass around the extra_controller_params.  The one thing I couldn't figure out was the tests in solr_helper for grouping.  Are those still relevant?  I couldn't see how such a request could be made from Catalog. Was it your intention that it was triggered by an override?

cbeer:
An override, setting `default_params`, or having the parameter set in the solr request handler should all trigger grouping behaviors.

jcoyne:
@cbeer In that case, I think I'll just move those tests into search_builder_spec.rb

jcoyne:
@cbeer on second though, just expecting that usage to be deprecated is fine. The "correct" way would be to set default_params.

jcoyne:
@cbeer changed my mind again. I refactored the test to use the `default_params` approach.

PR #1076, opened 2015-02-12
Rename solr_repository to repository to support multiple backends

jcoyne:
This is one of the components of #1047.  That PR was too complex to rebase so I'm doing it one commit at a time.


lib/blacklight/solr_helper.rb
```diff
@@ -260,12 +262,21 @@ def grouped_key_for_results
     blacklight_config.index.group
   end
 
+  def repository_class
+    Blacklight::SolrRepository
+  end
+
+  def repository
```
cbeer: Still looking for a better name for this.

PR #1081, opened 2015-02-16
Ensure config#solr_response_model and config#solr_document_model are not...

### LINKED ISSUE #1061: Custom solr_document_model and solr_response_model classes become anonymous after deep_copy
If the intention is to allow custom classes for `solr_document_model` or `solr_response_model` to be configurable in a `configure_blacklight` block like:

```ruby
  config.solr_document_model = CustomSolrDocument
  config.solr_response_model = CustomSolrResponse
end
```

... then this behaviour is broken where deep copies of the configuration are used by, e.g. `Blacklight::SolrRespository`

When `Blacklight::Configurable#blacklight_config` creates a deep copy of the Blacklight configuration, the values of `solr_document_model` and `solr_response_model` will change from `CustomSolrDocument` and `CustomResponse` to an anonymous class:

```ruby
class CustomSolrDocument; end
CatalogController.blacklight_config.solr_document_model = CustomSolrDocument
CatalogController.blacklight_config.solr_document_model #=> CustomSolrDocument
CatalogController.blacklight_config.deep_copy.solr_document_model #=> #<Class:0x0000000553e490>
```

Observed using Blacklight 5.8.2 and Rails 4.2.0.

cbeer:
... deep-dup'ed to preserve the original classes

Fixes #1061

If those class names are dup'ed, methods that expect ActiveModel::Naming don't
behave correctly

### DIFF
--- a/lib/blacklight/configuration.rb
+++ b/lib/blacklight/configuration.rb
@@ -252,8 +252,16 @@ module Blacklight
       end
       alias_method :inheritable_copy, :deep_copy
     else
-      alias_method :deep_copy, :deep_dup
-      alias_method :inheritable_copy, :deep_dup
+      ##
+      # Rails 4.x provides `#deep_dup`, but it aggressively `#dup`'s class names
+      # too. These model names should not be `#dup`'ed or we might break ActiveModel::Naming.
+      def deep_copy
+        deep_dup.tap do |copy|
+          copy.solr_response_model = self.solr_response_model
+          copy.solr_document_model = self.solr_document_model
+        end
+      end
+      alias_method :inheritable_copy, :deep_copy
     end
 
     ##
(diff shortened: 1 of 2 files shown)


lib/blacklight/configuration.rb
```diff
@@ -252,8 +252,13 @@ def deep_copy
       end
       alias_method :inheritable_copy, :deep_copy
     else
-      alias_method :deep_copy, :deep_dup
-      alias_method :inheritable_copy, :deep_dup
+      def deep_copy
+        deep_dup.tap do |copy|
+          copy.solr_response_model = self.solr_response_model
```
jcoyne: It would be helpful to put a comment here.

cbeer: Thanks. Commented.

PR #1096, opened 2015-02-19
Rescue request exceptions from setup_previous_and_next_documents_for_search

cbeer:


jcoyne:
It's unclear to me why we'd want to swallow this exception.  Why was it raised for you in the first place?

cbeer:
See https://github.com/sul-dlss/spotlight/pull/940#discussion_r25036664.

If that query fails (but actually getting the document didn't), we don't need to break the whole page.. we just can't show the prev/next navigation.

PR #1098, opened 2015-02-21
The user's view preference must be active to be used

cbeer:
Fixes https://github.com/sul-dlss/spotlight/issues/959


lib/blacklight/catalog.rb
```diff
@@ -132,7 +132,6 @@ def has_search_parameters?
     # them default to the last used view.
     def preferred_view
```
jcoyne: This ought to be renamed. "store_preferred_view" or similar.

cbeer: ok

PR #1108, opened 2015-02-24
SearchHelper#fetch method can handle a single id or an array of ids

jcoyne:


lib/blacklight/search_helper.rb
```diff
@@ -286,22 +255,59 @@ def blacklight_solr
...
+    end
 
-    # add our document id to the document_unique_id_param query parameter
-    p = blacklight_config.default_document_solr_params.merge({
-      # this assumes the request handler will map the unique id param
-      # to the unique key field using either solr local params, the
-      # real-time get handler, etc.
-      blacklight_config.document_unique_id_param => id
-    })
+    alias_method :get_solr_response_for_document_ids, :fetch_many
```
cbeer: Does this work, even with a private method?

jcoyne: SearchHelper is usually included into the class that would call `get_solr_response_for_document_ids`, so yes.

PR #1112, opened 2015-02-24
Memoize the search_builder. Fixes #1110

### LINKED ISSUE #1110: Memoize search_builder
I'd like to be able to do this in a method:

```ruby
      search_builder.collection = @collection
      # run the solr query to find the collections
      (@response, @member_docs) = search_results(solr_params, search_params_logic)
```

However, search_results is using it's own instance.

jcoyne:


lib/blacklight/request_builders.rb
```diff
@@ -56,7 +56,7 @@ def search_builder_class
     end
 
     def search_builder processor_chain = search_params_logic
-      search_builder_class.new(processor_chain, self)
+      @search_builder ||= search_builder_class.new(processor_chain, self)
```
cbeer: Sorry, why does this need to be memoized?

jcoyne: I gave an explanation on #1110, so we can pass things to the search_builder (i.e. current_ability) without overriding `search_builder`

cbeer:
Extracted #1118 and #1119 without the memoization change.

PR #1115, opened 2015-02-24
Blacklight facet, index and show parameters should support a blacklight ...

### LINKED ISSUE #1114: Blacklight configuration should distinguish between local request params and solr field names
e.g. these should use the key, not the field:

https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/configuration/fields.rb#L138
https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/configuration/fields.rb#L113

cbeer:
...key that may be different from the solr field name.

Fixes #1114


jcoyne:
Nice! 5.10?

cbeer:
Hope so.

PR #1117, opened 2015-02-24
Lighten SearchBuilder#scope to be protected instead of private

### LINKED ISSUE #1116: SearchBuilder should have a `scope` reader

cbeer:
Fixes #1116

### DIFF
--- a/lib/blacklight/search_builder.rb
+++ b/lib/blacklight/search_builder.rb
@@ -51,9 +51,9 @@ module Blacklight
     def processed_parameters
       Blacklight::Solr::Request.new.tap do |request_parameters|
         @processor_chain.each do |method_name|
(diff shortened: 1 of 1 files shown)


jcoyne:
I'm not sure we need this to be protected. An subclass can access the private members of it's parent. https://github.com/projecthydra-labs/hydra-collections/commit/19271b3f380edc5a6a0d93a97d64794f9c6d386a#diff-861d0260f73ea1a164234913380a5d58R5
However, I'm not against this PR and It might show that this is something we intent to be part of the API.

cbeer:
>  It might show that this is something we intent to be part of the API.

That was my thinking. I have no problem with subclasses on the builder getting at scope, but someone outside probably should be discouraged from it.

PR #1119, opened 2015-02-25
Add SearchBuilder#where

cbeer:
Extracted from #1112

### DIFF
--- a/lib/blacklight/search_builder.rb
+++ b/lib/blacklight/search_builder.rb
@@ -21,6 +21,13 @@ module Blacklight
     end
 
     ##
+    # Update the :q (query) parameter
+    def where conditions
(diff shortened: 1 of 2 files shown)


lib/blacklight/search_builder.rb
```diff
@@ -21,6 +21,13 @@ def with blacklight_params = {}
     end
 
     ##
+    # Update the :q (query) parameter
+    def where conditions
```
cbeer: @jcoyne I wonder if we actually need `where` vs adding a `append_params` or similar method.

jcoyne: I don't actually care what it's called.

cbeer: I'm just trying to clarify the goal.

PR #1126, opened 2015-02-28
Use RSolr.solr_escape in place of ::escape

awead:
Avoids deprecation warning in RSolr.

### DIFF
--- a/blacklight.gemspec
+++ b/blacklight.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rails",     ">= 3.2.6", "< 5"
   s.add_dependency "nokogiri",  "~>1.6"     # XML Parser
   s.add_dependency "kaminari", "~> 0.13"  # the pagination (page 1,2,3, etc..) of our search results
-  s.add_dependency "rsolr",     "~> 1.0.6"  # Library for interacting with rSolr.
+  s.add_dependency "rsolr",     "~> 1.0.11"  # Library for interacting with rSolr.
   s.add_dependency "bootstrap-sass", "~> 3.2"
   s.add_dependency "deprecation"
 
--- a/lib/blacklight/solr/search_builder.rb
+++ b/lib/blacklight/solr/search_builder.rb
@@ -241,13 +241,13 @@ module Blacklight::Solr
           facet_config.query[value][:fq]
         when (facet_config and facet_config.date)
           # in solr 3.2+, this could be replaced by a !term query
-          "#{prefix}#{facet_field}:#{RSolr.escape(value)}"
+          "#{prefix}#{facet_field}:#{RSolr.solr_escape(value)}"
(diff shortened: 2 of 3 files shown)


cbeer:
@awead you'll also have to update the gemspec to require that newer version of RSolr. I'm curious why the method name changed, if it is still performing the same task..

awead:
@cbeer I updated the gemspec and also altered the tests that were failing to illustrate the issue. I don't know if `1.11` versus `1\\.11` is significant. It seemed significant when I tried it out in some searches. A possible solution would be to use a local method of escaping that calls the RSolr method and then adds escaped floats. This is similar to what I did in projecthydra/active_fedora#722

cbeer:
Is it significant in a way that rsolr/rsolr should fix, or just different?

awead:
@cbeer I don't know the use case the Blacklight is trying to solve with float. However, the decimal point is not listed as an escaped special character: http://lucene.apache.org/core/4_0_0/queryparser/org/apache/lucene/queryparser/classic/package-summary.html
So, I would say that rsolr/rsolr should keep in line with Lucene and that we should amend Blacklight to address the use case, or either revisit the use case altogether.

cbeer:
What were some of the differences you saw? I'm not sure I have any solr data with floats..

awead:
@cbeer just that I seemed to get search results with an escaped decimal, and did not otherwise. This was using the solr data in the blacklight test suite, which I don't think is a good example. I don't think you have any solr data with floats either.

awead:
@cbeer I ran a test with a document:
`{ "title_t" => "1.11" }`
with the following search:
```
Blacklight.solr.get "select", params: { q: "1.11" }
```
It retrieves the document, so it looks like escaping decimal points in floats in unnecessary. This PR reflects that change in the tests.

jcoyne:
@awead update the gemspec and we can merge this.

awead:
@jcoyne it's `"~> 1.0.6"` right now. Do you want it changed to `"~> 1.0.11"` ?

cbeer:
Or can we support either version with a little cleverness?

jcoyne:
@awead that's what I was suggesting.

awead:
@jcoyne It's a breaking change in RSolr, I think, so it would go to 2.0. `">= 1.0.6"` then?

jcoyne:
@awead Just change this line:
https://github.com/projectblacklight/blacklight/blob/master/blacklight.gemspec#L23

to ` ~> 1.0.11` I don't know why you think that would be a breaking change. We're just updating our dependencies. What is broken?

awead:
@jcoyne ok done. I was over-thinking it.  I thought you wanted to support RSolr 1 and 2, when it came out because the deprecated method would mean a breaking change.

jcoyne:
@awead Great. (FWIW there is no RSolr 2).

PR #1128, opened 2015-03-03
Use modules instead of classes for SearchBuilder

jcoyne:
This enables us to generate custom SearchBuilders that aren't tightly coupled to Solr (or ElasticSearch, etc.)


lib/blacklight/search_builder_behavior.rb
```diff
@@ -1,5 +1,5 @@
 module Blacklight
-  class SearchBuilder
+  module SearchBuilderBehavior
```
cbeer: I'm not sure I understand why this shouldn't be a class. It sure looks like we took class inheritance and turned it into module includes instead.

cbeer:
:-1: I'm not sure what this is buying us other than turning classes that behaved as classes into modules that very nearly behave as classes but require an intermediate, usually empty, class instead. `Blacklight::Solr::SearchBuilderBehavior` has the same coupling to Solr as before, doesn't it?

jcoyne:
@cbeer but now when you make a custom search builder it doesn't need to extend Blacklight::Solr::SearchHelper.  I'm thinking in the future the implementation specific helper can be included automatically based on configuration in blacklight.yml

jcoyne:
@cbeer also if we generate a custom search helper (say for advance search) we can do:
```ruby
class SearchBuilder
  include Blacklight::Solr::SearchBuilder
  # advance search stuff.
end
```

cbeer:
Isn't that what @barmintor's `ansr` is going to give us (and push the abstraction down a level)? Right now, Solr and anything else are not interchangeable and there's no overlap between a Solr parameter translator and something for ES.

PR #1132, opened 2015-03-09
Fix so that ruby 1.9 can find the SolrDocument class

jcoyne:
See
https://travis-ci.org/projectblacklight/blacklight_advanced_search/builds/52804516


cbeer:
5.10.2?

PR #1140, opened 2015-03-11
Gracefully support Rails version differences instead of doing version ch...

### LINKED ISSUE #1138: add attr_accessible if Rails version is less than 4 or Object.const_defined?('ProtectedAttributes')
I'm upgrading a large application from Rails 3 to Rails 4. To smooth the upgrade I'm using the protected_attributes gem and using attr_accessible to my models. This application includes a rather standard Blacklight within it which is now upgraded to latest 5.10. For the Blacklight upgrade some functionality like going to a show view with a current_session causes an error because the search is created but the query_params are not saved. Setting attr_accessible :query_params fixes this.

In app/models/search.rb:8 something like the following would fix this:

```ruby
if Rails::VERSION::MAJOR < 4 || Object.const_defined?('ProtectedAttributes')
  attr_accessible :query_params
end
if Rails::VERSION::MAJOR < 4
  scope :none, where(:id => nil).where("id IS NOT ?", nil)
end
```

Is this something that could be added? The alternative for me would be to override the Search model completely.

cbeer:
...

Fixes #1138 

@jronallo can you give this a try?

### DIFF
--- a/app/models/bookmark.rb
+++ b/app/models/bookmark.rb
@@ -5,7 +5,10 @@ class Bookmark < ActiveRecord::Base
   belongs_to :document, polymorphic: true
 
   validates_presence_of :user_id, :scope=>:document_id
-  attr_accessible :id, :document_id, :document_type, :title if Rails::VERSION::MAJOR < 4
+
+  if Blacklight::Utils.needs_attr_accessible?
+    attr_accessible :id, :document_id, :document_type, :title
+  end
 
   def document
     document_type.new document_type.unique_key => document_id
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -5,10 +5,13 @@ class Search < ActiveRecord::Base
 
   serialize :query_params
 
-  if Rails::VERSION::MAJOR < 4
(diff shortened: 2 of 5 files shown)


cbeer:
I guess that answers why. Rails 4.0 defines the method, and then raises an error when you call it:
https://github.com/rails/rails/blob/v4.0.13/activemodel/lib/active_model/deprecated_mass_assignment_security.rb#L14

jronallo:
@cbeer So this definitely works for me, but as the test results show it would fail for anyone on Rails 4 that does not use the protected_attributes gem. Since rails continues to define the method and then raise, I don't think we can just check for whether the method is defined. Since the default and suggestion in Rails 4 is to protect mass assignment outside the model, checking in this way seems like a bad default for Blacklight to ship with.

jcoyne:
I've written this patch before: https://github.com/jcoyne/devise/commit/b7e67115fce9629a90e974a95906bce34716173e

cbeer:
Thanks @jcoyne, that looks pretty good. I'll give that a go.

PR #1144, opened 2015-03-13
Add a more helpful default _home_text.html.erb partial

cbeer:
screen shot 2015-03-13 at 9 59 03 am

If you have any suggestions for this page, feel free to make the change directly on this branch.

### DIFF
--- a/app/views/catalog/_home_text.html.erb
+++ b/app/views/catalog/_home_text.html.erb
@@ -1,9 +1,49 @@
-<h4><%= t('blacklight.welcome') %></h4>
-<p>&nbsp;</p>
-<p>To modify this text to your specifications, copy this file located in the blacklight plugin directory:<br/>
-&nbsp;&nbsp;&nbsp;<%= Blacklight.root %>/app/views/catalog/_home_text.html.erb</p>
-<p>to your top level rails app:<br/>
-&nbsp;&nbsp;&nbsp;<%= Rails.root %>/app/views/catalog/_home_text.html.erb</p>
-<p>Note that you will need to create the 'catalog' directory under views.</p>
-<% #Creates a bit space in the page -%>
+<div class="page-header row">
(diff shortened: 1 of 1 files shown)


app/views/catalog/_home_text.html.erb
```diff
@@ -1,9 +1,49 @@
...
+
+</div>
+
+<div id="getting-started">
+  <h2>Here&rsquo;s how to get started:</h2>
+  
+  <ol>
+    <li>To modify this text, you need to <a href="http://guides.rubyonrails.org/engines.html#improving-engine-functionality">override the Blacklight-provided view</a>.
+You can copy this file, located in the blacklight gem: <br />
+<%= Blacklight.root %>/app/views/catalog/_home_text.html.erb <br />
```
jkeck: I have thought about this recently as I've seen people deploy vanilla Blacklight to a server and then the file path to where the app exists on their server is now on their home-page.  I wonder if we could just put a place-holder piece of text that tell them that we mean the app root `#{YOUR_APP_ROOT}` or maybe not put the full file path in unless under development.

No big deal, just a thought that came up after seeing a few BL demo instances out in the wild.  Take it or leave it.

cbeer: I'm hoping the more explicit call-to-action will help. We could add all kinds of checks (`Rails.env.development?`, `localhost`, etc), but if they don't feel compelled to blank out a partial, I'm not sure how much more help we should be.

jkeck: :+1: totally agree.

It's also not up to us to make sure folks don't put sensitive information in their file paths (even though I see it all the time...)  Just thought I would throw it out there since we're touching the file (for the first time since 2009 it looks like!)

cbeer: At then some, I'd guess. I saw it talking about the `plugins` directory and figured it was time for an overhaul.

PR #1148, opened 2015-03-17
adds in semantic classes

mejackreed:
Pulled out semantic classes from https://github.com/projectblacklight/blacklight/pull/1124


cbeer:
:+1:

PR #1152, opened 2015-03-19
Generate SearchBuilder into the host application

jcoyne:
This gives plugins a place to add additional search methods without
having an complex inheritance hierarchy.


lib/blacklight/search_builder.rb
```diff
@@ -1,140 +1,6 @@
 module Blacklight
   class SearchBuilder
-    extend Deprecation
```
cbeer: I still don't understand why this can't be a class, and have `::SearchBuilder` extend it if it needs to.

lib/generators/blacklight/document_generator.rb
```diff
@@ -16,5 +16,9 @@ def create_solr_document
       template "solr_document.rb", "app/models/#{model_name}.rb"
     end
 
+    def create_search_builder
```
cbeer: Can we make this into a search_builder_generator instead of adding it here?

lib/blacklight/configuration.rb
```diff
@@ -212,7 +212,7 @@ def repository_class
     end
 
     def search_builder_class
-      super || Blacklight::Solr::SearchBuilder
+      super || ::SearchBuilder
```
cbeer: Can we do this and still claim backwards compatibility, or do we need to check if it is defined first?

lib/blacklight/search_builder.rb
jcoyne: That would be fine with me. I can rework it.

cbeer: Thanks.

lib/generators/blacklight/document_generator.rb
jcoyne: Yeah, no problem

lib/blacklight/configuration.rb
```diff
@@ -212,7 +212,14 @@ def repository_class
...
 
     def search_builder_class
-      super || Blacklight::Solr::SearchBuilder
+      super || locate_search_builder_class
+    end
+
+    def locate_search_builder_class
+      ::SearchBuilder
+    rescue NameError
+      Deprecation.warn(Configuration, "Your application is missing the SearchBuilder. Have you run `rails generate blacklight:search_builder`? Falling back to Blacklight::Solr::SearchBuilder")
```
cbeer: Do we really need to deprecation `Blacklight::Solr::SearchBuilder`? It seems like it works fine as-is for common uses..

jcoyne: I'm for removing as much code as possible, so I think that deprecating it is the thing to do.

PR #1157, opened 2015-03-20
Update #facet_in_params? and #facet_field_in_params? to use the blacklight-configured field names

cbeer:


app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -170,13 +170,17 @@ def facet_field_in_params? field
   # @param [Object] facet value
   # @return [Boolean]
   def facet_in_params?(field, item)
-    if item and item.respond_to? :field
-      field = item.field
```
jkeck: Is it possible that somebody would be depending on this behavior (albeit pretty strange to use IMO)?

cbeer: `facet_configuration_for_field(field)` (in `#facet_params`) should be equivalent.

Issue #1130, opened 2015-03-05
when facet.missing=true


whisere:
When facet field set as true in solr
The facet item shows as url instead of 'missing' on the facet bar
eg:
Event 100
http://localhost:3000/spotlight/default-exhibit/catalog?exhibit_id=default-exhibit&f%5Blocation%5D%5B%5D=&only_path=true 8

jcoyne:
I'm not sure I understand the issue here. Do you have a problem with Blacklight or with Spotlight? Can you explain how to reproduce the issue or share a picture of the problem?

cbeer:
I believe this is from the blacklight list. Blacklight does not support Solr's `facet.missing` option at all, and with the way Solr implements `facet.missing` (tacking on a `nil` => (hits) to the facet values), it breaks Rails routing is bad ways. It also seems fairly non-trivial to try add support for `facet.missing` given its serialization, so we'd have to bake-in support for it throughout the stack.

whisere:
![screenshot at 2015-03-05 16 05 46](https://cloud.githubusercontent.com/assets/2360494/6499563/a672e600-c351-11e4-95a0-42792bace7b4.png)

whisere:
I have attached a picture of the facet bar item when facet.missing is set to true. Thanks!

jcoyne:
@whisere are you going to work on a fix for this?  I doubt that anyone else will be working on it.

whisere:
I have tried but without success, any suggestions? I am a bit new to ruby on rails and Blacklight...

whisere:
I found one solution to this, instead of fixing it at this end, I did some change at solr config: 1. disable facet.missing=true in solrconfig.xml; 2. change the content of the field from empty to 'missing' or other preferred value eg 'unknown' in data-config.xml 
But this may be more straightforward with JdbcDataSource 
Let me know if you find other solutions.

PR #1162, opened 2015-03-20
Add Blacklight::SolrResponse#aggregations to abstract solr's various facet flavors

cbeer:


app/helpers/blacklight/catalog_helper_behavior.rb
```diff
@@ -14,11 +14,9 @@ def page_entries_info(collection, options = {})
     entry_name = if options[:entry_name]
       options[:entry_name]
     elsif collection.respond_to? :model  # DataMapper
-        collection.model.model_name.human.downcase
+      collection.model.model_name.human.downcase
     elsif collection.respond_to? :model_name and !collection.model_name.nil? # AR, Blacklight::PaginationMethods
-        collection.model_name.human.downcase
-    elsif collection.is_a?(::Kaminari::PaginatableArray)
-      'entry'
```
cbeer: "entry" is our i18n label too.

lib/blacklight/solr_response/pagination_methods.rb
```diff
@@ -14,21 +14,4 @@ def offset_value #:nodoc:
   def total_count #:nodoc:
     total
   end
-
-  def model_name
```
cbeer: `#model_name` didn't seem to work (or, when it worked like I would expect, we had test failures.), so I've removed it.

lib/blacklight/solr_response/pagination_methods.rb
```diff
@@ -14,21 +14,4 @@ def offset_value #:nodoc:
...
-    if !docs.empty? and docs.first.respond_to? :model_name
-      docs.first.model_name
-    end
-  end
-
-  ## Methods in kaminari master that we'd like to use today.
-  # Next page number in the collection
-  def next_page
```
cbeer: These were added in kaminari 0.15, so I've dropped them here.

cbeer:
Also, should fix #1130.

PR #1165, opened 2015-03-21
When providing default or backwards-compatible options for a field if/un...

cbeer:
...less condition, don't overwrite the boolean false


jcoyne:
Looks good, but needs a rebase.

PR #1170, opened 2015-03-23
Consistently use Response#aggregation

cbeer:
Also fixes deprecation warnings from calling old behavior


jcoyne:
:+1: Looks good. Thanks.

PR #1174, opened 2015-03-25
Relax spec on kaminari, so the 1.0.0.alpha (master) branch can be used

jcoyne:


blacklight.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency "rails",     ">= 3.2.6", "< 5"
   s.add_dependency "nokogiri",  "~>1.6"     # XML Parser
-  s.add_dependency "kaminari", "~> 0.15"  # the pagination (page 1,2,3, etc..) of our search results
+  s.add_dependency "kaminari", ">= 0.15", "< 1.1.0"   # the pagination (page 1,2,3, etc..) of our search results
```
cbeer: Why bother with `< 1.1.0`? Seems like anything could happen between `0.15` and `1.0` anyway..

jcoyne: They already have bumped the version in master to 1.0.0.alpha.  I have this branch that I want to use: https://github.com/jcoyne/kaminari/tree/route_prefix_prototype

cbeer: Yes, I get that, I'm just wondering why we're being optimistic about `0.15` to `1.0`, but pessimistic about `1.0` to `1.1`. If they're doing semver, I'd think we could allow any `1.x` release.

cbeer:
:+1:

PR #1179, opened 2015-03-27
Pass the Blacklight::SearchBuilder object to the repository instead of just the request parameters

cbeer:


jcoyne:
Can you say a bit more about why the change tracking was added?

cbeer:
In order to cache the result of `#to_hash`, which I believe gets used multiple times downstream.

PR #1180, opened 2015-04-02
Configurable show classes

mejackreed:
This pull request enables more configuration on a document's show page, without having to override a partial.

 1. 996a6ae moves most of the content within https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/show.html.erb#L3-29 into a new partial `_show_main_content.html.erb`
 1. 4461f5b makes the classes used for both the content and sidebar divs configurable via new helpers. These new helpers are part of a new `LayoutHelper` module

### DIFF
--- a/app/helpers/blacklight/layout_helper_behavior.rb
+++ b/app/helpers/blacklight/layout_helper_behavior.rb
@@ -1,0 +1,34 @@
+# Methods added to this helper will be available to all templates in the hosting
+# application
+module Blacklight
+  # A module for useful methods used in layout configuration
+  module LayoutHelperBehavior
+    ##
+    # Classes added to a document's show content div
+    # @return [String]
+    def show_content_classes
+      "#{main_content_classes} show-document"
+    end
+
+    ##
+    # Classes added to a document's sidebar div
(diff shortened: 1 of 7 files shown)


app/helpers/blacklight/layout_helper_behavior.rb
```diff
@@ -0,0 +1,20 @@
+# Methods added to this helper will be available to all templates in the hosting
+# application
+module Blacklight
+  # A module for useful methods used in layout configuration
+  module LayoutHelperBehavior
+    ##
+    # Classes to be added to a document's show content div
```
cbeer: It might be good to separate the column sizes from the semantic classes.. Or, I know we use the same column sizes in `catalog/index.html.erb` and we should probably be consistent.

mejackreed:
@cbeer great suggestion. Updated accordingly, and added modified index to use new helpers.

cbeer:
:+1: (although I wish I could offer better names for that partial or helper..)

mejackreed:
Yes... name suggestions are welcome. The most time was spent on figuring out what to name it

PR #1183, opened 2015-04-06
Hash access methods

cbeer:


lib/blacklight/utils.rb
```diff
@@ -26,7 +26,7 @@ def self.protected_attributes_enabled?
   ##
   # An OpenStruct that responds to common Hash methods
   class OpenStructWithHashAccess < OpenStruct
-    delegate :keys, :each, :map, :has_key?, :empty?, :delete, :length, :reject!, :select!, :include, :fetch, :to_json, :as_json, :to => :to_h
+    delegate :keys, :each, :map, :has_key?, :include?, :empty?, :length, :delete, :delete_if, :keep_if, :clear, :reject!, :select!, :replace, :fetch, :to_json, :as_json, to: :to_h
```
cjcolvar: Does it matter that #include changed to #include? ?  I think the change makes sense since it aligns with the method name on Hash but unsure if anything called #include that would break with the change.

cbeer: I don't think `Hash#include` was ever a reasonable method to delegate, and I think `#include?` was always the intended method`.

PR #1187, opened 2015-04-09
SolrRepository#find should prefer the default_document_solr_params[:qt] over document_solr_request_handler

### LINKED ISSUE #1186: config.default_document_solr_params seems to have no effect
Using Blacklight 5.11.0 I have the following setting in my catalog_controller.rb:

```ruby
config.default_document_solr_params = {
   qt: 'document_asbury'
}
```

When I look in my logs for a show view request I still see that the qt parameter is set to "document".

What I have been able to do to work around this is the following as I am only setting the request handler:

```ruby
config.document_solr_request_handler = 'document_asbury'
```

cbeer:
Fixes #1186

### DIFF
--- a/lib/blacklight/solr_repository.rb
+++ b/lib/blacklight/solr_repository.rb
@@ -12,8 +12,8 @@ module Blacklight
     # @param [String] document's unique key value
     # @param [Hash] additional solr query parameters
     def find id, params = {}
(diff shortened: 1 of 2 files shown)


cbeer:
I guess we should back-port this too...

jcoyne:
Looks like execjs and ruby 1.9 don't play well together.

PR #1192, opened 2015-04-13
Add clearer error on missing secret key

jechols:
Helps when doing a quick test of a production environment that isn't yet set up to know we only need the secret key base.

### DIFF
--- a/lib/blacklight/rails/routes.rb
+++ b/lib/blacklight/rails/routes.rb
@@ -18,7 +18,7 @@ module ActionDispatch::Routing
     private
     def raise_no_blacklight_secret_key #:nodoc:
       raise <<-ERROR
-Blacklight.secret_key was not set. Please add the following to an initializer:
+Blacklight.secret_key was not set. If you are using Rails 4.1+, set your app's secret key base (config/secrets.yml) OR add the following to an initializer:
 
 Blacklight.secret_key = '#{SecureRandom.hex(64)}'
 


lib/blacklight/rails/routes.rb
```diff
@@ -18,7 +18,7 @@ def blacklight_for(*resources)
     private
     def raise_no_blacklight_secret_key #:nodoc:
       raise <<-ERROR
-Blacklight.secret_key was not set. Please add the following to an initializer:
+Blacklight.secret_key was not set. Please either set your app's secret key base (generally in config/secrets.yml) or add the following to an initializer:
```
cbeer: Should we mention `config/secrets.yml` is a Rails 4.1+ thing?

jechols: Almost definitely - didn't think of that

jechols:
Better?

lib/blacklight/rails/routes.rb
```diff
@@ -18,7 +18,7 @@ def blacklight_for(*resources)
     private
     def raise_no_blacklight_secret_key #:nodoc:
       raise <<-ERROR
-Blacklight.secret_key was not set. Please add the following to an initializer:
+Blacklight.secret_key was not set. Please either add the following to an initializer or, in Rails 4.1+, set your app's secret key base (Rails 4.1+ config/secrets.yml).
```
jcoyne: Let's rephrase slightly:
`Blacklight.secret_key was not set. If you are using Rails 4.1+, set your app's secret key base (config/secrets.yml) OR add the following to an initializer:`

cbeer: :+1:

jechols:
Sorry, totally lost track of this one.  Wording now reflects @jcoyne's suggestion

cbeer:
:+1:

PR #1199, opened 2015-04-20
Add URL method config for facets

### LINKED ISSUE #1196: Override Path Facet Goes To
I think something like a :url_method setter on the facet configuration would be able to handle this. My use case is making it so that when on the root blacklight page you can click a collection facet and rather than facet on that collection it sends you to a collection landing page. We do this now, but we did it via hacking helpers and this seems like a prettier way to implement it.

terrellt:
This allows for facets to go somewhere other than the catalog path when clicked.
Closes #1196

### DIFF
--- a/app/helpers/blacklight/facets_helper_behavior.rb
+++ b/app/helpers/blacklight/facets_helper_behavior.rb
@@ -123,10 +123,24 @@ module Blacklight::FacetsHelperBehavior
   # @option options [Boolean] :suppress_link display the facet, but don't link to it
   # @return [String]
   def render_facet_value(facet_field, item, options ={})
-    path = search_action_path(add_facet_params_and_redirect(facet_field, item))
+    path = path_for_facet(facet_field, item)
     content_tag(:span, :class => "facet-label") do
       link_to_unless(options[:suppress_link], facet_display_value(facet_field, item), path, :class=>"facet_select")
     end + render_facet_count(item.hits)
+  end
+
+  ##
+  # Where should this facet link to?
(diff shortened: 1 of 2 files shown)


app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -123,7 +123,12 @@ def facet_partial_name(display_facet = nil)
   # @option options [Boolean] :suppress_link display the facet, but don't link to it
   # @return [String]
   def render_facet_value(facet_field, item, options ={})
-    path = search_action_path(add_facet_params_and_redirect(facet_field, item))
+    facet_config = facet_configuration_for_field(facet_field)
+    if facet_config.url_method
```
jcoyne: Please refactor this block into a new method:
```ruby
path = path_for_facet(...)
```

terrellt: Sure thing, should be taken care of.

terrellt:
@jcoyne Missing anything else?

cbeer:
+0

terrellt:
> +0

Ambivalence?

cbeer:
Exactly, but not opposed. I can almost see how we might use it locally.

terrellt:
@cbeer: The only other way I could get this to work before was with code like https://github.com/OregonDigital/oregondigital/blob/master/app/helpers/collections_helper.rb#L28-L38, which keeps me up at night.

terrellt:
@jcoyne @cbeer So is this mergable?

PR #1201, opened 2015-05-13
Add a CacheKey module that is mixed into Blacklight::Document

jkeck:
This returns a cache_key method for classes that mixin `Blacklight::Document` that uses the class name, item id, and a formatted updated at timestamp.

### DIFF
--- a/lib/blacklight/document/cache_key.rb
+++ b/lib/blacklight/document/cache_key.rb
@@ -1,0 +1,23 @@
+# -*- encoding : utf-8 -*-
+# This module provides the cache key which can be used by rails
+# caching to determine when to expire a particular object's cache
+# See http://apidock.com/rails/ActiveRecord/Integration/cache_key
+# This key should be used in conjunction with additional data to
+# determine when a document can be cached (e.g. for different view
+# types in search results like gallery and list)
+module Blacklight::Document::CacheKey
+  def cache_key
+    case
+    when new_record?
+      "#{self.class.model_name.cache_key}/new"
+    when cache_version_value = self[cache_version_key]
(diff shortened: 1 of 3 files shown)


lib/blacklight/document/cache_key.rb
```diff
@@ -0,0 +1,28 @@
...
+    when timestamp = self[timestamp_key]
+      timestamp = Time.zone.parse(timestamp).utc.to_s(cache_timestamp_format)
+      "#{self.class.model_name.cache_key}/#{id}-#{timestamp}"
+    else
+      "#{self.class.model_name.cache_key}/#{id}"
+    end
+  end
+
+  def timestamp_key
+    :last_updated
```
cbeer: I'm not sure we should tie this directly to timestamps; I can imagine using `__version__` (baking into Solr 5.x, and a requirement for SolrCloud in 4.x) here instead.

jkeck: :+1: fair enough.  Let me know if how I've changed it is more agreeable.

lib/blacklight/document/cache_key.rb
```diff
@@ -0,0 +1,23 @@
...
+# See http://apidock.com/rails/ActiveRecord/Integration/cache_key
+# This key should be used in conjunction with additional data to
+# determine when a document can be cached (e.g. for different view
+# types in search results like gallery and list)
+module Blacklight::Document::CacheKey
+  def cache_key
+    case
+    when new_record?
+      "#{self.class.model_name.cache_key}/new"
+    when timestamp = self[cache_version_key]
```
cbeer: `s/timestamp/cache_key_value/`?

Issue #1207, opened 2015-05-28
RSolr gemfile warning triggered by default_configuration


rwd:
The deprecation warning in `Blacklight::SolrRepository` about RSolr not being loaded is always triggered when `Blacklight::Configurable.default_configuration` is called, even if the application sets its own `repository_class`.

This is due to the logic in `Blacklight::Configuration.deep_copy` for Rails 4+ which loads the default classes even if they are not yet set.

As well as triggering the deprecation warning which is not relevant for an application not using Solr as its data source, this causes the autoloading of Blacklight classes not needed by the application, such as `Blacklight::SolrRepository`.

There is a fix for this in rwd/blacklight@739641dd29e2165668e99ca55d03db1ba442a32d, but a pull request for that is dependent on prior merging of pull request projectblacklight/blacklight/pull/1203

cbeer:
Looks like we merged this in #1208.

PR #1208, opened 2015-05-28
Don't copy configuration model classes if nil.

rwd:
Fixing issue #1207 by leaving the class configuration settings alone if they are `nil`.


mejackreed:
@cbeer 5.14?

PR #1211, opened 2015-06-08
has_facet_values? should return false if facets have values but are not displayable

cjcolvar:
I ran into this edge case where there are facets with values but they are not displayable.  This leaves the facets header ("Limit your search") on the page without any facets underneath it.


jcoyne:
Looks good to me. :+1:

PR #1216, opened 2015-06-10
Drop support for Rails 3

### LINKED ISSUE #879: Drop support for Rails 3.x
Rails 4.0 and 4.1 have been released, and Rails 3.2 is likely in a security-patch-only mode. We're explicitly supporting Rails 3.2 in several places, and there are features in Rails 4 we could take advantage of.

cjcolvar:
Fixes #879

### DIFF
--- a/lib/generators/blacklight/templates/config/initializers/blacklight_initializer.rb
+++ b/lib/generators/blacklight/templates/config/initializers/blacklight_initializer.rb
@@ -1,9 +1,0 @@
-# A secret token used to encrypt user_id's in the Bookmarks#export callback URL
-# functionality, for example in Refworks export of Bookmarks. In Rails 4, Blacklight
-# will use the application's secret key base instead.
-#
-<% if Rails::VERSION::MAJOR == 4 %>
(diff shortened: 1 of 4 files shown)


lib/generators/blacklight/templates/config/initializers/blacklight_initializer.rb
```diff
@@ -2,8 +2,4 @@
 # functionality, for example in Refworks export of Bookmarks. In Rails 4, Blacklight
```
cbeer: Can we clean this up even more? We should use Rails' SECRET_KEY_BASE to seed `Blacklight.secret_key`, and not bother exposing any special configuration.

cjcolvar: Isn't it already seeded? https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/engine.rb#L23-L29

So it should be safe to remove this blacklight_initalizer template then?

cbeer: :+1: sounds good to me.

PR #1222, opened 2015-06-15
Add support for globalid

### LINKED ISSUE #1120: Add GlobalID Support for SolrDocument
This will not work in older versions of rails, so I'm not exactly sure how we would should handle this.

GlobalID support can be useful in that it will allow you to pass a SolrDocument as an argument to an ActiveJob (or for other useful serialization needs).

Our solr documents can return a global id simply by including `GlobalID::Identification`

    SolrDocument.new(id: '123').to_global_id
    => gid://internal/SolrDocument/123

However; since our SolrDocument class does not respond to #find w/ the document ID we can't lookup documents based on their global ID.

    GlobalID::Locator.locate(
      SolrDocument.new(id: '123').to_global_id
    )
    => undefined method `find' for SolrDocument:Class

We can use a [Custom App Locator](https://github.com/rails/globalid#custom-app-locator) to set our own locator for SolrDocuments but I'm not sure that's wiser that having SolrDocument respond to #find.

cjcolvar:
Fixes #1120

This allows for inclusion of ActiveModelShim to provide #id and #find if they aren't supplied by the class including Blacklight::Document.  These two methods are required for globalid to function properly.  It looks like the ActiveRecord and ElasticSearch Document implementations in #1156 supply these already.

@jkeck @cbeer

### DIFF
--- a/lib/blacklight/document/active_model_shim.rb
+++ b/lib/blacklight/document/active_model_shim.rb
@@ -14,6 +14,14 @@ module Blacklight::Document
       def base_class
         self
       end
+
+      def repository
+        Blacklight.default_index
+      end
+
+      def find id
+        repository.find(id).documents.first
+      end
     end
     
     ##
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ rvm:
 matrix:
   include:
     - rvm: 2.2.2
-      env: "RAILS_VERSION=4.0.12"
-    - rvm: 2.2.2
       env: "RAILS_VERSION=4.1.9"
     - rvm: 2.1.6
       env: "RAILS_VERSION=4.2.1"
--- a/blacklight.gemspec
+++ b/blacklight.gemspec
@@ -17,7 +17,8 @@ Gem::Specification.new do |s|
   s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
   s.require_paths = ["lib"]
 
-  s.add_dependency "rails",     ">= 3.2.6", "< 5"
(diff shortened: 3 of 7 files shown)


lib/blacklight/document/active_model_shim.rb
```diff
@@ -14,6 +14,11 @@ def primary_key
       def base_class
         self
       end
+
+      def find id
+        response = Blacklight.default_index.find id
```
cbeer: Can we pull `Blacklight.default_index` into a method that we can override?

lib/blacklight/document/active_model_shim.rb
```diff
@@ -14,6 +14,11 @@ def primary_key
...
+        self.new(response.documents.first, response)
```
cbeer: Do we need the `self.new` business? `response.documents.first` should be equivalent (unless we're worried about casting between document types, i suppose?)

cbeer:
Nice

lib/blacklight/document/active_model_shim.rb
cjcolvar: Ah, I see now.  Response is supposed to cast each document into blacklight_config.document_model.  I'll take the self.new out and clean up the tests below.

cbeer:
And it looks like you should drop the Rails 4.0 build from the travis matrix.

jkeck:
:+1:

cbeer:
:+1: I wonder if we should leave this open for a little bit, in case anyone wants to plead their case for Rails 4.0 support in Blacklight 6.

PR #1224, opened 2015-06-17
Fix facet wrapping/hyphenation in Firefox

bkeese:
This also allows IE and chromium (and maybe others) to split very long words instead of overflowing the sidebar. This solution uses javascript to set the width of the facet count columns so that browsers can calculate the width of the facet label column. They need this width to correctly split overflowing words.

### DIFF
--- a/app/assets/javascripts/blacklight/facet_load.js
+++ b/app/assets/javascripts/blacklight/facet_load.js
@@ -1,4 +1,3 @@
-//= require blacklight/core
 (function($) {
   Blacklight.onLoad(function() {
     // adjust width of facet columns to fit their contents
--- a/app/assets/javascripts/blacklight/facet_load.js
+++ b/app/assets/javascripts/blacklight/facet_load.js
@@ -5,7 +5,7 @@
     // adjust width of facet columns to fit their contents
     function longer (a,b){ return b.textContent.length - a.textContent.length; }
     $('ul.facet-values, ul.pivot-facet').map(function(){
-      var longest = $(this).find('.facet-count span').sort(longer).first();
+      var longest = $(this).find('span.facet-count').sort(longer).first();
       var clone = longest.clone().css('visibility','hidden');
       $('body').append(clone);
       $(this).find('.facet-count').first().width(clone.width());
--- a/app/assets/javascripts/blacklight/facet_load.js
+++ b/app/assets/javascripts/blacklight/facet_load.js
@@ -2,10 +2,10 @@
 (function($) {
   Blacklight.onLoad(function() {
     // adjust width of facet columns to fit their contents
+    function longer (a,b){ return b.textContent.length - a.textContent.length; }
     $('ul.facet-values, ul.pivot-facet').map(function(){
-      function longer (a,c){return a.textContent.length>c.textContent.length?a:c;}
-      var longest = Array.prototype.reduce.apply($(this).find('.facet-count span'),[longer]);
-      var clone = $(longest).clone().css('visibility','hidden');
+      var longest = $(this).find('.facet-count span').sort(longer).first();
+      var clone = longest.clone().css('visibility','hidden');
       $('body').append(clone);
       $(this).find('.facet-count').first().width(clone.width());
       clone.remove();
--- a/app/assets/javascripts/blacklight/facet_load.js
+++ b/app/assets/javascripts/blacklight/facet_load.js
@@ -1,0 +1,14 @@
+//= require blacklight/core
+(function($) {
+  Blacklight.onLoad(function() {
+    // adjust width of facet columns to fit their contents
+    $('ul.facet-values, ul.pivot-facet').map(function(){
+      function longer (a,c){return a.textContent.length>c.textContent.length?a:c;}
+      var longest = Array.prototype.reduce.apply($(this).find('.facet-count span'),[longer]);
+      var clone = $(longest).clone().css('visibility','hidden');
+      $('body').append(clone);
+      $(this).find('.facet-count').first().width(clone.width());
+      clone.remove();
+    });
+  });
+})(jQuery);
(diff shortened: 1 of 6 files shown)


cjcolvar:
:+1:  This fixes #1080

app/assets/stylesheets/blacklight/_facets.scss
```diff
@@ -67,33 +68,33 @@
     // breaks long words apart so they don't cause the containing div to
     // be too big
     // from http://kenneth.io/blog/2012/03/04/word-wrapping-hypernation-using-css/
```
cbeer: Should this link be removed?

app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -124,7 +124,7 @@ def facet_partial_name(display_facet = nil)
   # @return [String]
   def render_facet_value(facet_field, item, options ={})
     path = path_for_facet(facet_field, item)
-    content_tag(:span, :class => "facet-label") do
+    content_tag(:span, :class => "facet-label col-xs-10 col-sm-8 col-md-9 col-lg-10") do
```
cbeer: Can we pull this list out into a helper or something (@extend in the CSS for `.facet-label`?) to avoid repeating it?

cbeer:
One of the "features" that this PR doesn't support is ensuring the first line of the facet value and count are on the same line:

screen shot 2015-06-17 at 8 54 20 am

After this PR, I get:

screen shot 2015-06-17 at 8 55 08 am

app/assets/stylesheets/blacklight/_facets.scss
```diff
@@ -36,12 +36,12 @@
 
 .facet-values {
   display: table;
+  table-layout: fixed;
```
bkeese: This is a different approach, but still has the problem of needing a fixed width on the "columns". This sets the count column larger to accommodate larger numbers.

jrochkind:
at one point I think i did a bunch of CSS hacking to try and get the count on the same line as the first line of the label.  Would prefer that not go away. nice catch @cbeer.

app/assets/javascripts/blacklight/facet_load.js
```diff
@@ -0,0 +1,15 @@
+(function($) {
+  Blacklight.onLoad(function() {
```
cbeer: @jkeck i know we're run into this elsewhere -- any idea how to deal with it? Lots of js requires?

jkeck: AFAICT even explicitly requiring the file does not help.  I don't think @houndci evaluates required files (but I could be wrong, that's just in my limited experience).

I think I have just let these go for the most part in the requests project because I could never figure out how to get hound to stop complaining.

app/assets/javascripts/blacklight/facet_load.js
```diff
@@ -0,0 +1,14 @@
+//= require blacklight/core
+(function($) {
+  Blacklight.onLoad(function() {
+    // adjust width of facet columns to fit their contents
+    $('ul.facet-values, ul.pivot-facet').map(function(){
+      function longer (a,c){return a.textContent.length>c.textContent.length?a:c;}
+      var longest = Array.prototype.reduce.apply($(this).find('.facet-count span'),[longer]);
```
mejackreed: Do we have an IE version that we explicitly don't support? This will break on IE8

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

cjcolvar:
Where does this stand?  @bkeese left the commit history with a number of approaches.  Does the current approach work for everyone?  Please provide feedback so we can get this resolved.

app/assets/javascripts/blacklight/facet_load.js
```diff
@@ -0,0 +1,14 @@
+//= require blacklight/core
```
mejackreed: If `blacklight/core` is already being required, do we need to require it here again?

app/assets/javascripts/blacklight/facet_load.js
mejackreed: Ok so figured out the answer for this.. @jkeck @cbeer 

You can do this at the top of the file:

```javascript
/*global Blacklight */
```

cbeer: :+1:

cjcolvar:
Is it possible to get resolution on this during HydraConnect?

@mtribone, you mentioned that you could take a look over it.  Find me or @bkeese if you want to talk in person.

mtribone:
@cjcolvar Yeah man. Let me get through my presentations this morning and I'll give it a look-see. I'll come get you if I have questions.

mtribone:
Oh boy this is a tough one. I feel that using JS is a big hammer to get word breaks to work properly and I'm not a big fan of fixed width containers. I do understand that you're just trying to get a working solution. Is the problem with vendor prefixes is that Hound doesn't like them?

https://kenneth.io/blog/2012/03/04/word-wrapping-hypernation-using-css/

Could CSS hyphenation be an acceptable compromise?

mtribone:
Just had a good conversation with @bkeese and @cjcolvar about some of my concerns and these have been addressed. The "fixed" width is recalculated based on the size of the browser and won't, for example, be stuck at 250px. It also addresses @cbeer concern about the facet count being on the same line.

I'm good with this change. :+1:

app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -163,8 +163,10 @@ def render_selected_facet_value(facet_field, item)
   # @option options [Array<String>]  an array of classes to add to count span.
   # @return [String]
   def render_facet_count(num, options = {})
-    classes = (options[:classes] || []) << "facet-count"
-    content_tag("span", t('blacklight.search.facets.count', :number => number_with_delimiter(num)), :class => classes)
+    classes = (options[:classes] || [])
+    content_tag("div", :class => "facet-count") do
```
cbeer: What benefit does the new `div` give us?

PR #1225, opened 2015-06-22
Fix #1223

### LINKED ISSUE #1223: WARNING: Can't mass-assign protected attributes for Bookmark: document_id, document_type
After I upgrade Rails from 3.2.13 to 4.0.1 and try to save item in 'Bookmark'  it return this warning

```
WARNING: Can't mass-assign protected attributes for Bookmark: document_id, document_type''

korabh:


cbeer:
This doesn't seem right to me. I think we introduced `needs_attr_accessible?` because you can still use `attr_accessible` in Rails 4 (by introducing a dependency, I think).

jcoyne:
yes, I'm with @cbeer. We should not remove this bit of code.

korabh:
@cbeer @jcoyne Fine :+1:

PR #1226, opened 2015-06-25
Create a generic FacetPaginator class

rwd:
This creates a generic `Blacklight::FacetPaginator` for data source agnosticism and makes `Blacklight::Solr::FacetPaginator` a sub-class of it.

Generic functionality is pretty much everything that was in the Solr facet paginator, except for the sort option defaulting to "count" which is Solr-specific.

### DIFF
--- a/lib/blacklight/facet_paginator.rb
+++ b/lib/blacklight/facet_paginator.rb
@@ -1,0 +1,121 @@
+module Blacklight
+  # Pagination for facet values -- works by setting the limit to max
+  # displayable. You have to ask Solr for limit+1, to get enough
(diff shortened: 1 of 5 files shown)


lib/blacklight/facet_paginator.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    def next_page
+      current_page + 1 unless last_page?
+    end
+   
+    #@deprecated
+    def has_next?
+      !last_page?
+    end
+    deprecation_deprecate :has_next? => "use #!last_page?"
```
cbeer: Should we leave these deprecations over on `Blacklight::Solr::FacetPaginator`?

Issue #1204, opened 2015-05-26
Add Rubocop to Gemfile, versioned


rwd:
Trying to verify code syntax locally with `rubocop`, I'm hitting various errors with different versions of it due to missing cops: Style/ClosingParenthesisIndentation in 0.29.1; Style/Blocks in 0.31.0.

Can we have the actual version of Rubocop used in the CI tests added to the Gemfile so that developers can get the correct version installed?

cbeer:
Hound has updated their version of rubocop to something a little more modern.

PR #1238, opened 2015-08-25
Add Rubocop 0.29.1 to bundle

rwd:
Re #1204, this simply adds Rubocop 0.29.1 to the bundle so that developers can run the style checks before pushing their commits. Locked to version 0.29.1 as that is the version used by [Hound](https://github.com/thoughtbot/hound/blob/master/Gemfile).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,10 @@ gemspec path: File.expand_path('..', __FILE__)
 gem 'simplecov', '~> 0.7.1', require: false
 gem 'coveralls', require: false
 
+group :development do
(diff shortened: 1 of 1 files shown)


cbeer:
I'm not thrilled with pinning to an older version of Rubocop. They've added useful checks in recent releases and I'd hate to lose those.

jcoyne:
Hound now uses 0.34.2 https://github.com/thoughtbot/hound/blob/master/Gemfile#L32

cbeer:
:+1:

PR #1241, opened 2015-09-09
[WIP] Add support for Solr pagination cursor

cbeer:
Not for merging (yet.. if ever...), but stashing here in case it's useufl.

### DIFF
--- a/app/views/catalog/_paginate_compact.html.erb
+++ b/app/views/catalog/_paginate_compact.html.erb
@@ -3,4 +3,4 @@
 
         <%= render :partial => "paginate_compact", :object => @response
 -%>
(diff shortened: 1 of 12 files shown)


magibney:
To go on the record here, I think the main challenge in implementing currently is that `cursorMark`-based paging is forward-only ... so in order to support bidirectional navigation (prev/next), we'd have to maintain potentially an enormous stack of state.
My current thinking is that this would best wait for stateless `cursorMark`-style paging to be implemented in Solr, per the idea proposed [here](https://issues.apache.org/jira/browse/SOLR-6635?focusedCommentId=15755112#comment-15755112).

PR #1251, opened 2015-09-29
add index navigation for facet more links

peetucket:
This is a feature we recently added to the Revs Digital Library website.  For facets that have many many values, users click on the "More" link to see them all then have to page through results 20 at a time.  If there are thousands of values, it is helpful when in "A-Z" mode to have index navigation (i.e. filter by letter).  There are still limitations (i.e. you will see all letters even if some have no values, and it is case sensitive due to solr restrictions) but it is a start.

### DIFF
--- a/app/assets/stylesheets/blacklight/_facets.scss
+++ b/app/assets/stylesheets/blacklight/_facets.scss
@@ -148,6 +148,22 @@ ul.facet_extended_list, .facet_extended_list ul
   @extend .clearfix;
 }
 
+ 
+// Facet modal alphabetical filter
+.alpha-filter {
+  margin-bottom: 18px;
+  text-align: center;
+
+  a {
+    padding: 0 2px;
+    
+    &.active {
+      color: inherit;
+      font-weight: bold;
+    }
+  }
+}
+
 
 @media (max-width: $grid-float-breakpoint) {
 
--- a/app/views/catalog/_facet_pagination.html.erb
+++ b/app/views/catalog/_facet_pagination.html.erb
@@ -1,21 +1,20 @@
-    <div class="prev_next_links btn-group pull-left">
-      <%= link_to_previous_page @pagination, raw(t('views.pagination.previous')), :params => params, :param_name => blacklight_config.facet_paginator_class.request_keys[:page], :class => 'btn btn-link', :data => {:ajax_modal => "preserve"} do %>
-       <span class="disabled btn btn-disabled"><%= raw(t('views.pagination.previous')) %></span>
-     <% end %>
-      <%= link_to_next_page @pagination, raw(t('views.pagination.next')), :params => params, :param_name => blacklight_config.facet_paginator_class.request_keys[:page], :class => 'btn btn-link',  :data => {:ajax_modal => "preserve"} do %>
-       
-       <span class="disabled btn  btn-disabled"><%= raw(t('views.pagination.next')) %></span>
-      <% end %>
-     
-    </div>
+  <%= render partial: 'facet_index_navigation' %>
+  
+  <div class="prev_next_links btn-group pull-left">
+    <%= link_to_previous_page @pagination, raw(t('views.pagination.previous')), params: params, param_name: blacklight_config.facet_paginator_class.request_keys[:page], class: 'btn btn-link', data: {ajax_modal: "preserve"} do %>
+     <span class="disabled btn btn-disabled"><%= raw(t('views.pagination.previous')) %></span>
+   <% end %>
+    <%= link_to_next_page @pagination, raw(t('views.pagination.next')), params: params, param_name: blacklight_config.facet_paginator_class.request_keys[:page], class: 'btn btn-link',  data: {ajax_modal: "preserve"} do %>       
+     <span class="disabled btn  btn-disabled"><%= raw(t('views.pagination.next')) %></span>
+    <% end %>     
+  </div>
  
   <div class="sort_options btn-group pull-right">
     <% if @pagination.sort == 'index' -%>
       <span class="active az btn btn-default"><%= t('blacklight.search.facets.sort.index') %></span><%= link_to_unless(@pagination.sort == 'count', t('blacklight.search.facets.sort.count'), 
-        @pagination.params_for_resort_url('count', params), :class => "sort_change numeric btn btn-default",  
-        :data => {:ajax_modal => "preserve"}) %>
+        @pagination.params_for_resort_url('count', params_no_facet_filter), class: "sort_change numeric btn btn-default", data: {ajax_modal: "preserve"}) %>
     <% elsif @pagination.sort == 'count' -%>
-      <%=  link_to(t('blacklight.search.facets.sort.index'), @pagination.params_for_resort_url('index', params), 
-        :class => "sort_change az btn btn-default",  :data => {:ajax_modal => "preserve"}) %><span class="active numeric btn btn-default"><%= t('blacklight.search.facets.sort.count') %></span>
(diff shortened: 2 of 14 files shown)


app/views/catalog/_facet_pagination.html.erb
```diff
@@ -1,3 +1,13 @@
+    <% if @facet.index_pagination && params["facet.sort"] == 'index' %>
+      <div class="alpha-filter">
+        <%= t('blacklight.search.facets.filter') %>:
+        <%= link_to(t('blacklight.search.facets.clear'), @pagination.params_for_resort_url('index', params.except(:"facet.prefix")), :data => {:ajax_modal => "preserve"}) %>
```
cbeer: Should we hide this link when a prefix isn't selected?

app/views/catalog/_facet_pagination.html.erb
```diff
@@ -1,3 +1,13 @@
+    <% if @facet.index_pagination && params["facet.sort"] == 'index' %>
```
cbeer: I wonder if the index pagination controls should get pulled out to a separate partial.

cbeer: Instead of checking the URL params directly, can we check if the facet field is sorted by index? Perhaps using:

https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/solr_response/facets.rb#L41-L43

lib/blacklight/catalog.rb
```diff
@@ -79,8 +79,11 @@ def track
 
     # displays values and pagination links for a single facet field
     def facet
+      
+      extra_controller_params = (params[:"facet.prefix"] ? {"facet.prefix"=>params[:"facet.prefix"]} : {})
```
cbeer: Can we push this logic down into the `Blacklight::SearchHelper` or part of `Blacklight::Solr::RequestBuilder`?

https://github.com/projectblacklight/blacklight/blob/6abc7e99c28d93490109acac47b0b63623901ca7/lib/blacklight/search_helper.rb#L94

spec/features/facets_spec.rb
```diff
@@ -29,9 +29,29 @@
...
-    
   end
+  
+  it "should be able to sort more facet window by letter" do
+    visit catalog_facet_path("subject_topic_facet")
+    within ".modal-footer" do
+      click_on "A-Z Sort"
+    end
+    expect(page).to have_selector '.facet-values li:first', text: "Accident insurance"
+    expect(page).to have_css('.facet-values li',count: 20)
```
cbeer: Can you add a space after the commas here?

Can you also pick whether to use parenthesis around arguemtns or not, and stick to it?

config/locales/blacklight.en.yml
```diff
@@ -183,6 +183,8 @@ en:
         counter: '%{counter}. '
       facets:
         title: 'Limit your search'
+        filter: 'Filter'
+        clear: 'Clear'
```
cbeer: Before merging this, we should get translations for our other languages. It might be as easy as pulling equivalent values out of other fields..

app/views/catalog/_facet_pagination.html.erb
```diff
@@ -12,10 +22,10 @@
   <div class="sort_options btn-group pull-right">
     <% if @pagination.sort == 'index' -%>
       <span class="active az btn btn-default"><%= t('blacklight.search.facets.sort.index') %></span><%= link_to_unless(@pagination.sort == 'count', t('blacklight.search.facets.sort.count'), 
-        @pagination.params_for_resort_url('count', params), :class => "sort_change numeric btn btn-default",  
+        @pagination.params_for_resort_url('count', params.except(:"facet.prefix")), :class => "sort_change numeric btn btn-default",
```
cbeer: We should pull `params.except(:"facet.prefix")` up to a helper method.

app/views/catalog/_facet_pagination.html.erb
```diff
@@ -1,3 +1,13 @@
+    <% if @facet.index_pagination && params["facet.sort"] == 'index' %>
+      <div class="alpha-filter">
+        <%= t('blacklight.search.facets.filter') %>:
+        <%= link_to(t('blacklight.search.facets.clear'), @pagination.params_for_resort_url('index', params.except(:"facet.prefix")), :data => {:ajax_modal => "preserve"}) %>
+        <% ('A' .. 'Z').to_a.each do |letter| %>
+          <%=  link_to(letter, @pagination.params_for_resort_url('index', params.merge(:"facet.prefix"=>letter)), :data => {:ajax_modal => "preserve"}, :class=>(params["facet.prefix"]==letter ? 'active' : '')) %>
```
cbeer: Can you remove the extra space after `<%=`?

app/views/catalog/_facet_pagination.html.erb
cbeer: Can you convert these to use Ruby 1.9-style hashes? While there are 1.8-style hashes around the project, we've been trying to migrate them to the new style as those lines of code are touched.

app/views/catalog/_facet_pagination.html.erb
```diff
@@ -1,3 +1,13 @@
+    <% if @facet.index_pagination && params["facet.sort"] == 'index' %>
+      <div class="alpha-filter">
+        <%= t('blacklight.search.facets.filter') %>:
```
cbeer: I /think/ the `:` should be part of the i18n label. We should also wrap the label and content in some element (`span` or `div` or otherwise?) so they can be styled downstream (or, for RTL languages, swapped around)

app/views/catalog/_facet_pagination.html.erb
cbeer: Maybe this should be wrapped in an `<ol>`?

app/views/catalog/_facet_pagination.html.erb
```diff
@@ -1,3 +1,13 @@
...
+        <%= link_to(t('blacklight.search.facets.clear'), @pagination.params_for_resort_url('index', params.except(:"facet.prefix")), :data => {:ajax_modal => "preserve"}) %>
+        <% ('A' .. 'Z').to_a.each do |letter| %>
```
cbeer: I think this range should be part of the facet configuration. It seems like it's both language and data dependent.

Also, do you know if `facet.prefix` is case-sensitive?

app/assets/stylesheets/blacklight/_facets.scss
```diff
@@ -148,6 +148,22 @@ ul.facet_extended_list, .facet_extended_list ul
   @extend .clearfix;
 }
 
+ 
+// Facet modal alphabetical filter
+.alpha-filter {
+  margin-bottom: 18px;
```
cbeer: `18px` seems oddly specific. Are we missing a native bootstrap class that would provide this? Or is there a bootstrap-provided scss variable we should use instead?

app/assets/stylesheets/blacklight/_facets.scss
```diff
@@ -148,6 +148,22 @@ ul.facet_extended_list, .facet_extended_list ul
...
+  text-align: center;
+
+  a {
```
cbeer: Same comment about this padding. Maybe there's a good bootstrap way to achieve the same effect?

app/views/catalog/_facet_pagination.html.erb
peetucket: yes, facet.prefix is case sensitive, both the parameter itself and its value.  it is a solr feature to allow facet filtering, and  i could not find an easy way to have it be case insensitive

cbeer: ok, then it seems doubly important to make this range configurable.

PR #1253, opened 2015-10-05
Extract #render_values presenter method for rendering multi-valued data

cbeer:


lib/blacklight/document_presenter.rb
```diff
@@ -87,7 +87,26 @@ def render_field_value value=nil, field_config=nil
...
-      safe_join(safe_values, (field_config.separator if field_config) || field_value_separator)
+      render_values(safe_values, field_config)
+    end
+
+    ##
+    # Render a fields values as a string
+    # @param [Array] values to display
+    # @param [Blacklight::Solr::Configuration::Field] solr field configuration
+    # @return [String]
+    def render_values(values, field_config = nil)
```
cbeer: I'm looking for a better name for this method, but haven't thought up anything yet...

lib/blacklight/document_presenter.rb
```diff
@@ -87,7 +87,26 @@ def render_field_value value=nil, field_config=nil
...
+      options = {}
+      options = field_config.separator_options if field_config && field_config.separator_options
+
+      if field_config && field_config.separator
+        Deprecation.warn(self.class, 'The field configuration #separator is deprecated. Use #separator_options instead')
```
cbeer: We should backport this commit to 5.x and remove the deprecated behavior before 6.x

PR #1254, opened 2015-10-06
use {!term} localparam instead of {!raw}

cfinucane:
Facet drill-downs weren't working for me as-is with Solr 5.3.0, and https://wiki.apache.org/solr/QueryParser says "Note: use 'term' for this in Solr 4.0 or above", which fixes it.

Sorry if I'm missing something here.


cfinucane:
OK, I clearly should have looked into this a little more.  Closing because I didn't even look at the tests.  My bad!

cbeer:
@cfinucane i think you're on the right track. We were using `!raw` for compatibility with older versions of Solr, but with Blacklight 6.x, I think we should feel free to drop those Solrs.

PR #1256, opened 2015-10-07
Custom facet value nil exception

peetucket:
Proposed fix for projectblacklight#1255 including 2 new tests.

### DIFF
--- a/lib/blacklight/solr/search_builder_behavior.rb
+++ b/lib/blacklight/solr/search_builder_behavior.rb
@@ -282,7 +282,12 @@ module Blacklight::Solr
 
       case
         when (facet_config and facet_config.query)
-          facet_config.query[value][:fq]
+          if facet_config.query[value]
+            facet_config.query[value][:fq]
+          else
+            # exclude all documents if the custom facet key specified was not found
+            '-*:*'
+          end
         when (facet_config and facet_config.date)
           # in solr 3.2+, this could be replaced by a !term query
           "#{prefix}#{solr_field}:#{RSolr.solr_escape(value)}"
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -69,7 +69,19 @@ describe CatalogController do
           expect(facet.items).to be_empty
         end
       end
+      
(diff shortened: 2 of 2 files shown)


lib/blacklight/solr/search_builder_behavior.rb
```diff
@@ -282,7 +282,7 @@ def facet_value_to_fq_string(facet_field, value)
 
       case
         when (facet_config and facet_config.query)
-          facet_config.query[value][:fq]
+          facet_config.query[value].blank? ? '-*' : facet_config.query[value][:fq]
```
cbeer: Is `-*` actually a Solr query? I think you might mean `-*:*`?

cbeer: Do you think this can this be simplified any?

`facet_config.query[value].blank? ? '-*' : facet_config.query[value][:fq]`

peetucket: 1. yup -*:* is what i meant   will fix
2. i thought that was pretty succint, any suggestions?

cbeer: Sorry, I wasn't thinking about succinctness, more about clarity and understandability. Maybe something like this?

```
if facet_config.query[value]
  facet_config.query[value][:fq]
else
  # exclude all documents for the bogus facet query key
  '-*:*'
end

peetucket: yeah, ok, more readable to humans, will update

cbeer:
Thanks. Want to squash your commits?

peetucket:
Commits squashed

cbeer:
Can you update the commit message to describe this change a little better?

cbeer:
:+1:

peetucket:
Thanks.  Quick question, will this get cherry-picked into v6.x or does that need a separate request?

cbeer:
Yes, I merged it into master in #1257. In the future, it would be better if you targeted pull requests at `master` and we backported it from there.

peetucket:
ok.  good to know, thanks.

PR #1259, opened 2015-10-15
Extract path helpers from helpers to a class

jcoyne:


app/helpers/blacklight/url_helper_behavior.rb
```diff
@@ -144,142 +144,14 @@ def link_to_previous_search(params)
...
-    request_keys = blacklight_config.facet_paginator_class.request_keys
-    new_params.except! *request_keys.values
-
-    new_params
+  # Get url parameters to a search within a grouped result set
+  #
+  # @param [Blacklight::SolrResponse::Group]
+  # @return [Hash]
+  def add_group_facet_params_and_redirect group
+    add_facet_params_and_redirect(group.field, group.key, params)
```
jcoyne: needs `current_path`

app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -137,20 +137,28 @@ def render_facet_value(facet_field, item, options ={})
...
   # @params (see #render_facet_value)
   def render_selected_facet_value(facet_field, item)
-    content_tag(:span, :class => "facet-label") do
-      content_tag(:span, facet_display_value(facet_field, item), :class => "selected") +
+    remove_href = search_action_path(current_path.remove_facet_params(facet_field, item))
+    content_tag(:span, class: "facet-label") do
+      content_tag(:span, facet_display_value(facet_field, item), class: "selected") +
       # remove link
-      link_to(content_tag(:span, '', :class => "glyphicon glyphicon-remove") + content_tag(:span, '[remove]', :class => 'sr-only'), search_action_path(remove_facet_params(facet_field, item, params)), :class=>"remove")
+      link_to(remove_href, class: "remove")do
```
jcoyne: add a space before `do`

lib/blacklight/path.rb
```diff
@@ -0,0 +1,129 @@
+module Blacklight
+  class Path
```
jcoyne: Add class documentation: This class decorates ActiveController::Params to provide methods to operate upon the parameters that serialize the blacklight state, namely: `:f, :q, :page, :per_page and, :sort`

app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -136,20 +136,28 @@ def render_facet_value(facet_field, item, options ={})
...
     if facet_config.url_method
-      path = send(facet_config.url_method, facet_field, item)
+      send(facet_config.url_method, facet_field, item)
     else
-      path = search_action_path(add_facet_params_and_redirect(facet_field, item))
+      search_action_path(current_path.add_facet_params_and_redirect(facet_field, item))
     end
   end
 
+  def current_path
```
cbeer: Is the `UrlHelperBehavior` module a better place for this method?

Should this be memoized at all?

I was confused, at first, if `current_path` was a built-in Rails helper. I wonder if there's a better name (although I can't think of any off-hand)

lib/blacklight/path.rb
```diff
@@ -0,0 +1,131 @@
+module Blacklight
+  # This class encapsulates the search state as represented by the query
+  # parameters namely: :f, :q, :page, :per_page and, :sort
+  class Path
```
cbeer: Should this go in `app/models/blacklight` now? Happy to be convinced either way.

lib/blacklight/path.rb
```diff
@@ -0,0 +1,131 @@
...
+    include Blacklight::Facet
```
cbeer: Is this just for `facet_configuration_for_field`? I wonder if we should push that up into `Blacklight::Configuration` (now, or in a future PR)

app/helpers/blacklight/facets_helper_behavior.rb
jcoyne: I'm fine with moving it to `UrlHelperBehavior` I put it here, because this is where all that functionality was extracted from.

It can be memoized, there is not reason not to.

I'll accept any name you suggest, this was the best I had.

cbeer:
:+1: I don't feel too strongly about any of my comments/questions. I think this is a good refactor and probably sets us up to clean up even more (`Blacklight::FacetPaginator`, for one).

app/helpers/blacklight/facets_helper_behavior.rb
cbeer: :+1: 

:ok_hand: if it's easy. it doesn't look like the class is much overhead, but it probably gets called a fair bit when rendering facets.

I'll think on it some more, but that shouldn't stop us from shipping.

lib/blacklight/path.rb
jcoyne: No, this is not a model in the sense that it doesn't represent a business object in a persistence store.

lib/blacklight/path.rb
jcoyne: Yes, I believe that is all.  I agree that belongs in the configuration. I don't want to keep mixing this in.

lib/blacklight/path.rb
mejackreed: ^^ I like this definition of what goes under `app/models` @jcoyne ++

PR #1262, opened 2015-10-17
Add facet prefix filtering when paginating within a facet

cbeer:
Fixes #1251. I took @peetucket's commit and tried to push logic down into the models and use some more bootstrap-y styling:

screen shot 2015-10-17 at 9 36 06 am

There's significant work left to make it a general purpose tool (say, disabling the prefixes that have 0 hits), but this may be good enough for some targeted use or in a very large index.

### DIFF
--- a/lib/generators/blacklight/templates/catalog_controller.rb
+++ b/lib/generators/blacklight/templates/catalog_controller.rb
@@ -53,9 +53,14 @@ class <%= controller_name.classify %>Controller < ApplicationController
     #
     # :show may be set to false if you don't want the facet to be drawn in the 
     # facet bar
+    
+    # set :index_pagination to true if you want the pop-up window for large facets to have alphabetical index navigation
+    #  (useful when user clicks "more" on a large facet and wants to navigate alphabetically across a large set of results)
+    # if you set :index_pagination to true, you MUST set :index_range to an array of characters that will be used to create the navigation (note: It is case sensitive when searching values)
+    
     config.add_facet_field 'format', :label => 'Format'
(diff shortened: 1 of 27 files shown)


lib/generators/blacklight/templates/catalog_controller.rb
```diff
@@ -53,9 +53,14 @@ class <%= controller_name.classify %>Controller < ApplicationController
     #
     # :show may be set to false if you don't want the facet to be drawn in the 
     # facet bar
+    
+    # set :index_pagination to true if you want the pop-up window for large facets to have alphabetical index navigation
+    #  (useful when user clicks "more" on a large facet and wants to navigate alphabetically across a large set of results)
+    # if you set :index_pagination to true, you MUST set :index_range to an array of characters that will be used to create the navigation (note: It is case sensitive when searching values)
+
```
mejackreed: The comment says to set `:index_pagination => true` but in the example `:index_pagination` is not set.

mejackreed: Could case sensitivity be an option here?

cbeer: Thanks for catching that. I've tried to revise the comment to reflect the new configuration.

Solr's `facet.prefix` option is case-sensitive. It looks like, in Solr 5.1, they landed a new [`facet.contains`](https://cwiki.apache.org/confluence/display/solr/Faceting#Faceting-Thefacet.containsParameter) parameter that may do case-insensitive filtering, so maybe there's hope for this in the future..

mejackreed: :+1:

PR #1265, opened 2015-10-17
Reorganize lib/blacklight to move classes to ./app/ as appropriate

cbeer:
Thoughts?

- [x] move controller-level mixins to `app/controllers/concerns`
- [x] move model-level mixins to `app/models/concerns`
- [x] move models to `app/models/blacklight`

Still TODO, but maybe after shipping this:

- [ ] disentangle mixins that get mixed into both controllers and models (`Blacklight::Configurable` is the biggest offender)
- [ ] disentangle mixins that are mixed into controller and helpers (and just use helper_class at the controller level) (`Blacklight::Facet`, which may only be mixed in to make our testing lives easier)
- [ ] backport (post-merge) any necessary deprecation warnings (`Blacklight::SolrRepository` => `Blacklight::Solr::Repository`, `Blacklight::SolrResponse` => `Blacklight::Solr::Response`)


cbeer:
I think the decrease in coverage is just from removing those autoload lines, not a real loss of test coverage.

mejackreed:
The rest of this for 6.0?

PR #1271, opened 2015-10-19
Test with JRuby 9k

cjcolvar:
Configure travis to test with JRuby 9.0.1.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,11 @@ matrix:
       env: "RAILS_VERSION=4.1.13"
     - rvm: 2.1.6
       env: "RAILS_VERSION=4.2.4"
-  allow_failures:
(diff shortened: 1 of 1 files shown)


cbeer:
:+1:

cjcolvar:
JRuby 9.0.0.0.pre1 (which travis was using when specifying jruby-9.0.0.0) couldn't even start the tests due to an error when trying to run coveralls.  So I added JRuby 9.0.1.0 and marked it as allow_failures since it runs really slowly and eventually times out from inactivity. Maybe we should turn on fast_finish (http://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing)?  Maybe travis needs to be configured in a different way to speed things up?

PR #1275, opened 2015-10-23
Use facet_field_label for rendering filters with I18n support on sear…

jipe:
…ch history.

- Add test coverage for I18n-labeled facets
- Update existing test for undefined facets to expect label
  to be humanized facet name instead of a specific string value


jipe:
Well... to say that I added test coverage for I18n-labeled facets is maybe stretching it a bit. The work I did was in the context of search history and making sure the filters are displayed with the correct I18n label.

cbeer:
Thanks!

Issue #1218, opened 2015-06-11
Take advantage of Rails 4.x features


cbeer:
After #1216 is merged, there are some Rails 4 features we should look at, including:

- [x] routing concerns (fixing #1200 ?)
- [x] GlobalID support for `Blacklight::Document` #1120 

- declarative etags (add user and controller information to the etag)

PR #1277, opened 2015-11-04
Use resourceful routing

### LINKED ISSUE #1200: CatalogController subclasses don't have access to all their routes by default
We're working on an engine that subclasses the `CatalogController`, and we recently attempted to add "more" links to facets based on a user request. However, when we first enabled it, we got an `ActionView::Template::Error` [(stacktrace here)](https://gist.github.com/anarchivist/6a37894d84d07425e136). 

@terrellt suggested that this was an issue he ran into before, and suggested adding the following workaround in our engine's `config/routes.rb`:

```ruby
Krikri::Engine.routes.draw do
  # ...snip...
  resources :records, only: [:index, :show] do
    collection do
      get "facet/:id", :to => "records#facet"
    end
  end
# ...snip...
end
```

cbeer:
This allows applications to use Rails 4's routing concerns to create Blacklight routes.

Here is an example of parity with existing routes:

```ruby
  concern :searchable, Blacklight::Routes::Searchable.new
  concern :exportable, Blacklight::Routes::Exportable.new

  resource :catalog, only: [:index] do
    concerns :searchable
  end

  resources :solr_documents, only: [:show], controller: 'catalog' do
    concerns :exportable
  end
```

This allows downstream applications to easily use Rails routing to configure their routes, e.g.:

```ruby
  resource :catalog, only: [:index], path: '/view' do
    concerns :searchable
  end

   # search_catalog GET|POST /view(.:format)                   catalogs#index
   # track_catalog POST     /view/:id/track(.:format)         catalogs#track
   # opensearch_catalog GET      /view/opensearch(.:format)        catalogs#opensearch
   # facet_catalog GET      /view/facet/:id(.:format)         catalogs#facet
```

or easily have multiple blacklight catalog controllers:

```ruby
  resource :image_catalog do
    concerns :searchable
  end

  resource :video_catalog do
    concerns :searchable
  end
```

Part of #1218 
Fixes #1200


jcoyne:
I really like this. :+1:  Needs a rebase though.

jkeck:
:+1: for routing concerns

jcoyne:
... and I merged another PR first, so rebase again.

cbeer:
Rebased. I think the loss of coverage is just removing well-covered lines of code.

PR #1280, opened 2015-11-04
Next/previous buttons should maintain the current search session

cbeer:


jcoyne:
Should this get backported?

cbeer:
Probably. I guess we should bisect releases and figure out where it got broke. We must have implicitly called `#current_search_session` from the `catalog#track` action, and at some point stopped doing that.

PR #1288, opened 2015-11-05
Adds autocomplete typeahead feature

mejackreed:
Using twitter-typeahead, with a default solr schema suggester.

Had to punt on feature specs as I couldn't get it working

### DIFF
--- a/app/assets/javascripts/blacklight/blacklight.js
+++ b/app/assets/javascripts/blacklight/blacklight.js
@@ -11,6 +11,7 @@
 //= require blacklight/search_context
 //= require blacklight/collapsable
 //= require blacklight/facet_load
+//= require blacklight/autocomplete.js
 //
 //Bootstrap JS for providing collapsable tablet/mobile menu/alert boxes
 //= require bootstrap/transition
@@ -18,6 +19,9 @@
 //= require bootstrap/dropdown
 //= require bootstrap/alert
 //= require bootstrap/modal
+
+// Twitter Typeahead for autocomplete
+//= require twitter/typeahead
 
 /* Blacklight has a Javascript setup meant to support local disabling, 
   modification, and use of Blacklight behaviors. 
--- a/app/assets/javascripts/blacklight/blacklight.js
+++ b/app/assets/javascripts/blacklight/blacklight.js
@@ -11,7 +11,7 @@
 //= require blacklight/search_context
 //= require blacklight/collapsable
 //= require blacklight/facet_load
-//= require blacklight/autocomplete.js
+//= require blacklight/autocomplete
 //
 //Bootstrap JS for providing collapsable tablet/mobile menu/alert boxes
 //= require bootstrap/transition
--- a/app/assets/stylesheets/blacklight/_twitter_typeahead.scss
+++ b/app/assets/stylesheets/blacklight/_twitter_typeahead.scss
@@ -11,10 +11,9 @@
     width: 100%;
   }
 
-  .tt-dropdown-menu {
+  .tt-menu {
     @extend .dropdown-menu;
     font-family: $font-family-base;
-    
     width: 100%;
   }
 
@@ -24,8 +23,8 @@
     text-decoration: none;
   }
 
-  .tt-suggestion p {
+  .tt-suggestion {
     font-size: 14px;
-    padding-left: 10px;
+    padding: 5px 5px 5px 10px;
   }
 }
--- a/app/assets/stylesheets/blacklight/_twitter_typeahead.scss
+++ b/app/assets/stylesheets/blacklight/_twitter_typeahead.scss
@@ -1,0 +1,31 @@
+.twitter-typeahead {
+  float: left;
+  width: 100%;
+  z-index: 10000;
+  
+  .tt-input.form-control {
+    width: 100%;
+  }
+  
+  .tt-hint.form-control {
+    width: 100%;
+  }
+
+  .tt-dropdown-menu {
+    @extend .dropdown-menu;
+    font-family: $font-family-base;
+    
+    width: 100%;
+  }
+
+  .tt-cursor {
+    background-color: $dropdown-link-hover-bg;
+    color: $dropdown-link-hover-color;
+    text-decoration: none;
+  }
+
+  .tt-suggestion p {
+    font-size: 14px;
+    padding-left: 10px;
+  }
+}
--- a/app/controllers/concerns/blacklight/suggest.rb
+++ b/app/controllers/concerns/blacklight/suggest.rb
@@ -1,0 +1,24 @@
+module Blacklight
+  module Suggest
+    extend ActiveSupport::Concern
+
+    included do
+      include Blacklight::Configurable
+      include Blacklight::SearchHelper
+
+      copy_blacklight_config_from(CatalogController)
+    end
+
+    def index
+      respond_to do |format|
+        format.json do
+          render json: suggestions_service.suggestions
+        end
+      end
+    end
+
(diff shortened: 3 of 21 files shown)


app/controllers/concerns/blacklight/suggest/suggest_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Suggest
+    module SuggestHelper
+      extend ActiveSupport::Concern
+      include Blacklight::SearchHelper
+
+      ##
+      # For now, only use the q parameter to create a
+      # Blacklight::Suggest::Response
+      # @param [Hash] params
+      # @return [Earthworks::Suggest::Response]
```
jcoyne: Incorrect return type.

app/controllers/concerns/blacklight/suggest/suggest_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+      def get_suggestions(params)
```
jcoyne: rename to `suggestions`

app/controllers/concerns/blacklight/suggest/suggest_helper.rb
```diff
@@ -0,0 +1,32 @@
+module Blacklight
+  module Suggest
+    module SuggestHelper
```
jcoyne: Why do we need this module? Can it be merged into Blacklight::Suggest?

blacklight.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rsolr",     "~> 1.0.11"  # Library for interacting with rSolr.
   s.add_dependency "bootstrap-sass", "~> 3.2"
   s.add_dependency "deprecation"
+  s.add_dependency "twitter-typeahead-rails"
```
jcoyne: Add a version specifier

lib/generators/blacklight/templates/catalog_controller.rb
```diff
@@ -180,6 +180,10 @@ class <%= controller_name.classify %>Controller < ApplicationController
     # If there are more than this many search results, no spelling ("did you 
     # mean") suggestion is offered.
     config.spell_max = 5
+
+    # Configuration for autocomplete suggestor
+    config.autocomplete_enabled = true
```
jcoyne: Can we just say that having a non-nil `autocomplete_path` implies that autocomplete is enabled?

app/assets/javascripts/blacklight/blacklight.js
```diff
@@ -11,6 +11,7 @@
 //= require blacklight/search_context
 //= require blacklight/collapsable
 //= require blacklight/facet_load
+//= require blacklight/autocomplete.js
```
jcoyne: I don't think the `.js` is required

app/controllers/concerns/blacklight/suggest.rb
```diff
@@ -0,0 +1,15 @@
+module Blacklight
+  module Suggest
+    extend ActiveSupport::Concern
+    include Blacklight::Suggest::SuggestHelper
+
+    def index
+      @response = get_suggestions params
```
jcoyne: Why is `@response` an instance variable?  Can it be a local?

app/controllers/concerns/blacklight/suggest/suggest_helper.rb
mejackreed: It could be. I wanted to separate the action methods from the concern of methods that "do more stuff"... Prefer to put it all in one place?

app/assets/stylesheets/blacklight/_twitter_typeahead.scss
```diff
@@ -0,0 +1,31 @@
...
+    width: 100%;
+  }
+
+  .tt-cursor {
+    background-color: $dropdown-link-hover-bg;
+    color: $dropdown-link-hover-color;
+    text-decoration: none;
+  }
+
+  .tt-suggestion p {
```
mejackreed: :facepunch:

lib/generators/blacklight/templates/catalog_controller.rb
mejackreed: Yeah I thought about this, but I left with the specified enabled, because you may have this defined but not enabled? The configuration felt different from enabling it. Not feeling too strongly about this, but seems more in your face way to disable it.

app/controllers/concerns/blacklight/suggest/suggest_helper.rb
jcoyne: I'd say either put it in one place or make it a class that gets instantiated. IMHO a mixin just makes it harder to read.

app/models/concerns/blacklight/suggest/response.rb
```diff
@@ -0,0 +1,26 @@
...
+        @response = response
+        @request_params = request_params
+        @suggest_path = suggest_path
+      end
+
+      ##
+      # Trys the suggestor response to return suggestions if they are
+      # present
+      # @return [Array]
+      def suggestions
```
cbeer: Is it possible to build this into the normal `Blacklight::Solr::Response`?

app/controllers/concerns/blacklight/suggest/suggest_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+        request_params = { q: params[:q] }
+        Blacklight::Suggest::Response.new suggest_results(request_params), request_params, suggest_handler_path
+      end
+
+      ##
+      # Query the suggest handler using RSolr::Client::send_and_receive
+      # @param [Hash] request_params
+      # @return [RSolr::HashWithResponse]
+      def suggest_results(request_params)
+        repository.connection.send_and_receive(suggest_handler_path, params: request_params)
```
cbeer: Do suggestions bypass any access control mechanisms? I'd guess they do, because they're repository wide, but want to make sure (and maybe we should document that as part of the configuration?)...

app/controllers/concerns/blacklight/suggest/suggest_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+      include Blacklight::SearchHelper
+
+      ##
+      # For now, only use the q parameter to create a
+      # Blacklight::Suggest::Response
+      # @param [Hash] params
+      # @return [Blacklight::Suggest::Response]
+      def suggestions(params)
+        request_params = { q: params[:q] }
+        Blacklight::Suggest::Response.new suggest_results(request_params), request_params, suggest_handler_path
```
cbeer: I wonder if some of this can use the `Blacklight::SearchBuilder` framework (so we could possibly support suggestions for e.g. Elasticsearch too)?

flyingzumwalt:
@mejackreed It occurred to me that you might want to use some of the documentation from https://github.com/flyingzumwalt/blacklight_suggest/blob/master/README.md I'm going to throw away the gem since it's redundant now (due to this PR) but it's basically the same code that's in this PR so the stuff in the README might be useful somewhere. Take a look.

mejackreed:
Thanks @flyingzumwalt ... i'll take a look and see where we can use that to improve the docs/wiki!

ronan-mch:
Just to note that the maintenance status of the Twitter Typeahead plugin seems a bit unclear. There's a thread on the GitHub with [a number of people complaining about lack of maintenance](https://github.com/twitter/typeahead.js/issues/1437) and the last commit to master was over six months ago.

mejackreed:
Good to note about that fork. Thanks!

PR #1300, opened 2015-11-06
rename blacklight_path to search_state

jcoyne:
blacklight_path is already used by a built in Rails routing helper, and
the routing helper was getting precidence over our method.


cam156:
:+1:  Looks good to me

PR #1304, opened 2015-11-06
Vagrantfile and bash provisioning script

jronallo:


provision.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+sudo yum install -y epel-release
+sudo yum install -y curl git java-1.7.0-openjdk-devel nodejs yum-utils gcc bzip2 kernel-devel dkms
```
jcoyne: Is java 8 available on centos?

hortongn:
Centos only has Java 6 and 7 when installed through packages.  Java 8 requires and manual download and install.

jcoyne:
Wow. That seems like a pretty insecure default.

PR #1305, opened 2015-11-06
Restore fallback to Blacklight::Solr::SearchBuilder, which will be re…

cbeer:
…moved in Blacklight 6


jcoyne:
Why?

cbeer:
This is just for Blacklight 5.x. It shouldn't have been removed yet, it's a backwards compatibility concern.

jcoyne:
:ok_hand:

PR #1317, opened 2015-11-09
Use Jbuilder to build json responses

jcoyne:


app/presenters/blacklight/json_presenter.rb
```diff
@@ -0,0 +1,42 @@
...
+       :limit_value, :offset_value, :total_count,
+       :first_page?, :last_page?].each do |k|
+        h[k] = @response.send(k)
+      end
+
+      h
+    end
+  end
+end
+
```
cbeer: Want to fix this extra line?

app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -183,26 +185,6 @@ def render_document_export_format format_name
       render text: @response.documents.map { |x| x.export_as(format_name) if x.exports_as? format_name }.compact.join("\n"), layout: false
     end
 
-    # override this method to change the JSON response from #index
```
cbeer: Should we deprecate these methods at all (either here, or back in 5.x)?

jcoyne: Does it make sense to do so?  Was anyone ever doing anything with this? Can we just add this as a breaking change in the release notes?

PR #1323, opened 2015-11-09
DeprecatedConstantProxy doesn't work with classes; do manual deprecations instead

cjcolvar:
This was causing blacklight_folders to break.

### DIFF
--- a/lib/blacklight.rb
+++ b/lib/blacklight.rb
@@ -23,18 +23,20 @@ module Blacklight
     attr_accessor :solr, :solr_config
   end
 
-  # For deprecating SolrRepository and SolrResponse
-  def self.const_missing(const_name)
-    case const_name
-    when :SolrRepository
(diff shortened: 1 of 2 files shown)


lib/blacklight.rb
```diff
@@ -25,6 +23,20 @@ class << self
     attr_accessor :solr, :solr_config
   end
 
+  # For deprecating SolrRepository and SolrResponse
+  def self.const_missing(const_name)
+    case const_name
```
cbeer: Instead of using `const_missing`, could we define and deprecate them like we did for modules?

https://github.com/projectblacklight/blacklight/commit/3ed993a53d8f9c2815284bc354f6c8a113cb5630

cjcolvar:
I had to lock the version of tins to 1.6.0 to avoid requiring ruby >= 2.0.  I created an issue for coveralls-ruby: https://github.com/lemurheavy/coveralls-ruby/issues/89

PR #1327, opened 2015-11-11
Add a CSS class and data attribute for the position of the document in the search results

cbeer:
This should be useful in a couple contexts:

- styling the top result differently
- relevancy analytics


mejackreed:
Not opposed here, but is `:first-child` insufficient?

https://developer.mozilla.org/en-US/docs/Web/CSS/%3Afirst-child

cbeer:
It is; e.g. if you're on page 2 of the results, `:first-child` will be the 11th record.

jcoyne:
This is sort of orthagonal and probably too much, but I was wondering if we shouldn't do search results like this:
```html
<ol start="50">
   <li>A result</li>
   <li>Another result</li>
   ...
</ol>
```

PR #1332, opened 2015-11-13
Non image objects in search results

peetucket:
Instead of rendering a blank HTML image tag when there is no thumbnail for an object, just don't render anything.

### DIFF
--- a/app/helpers/blacklight/catalog_helper_behavior.rb
+++ b/app/helpers/blacklight/catalog_helper_behavior.rb
@@ -37,7 +37,7 @@ module Blacklight::CatalogHelperBehavior
     else
       collection.total_count
     end
-      
+
     case collection.total_count
       when 0; t('blacklight.search.pagination_info.no_items_found', :entry_name => entry_name ).html_safe
       when 1; t('blacklight.search.pagination_info.single_item_found', :entry_name => entry_name).html_safe
@@ -54,7 +54,7 @@ module Blacklight::CatalogHelperBehavior
     offset ||= @response.start if @response
     offset ||= 0
 
-    unless render_grouped_response? 
+    unless render_grouped_response?
       idx + 1 + offset
     end
   end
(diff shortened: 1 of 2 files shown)


app/helpers/blacklight/catalog_helper_behavior.rb
```diff
@@ -172,12 +172,13 @@ def has_thumbnail? document
...
   # Render the thumbnail, if available, for a document and
   # link it to the document record.
-  # 
+  #
   # @param [SolrDocument]
   # @param [Hash] options to pass to the image tag
   # @param [Hash] url options to pass to #link_to_document
   # @return [String]
   def render_thumbnail_tag document, image_options = {}, url_options = {}
+    return "" unless has_thumbnail?(document) # do not render any html if there is no thumbnail
```
jkeck: In the view we check the `has_thumbnail?` helper and the truthy-ness of the return value of this method (`""` is truthy).  Now we're also checking the `has_thumbnail?` helper in this helper as well.

I feel like we should reconcile this somehow so we're not redundantly checking the `has_thumbnail?` helper and ensure we are not checking the truthy-ness of an empty string.

Just my $.02

https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/_thumbnail_default.html.erb#L1

peetucket: interesting, the issue i was trying to fix was in the gallery view, which comes from https://github.com/projectblacklight/blacklight-gallery 

so..this issue is currently not in the other search result views (which is explained by that method).  

So I think we have an option to do something very similar in the gallery gem (check for has_thumbnail?) before rendering. 

The one downside is that we ever decide to actually render something like a placeholder, we may actually want to do it in the the "render_thumbnail_tag" method itself and *remove* the check for has_thumbnail? in https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/_thumbnail_default.html.erb#L1

peetucket: sorry for all the other blank space diff noise, atom seems to remove those on save? not sure where it is happening

peetucket:
references sul-dlss/sul_exhibits_template#97

PR #1333, opened 2015-11-15
When rendering a field through a configured helper method, supply the…

### LINKED ISSUE #1331: Can we get a feature to add custom arguments to :helper_methods?
When adding index or show fields it would be nice if we could also pass in some custom arguments to the method (that get appended with document, field, and value). 

For example... I have an index field that is very large coming back from SOLR, I would like to pass it to a helper method that truncates that value, however, I would like to be able to pass in different "max_lengths" to that function. Currently, I have to define a helper_method for each field.

cbeer:
… field configuration as an additional option.

Fixes #1331


ismail44:
Awesome!!! Thanks!

PR #1336, opened 2015-11-16
Extract generators for solr 4 and 5, and default to installing solr 5

### LINKED ISSUE #1312: Update application generator to supply Solr 5

cbeer:
Fixes #1312


lib/generators/blacklight/install_generator.rb
```diff
@@ -15,7 +16,6 @@ class Install < Rails::Generators::Base
     desc """
   This generator makes the following changes to your application:
    1. Generates blacklight:models
-   2. Adds rsolr to the Gemfile
    3. Adds globalid to the Gemfile
```
jcoyne: numbering issue. Should be 2.

PR #1339, opened 2015-11-17
Use Time.zone.now instead of Time.now to correctly use the application time

cbeer:


app/views/catalog/_document_default.atom.builder
```diff
@@ -3,7 +3,7 @@ xml.entry do
   xml.title presenter(document).render_document_index_label(document_show_link_field(document))
   
   # updated is required, for now we'll just set it to now, sorry
-  xml.updated Time.now.strftime("%Y-%m-%dT%H:%M:%SZ")
+  xml.updated Time.zone.now.strftime("%Y-%m-%dT%H:%M:%SZ")
```
jcoyne: If we use time.zone, we should use the correct UTC offset, rather than saying `Z` (GMT)

app/views/catalog/index.atom.builder
```diff
@@ -43,7 +43,7 @@ xml.feed("xmlns" => "http://www.w3.org/2005/Atom",
   
   
   # updated is required, for now we'll just set it to now, sorry
-  xml.updated Time.now.strftime("%Y-%m-%dT%H:%M:%SZ")
+  xml.updated Time.zone.now.strftime("%Y-%m-%dT%H:%M:%SZ")
```
jcoyne: Same here.

app/views/catalog/_document_default.atom.builder
cbeer: Good catch, I didn't look closely at the `strftime` format. It seems like we should just use `Time#iso8601` to get this right..

PR #1345, opened 2015-12-08
Recent searches blank query

korabh:


cbeer:
Isn't this also excluding searches that only use facets? What problem are you trying to solve?

korabh:
@cbeer If q params is blank than it will return as saved search item.
screen shot 2015-12-08 at 2 06 33 am

cbeer:
No, I mean, what if, instead of using search terms, I navigate using facets:

E.g. to:
http://demo.projectblacklight.org/?f%5Bformat%5D%5B%5D=Book

Which currently displays in the history like:

screen shot 2016-01-14 at 4 59 10 pm

PR #1350, opened 2015-12-17
Add blacklight initializer to explicitly precompile favicon.ico for compatiblity with sprockets-rails 3.x

cbeer:


jcoyne:
:+1: thanks @cbeer

PR #1365, opened 2016-01-18
Fetch the adapter from the config

jcoyne:
`nil` is not a valid option for `:adapter`

### DIFF
--- a/lib/blacklight.rb
+++ b/lib/blacklight.rb
@@ -39,7 +39,8 @@ module Blacklight
     when /::/
       connection_config[:adapter].constantize
     else
-      Blacklight.const_get("#{connection_config[:adapter]}/Repository".classify)
+      raise "The value for :adapter was not found in the blacklight.yml config" unless connection_config.key? :adapter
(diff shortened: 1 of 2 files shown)


lib/blacklight.rb
```diff
@@ -39,7 +39,7 @@ def self.repository_class
     when /::/
       connection_config[:adapter].constantize
     else
-      Blacklight.const_get("#{connection_config[:adapter]}/Repository".classify)
```
cbeer: Ought we just raise an exception if the adapter isn't defined?

jcoyne: If you don't do this, you get the very confusing message that class "Repository" could not be found. I think a `KeyError` does a better job pointing to the problem

cbeer: I meant, raise an exception ourselves that describes the problem and how to fix it (likely a misconfiguration or a Blacklight 5 -> 6 migration error?). If you think just a key error is sufficient, fine.

jcoyne:
@cbeer updated.

cbeer:
:+1:

PR #1371, opened 2016-01-20
Add frozen_string_literals pragma

jcoyne:


app/presenters/blacklight/document_presenter.rb
```diff
@@ -216,5 +217,20 @@ def get_field_values field, field_config, options = {}
...
     end
+
+    private
+
+      # @param [Array<String,Fixnum>] values
+      # @return [Array] an array with all strings converted to UTF-8
+      def recode_values(values)
+        values.collect do |value|
+          if value.respond_to?(:encoding) && value.encoding != Encoding::UTF_8
+            Rails.logger.warn "Found a non utf-8 value in Blacklight::DocumentPresenter. \"#{value}\" Encoding is #{value.encoding}"
```
cbeer: :+1:, but should this be in a separate commit?

jcoyne: This gets around forcing the encoding, which causes an attempt to modify frozen string error.

PR #1373, opened 2016-01-22
update blacklight-marc dependency for use in BL 6.0

tampakis:
blacklight-marc version 5.x is incompatible with blacklight 6.x. Updated install generator and test app generator to make sure user is able to add on MARC-related functionality in Blacklight 6.0.

### DIFF
--- a/lib/generators/blacklight/install_generator.rb
+++ b/lib/generators/blacklight/install_generator.rb
@@ -98,7 +98,8 @@ EOF
 
     def generate_blacklight_marc_demo
       if options[:marc]
-        gem "blacklight-marc", "~> 5.0"
+        blacklight_marc = String.new('blacklight-marc')
+        gem blacklight_marc, '~> 6.0'
 
         Bundler.with_clean_env do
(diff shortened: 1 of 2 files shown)


lib/generators/blacklight/install_generator.rb
```diff
@@ -98,7 +98,8 @@ def inject_blacklight_i18n_strings
 
     def generate_blacklight_marc_demo
       if options[:marc]
-        gem "blacklight-marc", "~> 5.0"
+        blacklight_marc = String.new('blacklight-marc')
```
cbeer: Why this line?

jcoyne: Because rails mutates the string you pass in. And we have the frozen string pragma set.

jcoyne: See https://github.com/rails/rails/issues/23137

cbeer: Ew.

PR #1375, opened 2016-01-25
Fix ERB indenting

jcoyne:


tampakis:
Looks good to me. I'd merge if I could.

tampakis:
Travis seemed to time out. I think you restart the build it should actually run the specs.

PR #1376, opened 2016-01-26
make pagination span full width

tampakis:
Allows the pagination to take up the full width of the main content container. Changes this:
pagination
To this:
updated_pagination

### DIFF
--- a/app/views/catalog/_results_pagination.html.erb
+++ b/app/views/catalog/_results_pagination.html.erb
@@ -1,6 +1,6 @@
 <% if show_pagination? and @response.total_pages > 1 %>
  <div class="row record-padding">
-  <div class="col-md-9"> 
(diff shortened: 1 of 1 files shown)


jcoyne:
:+1: @cbeer, @jkeck, @mejackreed  what do you think? Failures are due to network error fetching solr.

cbeer:
:+1: makes sense to me

mejackreed:
Might make sense to track this with the container class helpers that we already have: https://github.com/projectblacklight/blacklight/blob/1640c0cb4b1bd54c00c6ea647083e617dbb600d5/app/helpers/blacklight/layout_helper_behavior.rb#L21-L33

Definitely an enhancement, though (but so this can be customizable without overriding the partial).

PR #1378, opened 2016-01-29
Blacklight requires Ruby 2.1+

cbeer:


cbeer:
I don't necessarily know why we require 2.1 and not 2.0, but the release notes say 2.1+, and @blalbrit had trouble with 2.0.0.. Happy to look further into this, but requiring 2.1 also seems pretty safe.

PR #1385, opened 2016-02-20
Generate solr4 + jettywrapper if --jettywrapper is given as an option…

cbeer:
… to the blacklight:install generator


cbeer:
This should fix the issue @lmcglohon and others have reported about getting the wrong version of `blacklight-jetty`.

jcoyne:
Can't modify frozen string for 2.3.0.  This is a known bug in rails.  Send gem "whatever".dup instead

cbeer:
Thanks, that's certainly annoying.

PR #1395, opened 2016-03-18
Address bug in default template install process (Solr4Generator).

jkeck:
Don't require the jettywrapper option to be passed to the generator to install jettywrapper (it is already checked before running that generator).

Add the templates directory as a source_root so the generator can access to jetty.yml

### DIFF
--- a/lib/generators/blacklight/install_generator.rb
+++ b/lib/generators/blacklight/install_generator.rb
@@ -27,13 +27,16 @@ module Blacklight
     EOS
 
     def add_solr_wrapper
-      if options[:jettywrapper]
-        generate 'blacklight:solr4'
-      elsif solr_version == 'latest'
-        generate 'blacklight:solr5'
-      else
-        generate "blacklight:solr#{solr_version}"
-      end
+      generator_options = '--jettywrapper' if options[:jettywrapper]
+      solr_generator = case
+                       when options[:jettywrapper]
+                         'blacklight:solr4'
+                       when solr_version == 'latest'
+                         'blacklight:solr5'
+                       else
+                         "blacklight:solr#{solr_version}"
+                       end
+      generate solr_generator, generator_options
     end
 
     def bundle_install
--- a/lib/generators/blacklight/solr4_generator.rb
+++ b/lib/generators/blacklight/solr4_generator.rb
@@ -3,6 +3,11 @@ require 'rails/generators'
 
 module Blacklight
   class Solr4Generator < Rails::Generators::Base
+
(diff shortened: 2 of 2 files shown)


lib/generators/blacklight/solr4_generator.rb
```diff
@@ -10,10 +13,8 @@ class Solr4Generator < Rails::Generators::Base
     EOS
 
     def install_jettywrapper
-      return unless options[:jettywrapper]
```
jkeck: Not sure if this is controversial at all, but in the context that I could find, this is already checked before this generator is run. (See https://github.com/projectblacklight/blacklight/blob/master/lib/generators/blacklight/install_generator.rb#L30)

Not sure if there are other contexts where this would be run but we would not want to install jettywrapper (and only run this generator to install rsolr).

jcoyne:
:+1: LGTM

lib/generators/blacklight/solr4_generator.rb
cbeer: It's possible to also get here from:

https://github.com/projectblacklight/blacklight/blob/master/lib/generators/blacklight/install_generator.rb#L35

Seems like we ought to pass the generator options through, maybe?

jkeck: Yeah, I could see that though.

If you do get there, and you did not pass the jettywrapper options, you'll simply execute the following code: https://github.com/projectblacklight/blacklight/blob/master/lib/generators/blacklight/solr4_generator.rb#L23-L25, which doesn't seem particularly useful to me.

I also experimented w/ passing this options directly on w/o much luck.

jkeck: Also, none of the generators actually make use of any of those options.

lib/generators/blacklight/install_generator.rb
```diff
@@ -27,13 +27,16 @@ class Install < Rails::Generators::Base
...
+      generator_options = '--jettywrapper' if options[:jettywrapper]
+      solr_generator = case
+                       when options[:jettywrapper]
+                         'blacklight:solr4'
+                       when solr_version == 'latest'
+                         'blacklight:solr5'
+                       else
+                         "blacklight:solr#{solr_version}"
+                       end
+      generate solr_generator, generator_options
```
jkeck: @cbeer modified slightly from what we did last week.  Let me know what you think.

cbeer: :+1:

PR #1403, opened 2016-04-14
Stop memoizing the repository instance

cbeer:


app/controllers/concerns/blacklight/search_helper.rb
```diff
@@ -134,7 +134,7 @@ def repository_class
   end
 
   def repository
-    @repository ||= repository_class.new(blacklight_config)
+    @repository = repository_class.new(blacklight_config)
```
jcoyne: Why do we need the ivar?

cbeer: Good point; I was thinking about leaving just in case, but that's kinda silly.

PR #1405, opened 2016-05-07
Update build matrix

cbeer:


jcoyne:
Looks like rspec-rails isn't compatible with Rails 5 RC1

PR #1407, opened 2016-05-10
Convert query_params and title database columns to the binary type

cbeer:
Blocked by https://github.com/rails/rails/pull/24958


jcoyne:
@cbeer can you rebase this? I suspect it will just work when run on rails 5.

PR #1417, opened 2016-06-20
Provide separate index configuration per object type

jcoyne:
Still need to do this for non-index types of views.
Possibly handle multiple types per block.


tampakis:
Working on the rebase now

PR #1421, opened 2016-06-29
Do not use 'should' when describing tests

jcoyne:


spec/features/search_spec.rb
```diff
@@ -111,7 +111,7 @@
     expect(page).to_not have_selector "#q[value='history']"
   end
 
-  it "should gracefully handle searches with invalid facet parameters" do
+  it "gracefullies handle searches with invalid facet parameters" do
```
cbeer: `s/gracefullies/gracefully/`

jcoyne: would you believe that this was done with a script?

cbeer: 😱

cbeer:
Rebase (and refix with the new changes from #1415?)

PR #1423, opened 2016-06-30
[WIP]support Turbolinks 5 event

mejackreed:
Not really sure how to test this.. any ideas?

Relevant SO post: http://stackoverflow.com/a/18770589

### DIFF
--- a/app/assets/javascripts/blacklight/core.js
+++ b/app/assets/javascripts/blacklight/core.js
@@ -14,7 +14,8 @@ Blacklight = function() {
 }();
 
 // turbolinks triggers page:load events on page transition
-// If app isn't using turbolinks, this event will never be triggered, no prob. 
+// If app isn't using turbolinks, this event will never be triggered, no prob.
+// Support for Turbolinks classic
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/blacklight/core.js
```diff
@@ -14,7 +14,8 @@ Blacklight = function() {
 }();
 
 // turbolinks triggers page:load events on page transition
-// If app isn't using turbolinks, this event will never be triggered, no prob. 
+// If app isn't using turbolinks, this event will never be triggered, no prob.
+// Support for Turbolinks classic
 $(document).on('page:load', function() {
```
cbeer: Can this just be:

`$(document).on('page:load turbolinks:load', function() {`?

mejackreed: Maybe.. I'm trying to figure out what the best way to support this forward and backwards compatible is. I'm running into issues where activate will fire twice (once from `ready` and once from `turbolinks:load`)

PR #1427, opened 2016-07-01
Extract set_current_search_session and find_search_session methods from current_search_session

cbeer:


jcoyne:
👍 ✖️ 💯

mejackreed:
Nice! Working great for me. I'm wondering if there is a way that this can be tested so it doesn't get removed later on?

maybe in `./spec/features/search_context_spec.rb`?

jcoyne:
@mejackreed I'm inclined to merge as is. This is functionally tested, but not unit tested. I'm happy with that.

PR #1431, opened 2016-07-06
Add deprecated methods (deprecated by DocumentPresenter) into the new presenters

cbeer:
This should fix more backwards compatibility breaks introduced in 6.3.

### DIFF
--- a/spec/presenters/show_presenter_spec.rb
+++ b/spec/presenters/show_presenter_spec.rb
@@ -283,5 +283,31 @@ describe Blacklight::ShowPresenter do
       end
     end
   end
+
+  describe 'deprecated methods' do
+    before do
+      allow(Deprecation).to receive(:warn)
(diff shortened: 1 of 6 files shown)


spec/presenters/show_presenter_spec.rb
```diff
@@ -283,5 +283,31 @@ def mock_document_app_helper_url *args
       end
     end
   end
+
+  describe 'deprecated methods' do
+    before do
+      allow(Deprecation).to receive(:warn)
```
atz: If their being deprecated is part of the point, shouldn't this be a `before :each { expect ... }`?

cbeer: I'm not sure we need to care; I wasn't setting out to test if `deprecation` was working, just to keep it quiet.

PR #1433, opened 2016-07-07
Check that Turbolinks is not undefined before checking support

mattcg:
When Turbolinks is not being used, the variable will be undefined and you will get a `ReferenceError: Turbolinks is not defined`.

The fix is check that `Turbolinks` is not undefined, rather than checking its truthiness.


cbeer:
cc: @mejackreed  ^^?

mejackreed:
Err... yeah.  Thanks for the heads up.

PR #1434, opened 2016-07-07
Deprecate render_(index|show)_field_value and use the presenter

jcoyne:
I know this introduces a bunch of deprecation warnings in the tests which I intend to take care of.


cbeer:
Makes sense to me; ought we do the same for the show view too?

PR #1435, opened 2016-07-07
Fix strip_heredoc warning for the blacklight:user generator

cbeer:


jkeck:
`+17.6%` coverage increase on removing an `s`.  Well done!

PR #1436, opened 2016-07-08
Button style for bookmarks when JS is not enabled

mattcg:
Bookmark buttons should used the default button style when Javascript is not enabled. Otherwise, the buttons render using the browser default styling, which jars next to other buttons.


mejackreed:
👍

PR #1440, opened 2016-07-12
Bump devise-guests version in generator

### LINKED ISSUE #1428: bump devise-guests version
The generated dependency on devise-guests is sem-ver `~> 0.3`:
https://github.com/projectblacklight/blacklight/blob/v6.3.0/lib/generators/blacklight/user_generator.rb#L26

There are a half dozen releases since then.  We can bring that up to `~> 0.5` without shifting any of the underlying dependency compatibility, while delivering a more uniform and current base.

atz:
Closes #1428

### DIFF
--- a/lib/generators/blacklight/user_generator.rb
+++ b/lib/generators/blacklight/user_generator.rb
@@ -23,7 +23,7 @@ module Blacklight
         else
           gem "devise"
         end
-        gem "devise-guests", "~> 0.3"
+        gem "devise-guests", "~> 0.5"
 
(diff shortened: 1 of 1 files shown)


jcoyne:
@atz please rebase this.

atz:
Not sure why you want a rebase, since the 1-liner didn't conflict, but np, rebased.

jcoyne:
@atz  the rebase was so the tests would pass as you fixed the broken parts in another PR

PR #1444, opened 2016-07-19
Update solr5 generator solr_wrapper require.

grosscol:
Avoid requiring solr_wrapper in production environment as the gem will not be installed.


jcoyne:
This seems reasonable. We never had this problem with `jettywrapper` because we put it in the root of the gemfile rather than the `:development` and `:test` groups.

PR #1449, opened 2016-07-28
Consistently use FacetField#field when retrieving aggregation values

cbeer:
Fixes https://github.com/projectblacklight/spotlight/issues/1457


jcoyne:
Please add a test for `facet_by_field_name`

PR #1453, opened 2016-08-09
Update Blacklight::SearchState to play well with Rails 5 AC::Parameters class

cbeer:


lib/blacklight/search_state.rb
```diff
@@ -10,22 +10,26 @@ class SearchState
...
-        # https://github.com/rails/rails/pull/22913 (Rails > 4.2.5)
-        @params = params.with_indifferent_access
-      else
+      if params.respond_to?(:to_unsafe_h)
         # This is the typical (not-ActionView::TestCase) code path.
         @params = params.to_unsafe_h
         # In Rails 5 to_unsafe_h returns a HashWithIndifferentAccess, in Rails 4 it returns Hash
         @params = @params.with_indifferent_access if @params.instance_of? Hash
+      elsif params.is_a? Hash
+        # This is an ActionView::TestCase workaround. Will be resolved by
```
jcoyne: Will not be resolve apparently. Rafael declined to merge it.

cbeer: :/ .. I'll remove the comment too.

lib/blacklight/search_state.rb
```diff
@@ -10,22 +10,24 @@ class SearchState
...
-        # https://github.com/rails/rails/pull/22913 (Rails > 4.2.5)
-        @params = params.with_indifferent_access
-      else
+      if params.respond_to?(:to_unsafe_h)
         # This is the typical (not-ActionView::TestCase) code path.
         @params = params.to_unsafe_h
         # In Rails 5 to_unsafe_h returns a HashWithIndifferentAccess, in Rails 4 it returns Hash
         @params = @params.with_indifferent_access if @params.instance_of? Hash
+      elsif params.is_a? Hash
+        @params = params.dup.with_indifferent_access
```
jcoyne: Can you add a comment here that this is a workaround for AV::TestCase and is only needed for rails 4.2 (so we can remove it down the road?)

cbeer: 👌

PR #1466, opened 2016-08-20
Deprecation cleanup

### LINKED ISSUE #1463: destroy_all conditions deprecated
```
DEPRECATION WARNING: 
Passing conditions to destroy_all is deprecated and will be removed in Rails 5.1. 
To achieve the same use where(conditions).destroy_all. 
(called from delete_old_searches at /Users/atz/repos/blacklight/app/models/search.rb:23)
```

### LINKED ISSUE #1464: rails render method deprecations
```
DEPRECATION WARNING: 
`:nothing` option is deprecated and will be removed in Rails 5.1. 
Use `head` method to respond with empty response body. 
(called from block (2 levels) in additional_response_formats at .../app/controllers/concerns/blacklight/catalog.rb:146)
```
```
DEPRECATION WARNING: 
`render :text` is deprecated because it does not actually render a `text/plain` response. Switch to
`render plain: 'plain text'` to render as `text/plain`, 
`render html: '<strong>HTML</strong>'` to render as `text/html`, or 
`render body: 'raw'` to match the deprecated behavior and render with the default Content-Type, which is `text/plain`. 
(called from block (5 levels) in <top (required)> at .../spec/controllers/catalog_controller_spec.rb:187)
```

### LINKED ISSUE #1465: Mime type call deprecated
```
DEPRECATION WARNING: 
Accessing mime types via constants is deprecated. 
Please change `Mime::HTML` to `Mime[:html]`. 
```

In spec/controllers/catalog_controller_spec.rb lines 501 and 509.

atz:
Fixes #1464 
Fixes #1465 
Fixes #1463 

Cleans up catalog_controller_spec considerably.

### DIFF
--- a/app/controllers/concerns/blacklight/catalog.rb
+++ b/app/controllers/concerns/blacklight/catalog.rb
@@ -43,11 +43,9 @@ module Blacklight::Catalog
     # to add responses for formats other than html or json see _Blacklight::Document::Export_
     def show
       @response, @document = fetch params[:id]
-
       respond_to do |format|
         format.html { setup_next_and_previous_documents }
         format.json { render json: { response: { document: @document } } }
-
         additional_export_formats(@document, format)
       end
     end
@@ -71,15 +69,11 @@ module Blacklight::Catalog
       @facet = blacklight_config.facet_fields[params[:id]]
       @response = get_facet_field_response(@facet.key, params)
       @display_facet = @response.aggregations[@facet.key]
-
       @pagination = facet_paginator(@facet, @display_facet)
-
-
       respond_to do |format|
         # Draw the facet selector for users who have javascript disabled:
         format.html
         format.json
-
         # Draw the partial for the "more" facet modal window:
         format.js { render :layout => false }
       end
@@ -88,12 +82,8 @@ module Blacklight::Catalog
     # method to serve up XML OpenSearch description and JSON autocomplete response
     def opensearch
       respond_to do |format|
-        format.xml do
-          render :layout => false
-        end
-        format.json do
-          render :json => get_opensearch_response
-        end
+        format.xml { render layout: false }
+        format.json { render json: get_opensearch_response }
       end
     end
 
@@ -107,7 +97,7 @@ module Blacklight::Catalog
 
     ##
     # Check if any search parameters have been set
-    # @return [Boolean] 
+    # @return [Boolean]
     def has_search_parameters?
       !params[:q].blank? or !params[:f].blank? or !params[:search_field].blank?
     end
@@ -129,13 +119,10 @@ module Blacklight::Catalog
     ##
     # Render additional response formats for the index action, as provided by the
     # blacklight configuration
-    #
-    # example:
-    #
-    #   config.index.respond_to.txt = Proc.new { render text: "A list of docs." }
-    #
-    # Make sure your format has a well known mime-type or is registered in
-    # config/initializers/mime_types.rb
+    # @param [Hash] format
+    # @note Make sure your format has a well known mime-type or is registered in config/initializers/mime_types.rb
+    # @example
+    #   config.index.respond_to.txt = Proc.new { render plain: "A list of docs." }
     def additional_response_formats format
       blacklight_config.index.respond_to.each do |key, config|
         format.send key do
@@ -160,7 +147,7 @@ module Blacklight::Catalog
     # the document extension framework. See _Blacklight::Document::Export_
     def additional_export_formats(document, format)
       document.export_formats.each_key do | format_name |
-        format.send(format_name.to_sym) { render text: document.export_as(format_name), layout: false }
+        format.send(format_name.to_sym) { render body: document.export_as(format_name), layout: false }
       end
     end
 
@@ -169,7 +156,6 @@ module Blacklight::Catalog
     def document_export_formats format
       format.any do
         format_name = params.fetch(:format, '').to_sym
-
         if @response.export_formats.include? format_name
           render_document_export_format format_name
         else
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -129,21 +119,22 @@ def store_preferred_view
...
+    #   config.index.respond_to.txt = Proc.new { render plain: "A list of docs." }
     def additional_response_formats format
       blacklight_config.index.respond_to.each do |key, config|
         format.send key do
           case config
           when false
             raise ActionController::RoutingError, 'Not Found'
           when Hash
-            render config
+            if config[:nothing] # nothing param  deprecated in rails, but still in configs
```
cbeer: Do we need to handle this case? Wouldn't it be better to get downstream applications to update their config? At the very least, we should issue our own deprecation warning so they do.

app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -129,21 +119,22 @@ def store_preferred_view
...
+              head :ok, config.reject { |k, v| k == :nothing }
```
cbeer: `config.except(:nothing)` ought to work here, yeah?

atz: yeah, that's fine

app/controllers/concerns/blacklight/catalog.rb
atz: Data migration of existing production configs is outside my intended scope for this PR, but I otherwise agree with you.  I needed to catch it to resolve the last outstanding deprecation warnings.  Not sure what to advise the migrate to.  Since previous configs were passed to `render`, that was easy, but the deprecation we are trying to avoid is precisely that we can't do `render nothing: true` anymore.  Seems like we will need a conditional regardless.

atz:
@cbeer Need some insight on the Rails 4-specific failure on a single test, e.g.:
https://travis-ci.org/projectblacklight/blacklight/jobs/153731408#L791

I get why it isn't finding a template, but:
- don't controller tests stub templates? and
- why is it just that one version of rails?

app/controllers/concerns/blacklight/catalog.rb
cbeer: My point is, this isn't blacklight's deprecation to clean up. We're just an intermediate party, and downstream applications will need to get updated. I'm +0, though, on capturing their deprecation warning and giving them an equivalent but more accurate deprecation warning.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -654,45 +603,41 @@ def export_as_mock
...
-        describe "with invalid params" do
-          before { allow(controller).to receive(:validate_like_params).and_return(false) }
-          it "does not call the supplied method if validation failed" do
-            expect(controller).not_to have_received(:perform_like)
-          end
+      context "with failure on invalid params" do
+        it "does not call the supplied method" do
+          expect(controller).to receive(:validate_like_params).and_return(false)
+          expect(controller).not_to receive(:perform_like)
+          expect { post :like }.not_to raise_error
```
atz: This is now the sole failing test on Rails 4.  

Note: the old version of the test was bogus.  The post was already made at 683 _before_ `validate_like_params` was re-stubbed false.

app/controllers/concerns/blacklight/catalog.rb
atz: Blacklight set the config syntax and behavior.  To say it is the downstream app's problem means breaking those apps on rails 5.1 for the same configs blacklight currently uses (and maybe even generates).

spec/controllers/catalog_controller_spec.rb
cbeer: Ah, I see the problem now -- `before` blocks are executed and unwrapped from the most-specific block to the least, so:

```ruby
before { allow(controller).to receive(:validate_like_params).and_return(false) }
```

and 

```ruby
it "..."
expect(controller).to receive(:validate_like_params).and_return(false)
...
```

are executed at different times.

cbeer: Worse than that -- I'm not sure the `post` was even triggered. It seems like we're not effectively handling invalid input (in actual use, it'll fail over to the original template.. but still)

cbeer: For now, I'd suggest marking it as pending and move on.

app/controllers/concerns/blacklight/catalog.rb
cbeer: The parameters to `respond_to` are intended and expected to mirror the Rails parameters, we're not trying to invent our own `render` API. If downstream applications ignore the deprecation warning and upgrade to Rails 5.1 without heeding those warnings, that's on them.

atz: How do you expect them to handle the case currently borne by `render :nothing`?   If we don't expose `head`, there is no option they can pass via this switch to `render` that 5.1 will support.   That's not their failure to resolve deprecations, if we're the ones forcing them to pass options to `render` in the first place.

cbeer: This seems acceptable:

```
config.index.respond_to.txt = Proc.new { head :ok }
```

Also acceptable might be `config.index.respond_to.txt = { body: "" }`, or similar variations.

atz: Ok, I see what you mean.  You want me to back this change out?

cbeer: Yes, please.

app/models/concerns/blacklight/document/export.rb
```diff
@@ -40,56 +40,56 @@
...
 
     export_formats[short_name] =  { content_type: content_type }
   end
-  
+
   # Collects formats that this doc can export as.
   # Returns a hash, keys are format short-names that can
   # be exported. Hash includes:
-  #  :content-type => mime-content-type
+  #  :content-type => { content_type: content_type }
```
jcoyne: I think this was intended to be a placeholder for a string, so something like this would be more correct:
```ruby
:content_type => 'sample/content-type'
```

cbeer: Yeah, it seems like a bad comment. The method returns a hash, with values that are themselves hashes. The example here was presumably demonstrating the contents of a value, not the entire hash. Maybe we can rework the comment to be better?

atz: Yeah, it was clearly half-baked previously.  I was just trying to capture that there was internal structure (values are hashes).  I'll touch it up and merge.

PR #1469, opened 2016-08-23
Yard see and cleanup

atz:
One test error corrected (`reverse_merge`, not `merge`).  

The `@see` pointer helps us understand the relationship between params and keeps us from having to restate (and maintain) the same lines.

Also some rubocop-friendly whitespace cleanup, guard clauses, and method def parens.

### DIFF
--- a/app/controllers/concerns/blacklight/default_component_configuration.rb
+++ b/app/controllers/concerns/blacklight/default_component_configuration.rb
@@ -20,21 +20,26 @@ module Blacklight
       add_nav_action(:search_history, partial: 'blacklight/nav/search_history')
     end
 
-    def render_sms_action? config, options = {}
+    def render_sms_action?(_config, _options)
       sms_mappings.present?
     end
 
     module ClassMethods
+      # YARD will include inline disabling as docs, cannot do multiline inside @!macro.  AND this must be separate from doc block.
+      # rubocop:disable Metrics/LineLength
 
-      ##
+      # @!macro partial_if_unless
+      #   @param name [String] the name of the document partial
+      #   @param opts [Hash]
+      #   @option opts [Symbol,Proc] :if render this action if the method identified by the symbol or the proc evaluates to true. The proc will receive the action configuration and the document or documents for the action.
+      #   @option opts [Symbol,Proc] :unless render this action unless the method identified by the symbol or the proc evaluates to true. The proc will receive the action configuration and the document or documents for the action.
+
       # Add a partial to the tools for rendering a document
-      # @param partial [String] the name of the document partial
-      # @param opts [Hash]
-      # @option opts [Symbol,Proc] :if render this action if the method identified by the symbol or the proc evaluates to true.
-      #                             The proc will receive the action configuration and the document or documents for the action.
-      # @option opts [Symbol,Proc] :unless render this action unless the method identified by the symbol or the proc evaluates to true
-      #                             The proc will receive the action configuration and the document or documents for the action.
-      def add_show_tools_partial name, opts = {}
+      # @!macro partial_if_unless
+      # @option opts [Boolean] :define_method define a controller method as named, default: true
+      # @option opts [Symbol]  :validator method for toggling between success and failure, should return Boolean (true if valid)
+      # @option opts [Symbol]  :callback method for further processing of documents, receives Array of documents
+      def add_show_tools_partial(name, opts = {})
         blacklight_config.add_show_tools_partial(name, opts)
 
         return if method_defined?(name) || opts[:define_method] == false
@@ -43,9 +48,8 @@ module Blacklight
         define_method name do
           @response, @documents = action_documents
 
-          if request.post? and
-              opts[:callback] and
-              (opts[:validator].blank? || self.send(opts[:validator]))
+          if request.post? && opts[:callback] &&
+            (opts[:validator].blank? || self.send(opts[:validator]))
 
             self.send(opts[:callback], @documents)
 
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/blacklight/default_component_configuration.rb
```diff
@@ -64,39 +65,26 @@ def add_show_tools_partial name, opts = {}
...
-      ##
       # Add a tool to be displayed for each document in the search results.
-      # @param partial [String] the name of the document partial
-      # @param opts [Hash]
-      # @option opts [Symbol,Proc] :if render this action if the method identified by the symbol or the proc evaluates to true.
-      #                             The proc will receive the action configuration and the document or documents for the action.
-      # @option opts [Symbol,Proc] :unless render this action unless the method identified by the symbol or the proc evaluates to true
-      #                             The proc will receive the action configuration and the document or documents for the action.
-      def add_results_document_tool name, opts = {}
+      # @see #add_nav_action for params
```
cbeer: Does `@see` imply a stronger relation to `#add_nav_action` than actually exists? These methods just happen to share the same arguments. Maybe YARD macros would be more appropriate?

atz: No, it just points to it.  @see could also reference external URLs.  They happen to share the same args, because they are consumed in the same environments and the opts hash is common to all of them.  A macro might be more useful if we actually cared about generating standalone YARD/rdoc.  I don't.  I expect to read the docs in the code.

cbeer: I think we do expect generated documentation (at e.g. http://www.rubydoc.info/gems/blacklight). 

I don't see how the indirection with `@see` is any better or worse than `@!macro`.. of course, the repetition doesn't bother me at all in the first place.

jkeck: My $0.02 is to keep the duplication or use a macro.  If I were a developer browsing the online documentation (which is probably a lot more common than you might think), I would find this `@see` indirection confusing and not user-friendly (and would expect to see the method documented in-line).

Proliferating a pattern of connecting method documentation based on sharing a method signature feels potentially problematic to me.  It seems like potential downside of reducing this duplication outweighs the benefit that it is giving us.  Again, just my $0.02, hopefully it's not too controversial.

jrochkind: The problem with duplication in docs is maintenance. It's hard enough to remember to fix the docs to match code changes when it's only _one_ place you have to change, let alone when you have to remember to find _all_ the places you have to change (and I'm not sure how you'd find them easily?). 

Blacklight (and this is not unique to BL, it's common) has not always been great historically at updating inline docs to match code changes, it hasn't been super unusual to find inline docs that are out of date and no longer accurate.

atz: As a community we are at best inconsistent about producing yard/rdoc that is worth reading, throughout the stack.  I only read the docs on github (instead of rubydoc et al.), because it allows me to check immediately whether they are bogus or not.  These sections of blacklight are better than most (so kudos for that), but still have factual errors (e.g., parameter name is `name`, not `partial`).  

I'm amenable to using a macro, but to me that's less transparent than the `@see` statement, largely directed at an output I don't consume.  It does have the benefit of being visible at the point of concentration, with the downside of being essentially a novel piece of code to worry about.

jrochkind: so personally i think any method of DRY in the docs is better than manual duplication. Whether see or a macro. I suspect the main advantage of see is just that it's easy to implement and easy for future comment-maintainers to understand without having to look up how to do macros and make it so.

jkeck: I certainly prefer macro in this case (people working on this codebase should be able to reasonably find and understand how to use the tool chain, yard being one of the tools).

My comment was more to the affect that the repetition in these comments wouldn't have jumped out at me as problematic in the first place (doing nothing is less work than adding indirection).  But now that we've done it, macro feels like the proper tool for the job.

atz: Here is what the generated YARD looks like with `@see`.   To me, that is perfectly usable, but if you'd prefer the repetition of text, that's fine.

screen shot 2016-08-30 at 2 19 41 pm

jrochkind: I don't think the `see` looks great actually, but I _still_ think ti's better than copy-and-pasted duplicated tests. I think copy-and-pasted duplicated tests makes it even more likely that at least one place won't be kept up to date when the parameters change. Blacklight has not always done so well at keeping documented method signatures up to date even without duplication. I'd rather have docs with slightly less good UX if they are more likely to be up to date; a great UX on out of date docs helps nobody.

cbeer: > I don't think the see looks great actually,

I agree, but again, we can use the YARD macro syntax, remove duplicated documentation text, and maintain a good user experience.

Seeing `@see` in context makes me hate it even more; as a user, I click `for params`, get bounced around the page, and it's unclear what the association between the two locations is.

atz:
Replaced `@see` with `@macro` as suggested.  Looks like:

screen shot 2016-08-30 at 2 51 15 pm

While we're renewed fans of advanced YARD usage, I'll note that `yardoc` currently spews **130 warnings** including: unknown names, unknown tags, unresolvable links, syntactic and parser errors.  Will ticket separately.

atz:
Notably, one of the busted pieces of yard was:
```
# @params (see #render_facet_value)
```

Putting aside `@params` vs. `@param`, the standard of the actual code seems to have no problem with even a non-link pointer to another method for parameter docs.  See [here](https://github.com/projectblacklight/blacklight/commit/486d8a2d6b7c1dac3f62c2ecb041eeab3900906a#diff-40e61fb902391e24173431e7b37ff42fR126).  

That being said, the macros are fine (with the complication of not handling long line continuation, therefore needing rubocop exceptions, which cannot be attached inline, nor directly bordering the docs, without also becoming *part* of the docs).  Unless there are further questions, I'll merge shortly.

jcoyne:
@atz It's our practice in Blacklight to have a contributor other than the author to merge the PR.

PR #1470, opened 2016-08-23
Better getter setter

### LINKED ISSUE #1462: separate SearchBuilder getters and setters
`start`, `page` and `rows` all have bulky workarounds to operate as both getter and setter:

https://github.com/projectblacklight/blacklight/blob/d55d6f566d203a092f7f40987a37d8988df5500d/lib/blacklight/search_builder.rb#L157-L211

Method signatures are confused and overriding in a subclass unnecessarily complicated.
E.G., `start` returns `self` when used as a setter, and an integer (or nil) otherwise... except when setting to nil.  A subclass attempting to override the getter `rows` would break the app since the getter IS the setter.  Separation would provide clarity and utility.

I don't really care about the `return self` style since I doubt it is greatly used, but it could be maintained for the setters anyway.

atz:
Fixes #1462

If we wanted to drop/limit the builder pattern we would insert
deprecation warnings in those methods to be converted to pure getters.
If we want to keep the builder pattern, then we let it ride with the new explicit setters.

Updated calls where the returned `self` wasn't being used and the basic setter pattern is more idiomatic.

Added tests to make sure builder-style w/ arg calls underlying setter method.

### DIFF
--- a/app/controllers/concerns/blacklight/search_helper.rb
+++ b/app/controllers/concerns/blacklight/search_helper.rb
@@ -1,13 +1,12 @@
 # frozen_string_literal: true
 # SearchHelper is a controller layer mixin. It is in the controller scope: request params, session etc.
-# 
+#
 # NOTE: Be careful when creating variables here as they may be overriding something that already exists.
 # The ActionController docs: http://api.rubyonrails.org/classes/ActionController/Base.html
 #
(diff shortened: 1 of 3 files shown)


cbeer:
In the future, could you please separate functional changes from rubocop-driven cleanup? The line noise makes the commit very difficult to read.

PR #1475, opened 2016-08-30
Rake task for solr_wrapper commands exposed to bl-based apps

bibliotechy:
Allows a blacklight specific solr instance to be spun up with `rake blacklight:solr:start` to make it easier to get started developing a blacklight based app.

Runs the `solr_wrapper` shell command  but passes in options to use the default blacklight solr conf and a collection named blacklight-core.

Provides an option to persist the data.

If merged, I'll update the [Quickstart](https://github.com/projectblacklight/blacklight/wiki/Quickstart) to use this command.

### DIFF
--- a/lib/railties/blacklight.rake
+++ b/lib/railties/blacklight.rake
@@ -118,4 +118,15 @@ namespace :blacklight do
       exit 1 if errors > 0
     end
   end
+
+  namespace :solr do
+    desc "Start solr with default configurations"
+    task :start, [:persist?] => [:environment]  do |_, args|
+      args[:persist?] ||= false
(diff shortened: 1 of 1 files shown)


jcoyne:
could we generate a .solr_wrapper.yml into the test app instead?  Then it would just run `solr_wrapper`?

bibliotechy:
@jcoyne That does sound reasonable.  I think we'd have top copy over the solr/conf from blacklight as well, otherwise the originally templated path to the blacklight gem solr/conf would go stale as user upgrades gem versions. 

If that sounds right, I'll close this an open a new on e with those changes.

PR #1477, opened 2016-08-31
remove dependency in SearchState

codeforkjeff:
`Blacklight::Controller` conveniently allows you to override `#search_state` to provide your own subclass of `Blacklight::SearchState`. But when I try to create such a subclass, I get an 'uninitialized constant Blacklight::Facet' error, because that's a controller concern that hasn't been loaded yet.

This PR removes that dependency, making it possible to subclass SearchState.

### DIFF
--- a/app/controllers/concerns/blacklight/facet.rb
+++ b/app/controllers/concerns/blacklight/facet.rb
@@ -4,6 +4,9 @@ module Blacklight
   # They are only dependent on `blacklight_config` and `@response`
   #
   module Facet
+
+    delegate :facet_configuration_for_field, to: :blacklight_config
+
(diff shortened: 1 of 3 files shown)


atz:
Makes sense to me.  We definitely have an overall problem of modules not being properly isolated (anticipating features or dependencies in the includer's context, and not even bothering to check when included whether that is true or not).  Can you rebase it?

codeforkjeff:
Thanks for looking at this so quickly! I've rebased this branch with master.

jcoyne:
Looks good to me. 👍 Thanks. @codeforkjeff.

@cbeer would you care to take a look?

PR #1478, opened 2016-08-31
Generate solr conf and .solr_wrapper

bibliotechy:
Updates the solr5 install generator to add copy the blacklight solr directory and the .solr_wrapper file to the blacklight based app.

Will allow users who run the install generator to just run `solr_wrapper` to get a blacklight compatible solr instance for development.

Replaces #1475


spec/integration/generators/blacklight/solr5_generator_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'spec_helper'
+require 'generators/blacklight/solr5_generator'
+
+describe Blacklight::Solr5Generator do
```
cbeer: Do we need to test these? It seems like we're testing that the Rails generator works.

lib/generators/blacklight/solr5_generator.rb
```diff
@@ -16,6 +23,14 @@ def install_solrwrapper
       append_to_file "Rakefile", "\nrequire 'solr_wrapper/rake_task' unless Rails.env.production?\n"
     end
 
+    def copy_solr_conf
+      directory 'solr', 'solr'
+    end
+
+    def solr_wrapper_config
+      copy_file '.solr_wrapper', '.solr_wrapper'
```
cbeer: Can we please rename this file `.solr_wrapper.yml`? It was ok just in blacklight, but I don't want to propagate it downstream.

spec/integration/generators/blacklight/solr5_generator_spec.rb
bibliotechy: I thought it was worth it since the generator methods depend on the solr directory and .solr_wrapper.yml being in `Blacklight.root`, not in the standard generators template directory. I could see future changes moving the solr directory and breaking this generator. 

But if it feels like overkill, I can remove the test.

lib/generators/blacklight/solr5_generator.rb
bibliotechy: Renamed the file in blacklight root and updated the generator method. Now copies it to the bl based app as .solr_wrapper.yml

PR #1479, opened 2016-08-31
Add rubocop to the build

cbeer:


atz:
Did auto-corrections actually induce logical (test-breaking) changes?  I have seen that in our stack specifically for the `RedundantSelf` cop.

PR #1480, opened 2016-08-31
Fix #search_action_url for Rails 5 compatibility

cbeer:


jcoyne:
Can you add a test showing how this is fixed?

atz:
Just 1 failed test around whitelisting params:
```
 1) CatalogController search_action_url is the same as the catalog url
     Failure/Error: url_for(options.merge(action: 'index'))
     
     ArgumentError:
       Attempting to generate a URL from non-sanitized request parameters! An attacker can inject malicious data into the generated URL, such as changing the host. Whitelist and sanitize passed parameters to be secure.
     # /home/travis/.rvm/gems/ruby-2.3.1/gems/actionpack-5.0.0/lib/action_dispatch/routing/url_for.rb:176:in `url_for'
     # ./app/controllers/concerns/blacklight/catalog.rb:181:in `search_action_url'
     # ./spec/controllers/catalog_controller_spec.rb:650:in `block (3 levels) in <top (required)>'
...
rspec ./spec/controllers/catalog_controller_spec.rb:648 # CatalogController search_action_url is the same as the catalog url
```

cbeer:
@jcoyne I'm not sure there's a valuable test to write here. All we're doing is relaxing our API from `Hash` to something that responds to `#merge`. I started writing a test with an `AC::Parameters`, but it doesn't demonstrate anything particularly interesting.

jcoyne:
@cbeer isn't calling `url_for` with `AC::Parameters` deprecated?  Here's what I see:

```ruby
Rails.application.routes.url_helpers.url_for ActionController::Parameters.new('foo' => 'bar').merge('baz' => 'quix')
DEPRECATION WARNING: Method to_hash is deprecated and will be removed in Rails 5.1, as `ActionController::Parameters` no longer inherits from hash. Using this deprecated behavior exposes potential security problems. If you continue to use this method you may be creating a security vulnerability in your app that can be exploited. Instead, consider using one of these documented methods which are not deprecated: http://api.rubyonrails.org/v5.0.0.1/classes/ActionController/Parameters.html (called from irb_binding at (irb):5)
```

cbeer:
Yeah, it definitely is, but we have downstream code that was relying on that behavior (somewhere in advanced search, for one).. better to give them a deprecation warning than a nasty exception.

jcoyne:
Can we do something with #to_unsafe_h

cbeer:
I think I'd rather force our downstream consumer to do something risky like `#to_unsafe_h`.

PR #1484, opened 2016-09-06
Update examples that no longer rely on escaping

atz:
The query syntax was updated, but the example descriptions were not.


atz:
Note: Jobs are failing on attempting solr download.

PR #1497, opened 2016-09-10
Update to use bootstrap 4 styles

cbeer:
screen shot 2016-09-13 at 3 32 49 pm


cbeer:
I think this is starting to approach not-too-embarassing; my biggest complaint is font sizes are all over the place.

PR #1501, opened 2016-09-15
Blacklight::Document should freeze its input to guard against modification

cbeer:


cbeer:
It'd surprise me if anyone would, but that's one reason why it'll be in a major release.

PR #1505, opened 2016-09-15
Consolidate suggest behavior into the CatalogController

cbeer:


jcoyne:
Why are we moving this?

cbeer:
`SuggestController` struck me as pretty closely related to `CatalogController`, and I suspect if you have multiple `Blacklight::Catalog`-based implementations, you'd end up needing multiple `SuggestController`-subclassing too.

PR #1506, opened 2016-09-15
Pull layout information up from templates to layouts

cbeer:


cbeer:
I'm not sure this is necessarily worth the added complexity, but it might help accommodate sites with alternative layouts.

screen shot 2016-09-15 at 9 46 41 am

PR #1507, opened 2016-09-15
Include the bare partial in document partial lookups and use the bare name by default

cbeer:


mejackreed:
Will this also go and lookup `_default`'s too?

cbeer:
Yep, I left `_default` in the stack with a slightly higher priority so local overrides will continue to work.

PR #1516, opened 2016-09-17
Underscore unused arguments

jcoyne:


lib/blacklight/solr/search_builder_behavior.rb
```diff
@@ -112,7 +112,7 @@ def add_facetting_to_solr(solr_parameters)
         if facet.pivot
           solr_parameters.append_facet_pivot with_ex_local_param(facet.ex, facet.pivot.join(","))
         elsif facet.query
-          solr_parameters.append_facet_query facet.query.map { |k, x| with_ex_local_param(facet.ex, x[:fq]) }
+          solr_parameters.append_facet_query facet.query.map { |_k, x| with_ex_local_param(facet.ex, x[:fq]) }
```
cbeer: I wonder if using e.g. `values.map` would be clearer? :shrug:

lib/blacklight/engine.rb
```diff
@@ -8,7 +8,7 @@ class Engine < Rails::Engine
 
     # BlacklightHelper is needed by all helpers, so we inject it
     # into action view base here. 
-    initializer 'blacklight.helpers' do |app|
+    initializer 'blacklight.helpers' do |_app|
```
cbeer: Drop these unused block arguments entirely?

app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -8,7 +8,7 @@ module Blacklight::FacetsHelperBehavior
   # @param [Array<String>] fields
   # @param [Hash] options
   # @return [Boolean]
-  def has_facet_values? fields = facet_field_names, options = {}
```
cbeer: I don't know how downstream applications would use `options` here; maybe we can just drop it as part of this major release?

lib/railties/blacklight.rake
```diff
@@ -24,7 +24,7 @@ namespace :blacklight do
 
   namespace :check do
     desc "Check the Solr connection and controller configuration"
-    task :solr, [:controller_name] => [:environment] do |_, args|
+    task :solr, [:controller_name] => [:environment] do |_, _args|
```
cbeer: Drop the unused block arguments?

jcoyne:
@cbeer Changes made.

PR #1519, opened 2016-09-17
Align parameter and if/end blocks

jcoyne:


atz:
@jcoyne: rebased to resolve conflict in `app/helpers/blacklight/catalog_helper_behavior.rb`.

PR #1523, opened 2016-09-17
Wildcard solr schema

cbeer:
- [x] create and populate spellcheck fields
- [x] update search fields for general compatibility with previous config
- [x]  remove activefedora references
- [x] use real-time get


spec/features/bookmarks_spec.rb
```diff
@@ -58,26 +58,26 @@
...
 
-    visit solr_document_path('2007020969')
+    visit solr_document_path('2007020969') # Strong Medicine speaks
     click_button 'Bookmark'
 
     visit "/bookmarks?per_page=1"
-    expect(page).to have_content 'Ci an zhou bian'
-    expect(page).not_to have_content 'Strong Medicine speaks'
+    expect(page).to have_content 'Strong Medicine speaks'
+    expect(page).not_to have_content 'Ci an zhou bian'
```
jkeck: I'm a bit surprised to see the change that we're making affect this test.  Is there a particular reason that the sort changed here?

PR #1537, opened 2016-09-29
Demote rsolr to a development dependency

cbeer:


jcoyne:
Should we remove the `require 'rsolr'` here? https://github.com/projectblacklight/blacklight/blob/d986ca2ccc69011d2fd9f8cd41a46f199b441e23/app/models/concerns/blacklight/solr/document.rb#L2

cbeer:
Hm, good question. If we're assuming rsolr is getting injected upstream, I guess so, right?

jcoyne:
@cbeer that seems reasonable.

cbeer:
@jcoyne: fixed up.

PR #1538, opened 2016-09-29
Support RSolr 2.x pre-release candidates

cbeer:


lib/blacklight/solr/repository.rb
```diff
@@ -52,7 +52,7 @@ def send_and_receive(path, solr_params = {})
     protected
 
       def build_connection
-        RSolr.connect(connection_config)
+        RSolr.connect(connection_config.merge(adapter: connection_config[:http_adapter]))
```
jcoyne: Will we want to refactor this when we drop rsolr 1.x support?

cbeer: Nope; we're currently passing that value through to RSolr, even though it's a Blacklight-specific configuration. Until RSolr 2.x, the value was simply ignored, but in 2.x, we need to remove it.

PR #1541, opened 2016-10-03
Prevent autocomplete from re-initializing multiple times

cbeer:


cbeer:
I don't think so; at least, not in our usage.

PR #1542, opened 2016-10-03
Right-align the per-page menu

cbeer:
Before:

screen shot 2016-10-10 at 11 13 49

After:

screen shot 2016-10-10 at 11 13 12


jcoyne:
Can you add a before & after view?

PR #1544, opened 2016-10-04
add i18n-keys to project for testing i18n key availability and use

mejackreed:
Adds to the CI build automated checking of i18n key comparisons.

### DIFF
--- a/blacklight.gemspec
+++ b/blacklight.gemspec
@@ -46,4 +46,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "simplecov"
   s.add_development_dependency "rubocop", '~> 0.49'
   s.add_development_dependency "rubocop-rspec", '~> 1.8'
+  s.add_development_dependency "i18n-tasks"
 end
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -1,120 +1,17 @@
 # i18n-tasks finds and manages missing and unused translations: https://github.com/glebm/i18n-tasks
-
-# The "main" locale.
 base_locale: en
-## All available locales are inferred from the data by default. Alternatively, specify them explicitly:
 locales: [es, fr, de, it, pt-BR]
-## Reporting locale, default: en. Available: en, ru.
-# internal_locale: en
-
-# Read and write translations.
 data:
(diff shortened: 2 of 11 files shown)


config/i18n-tasks.yml
```diff
@@ -0,0 +1,120 @@
+# i18n-tasks finds and manages missing and unused translations: https://github.com/glebm/i18n-tasks
+
+# The "main" locale.
+base_locale: en
+## All available locales are inferred from the data by default. Alternatively, specify them explicitly:
+locales: [es, fr, de, it, pt-BR]
```
cbeer: Inferred locales don't work because we're in an engine?

blacklight.gemspec
```diff
@@ -46,4 +46,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "simplecov"
   s.add_development_dependency "rubocop"
   s.add_development_dependency "rubocop-rspec"
+  s.add_development_dependency "i18n-tasks", "~> 0.9.5"
```
cbeer: Does it need to be this pessimistic?

config/i18n-tasks.yml
mejackreed: I think so? Tried them but it didn't pick them up.

blacklight.gemspec
mejackreed: Probably not. Push in a separate commit?

config/i18n-tasks.yml
```diff
@@ -0,0 +1,120 @@
...
+  # Locale files or `File.find` patterns where translations are read from:
+  read:
+    ## Default:
+    - config/locales/%{locale}.yml
+    ## More files:
+    - config/locales/**/*.%{locale}.yml
+    ## Another gem (replace %#= with %=):
+    # - "<%#= %x[bundle show vagrant].chomp %>/templates/locales/%{locale}.yml"
+
+  # Locale files to write new keys to, based on a list of key pattern => file rules. Matched from top to bottom:
```
cbeer: Can we clean up some of these default comments? I think it'd be easier to see our specific guidelines without these.

cbeer: Ok, sounds good.

PR #1545, opened 2016-10-07
V5 backports

mejackreed:


jcoyne:
Looks like you'll need the engine_cart upgrade too.

blacklight.gemspec
```diff
@@ -31,6 +31,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rspec-collection_matchers", ">= 1.0"
   s.add_development_dependency "capybara"
   s.add_development_dependency "poltergeist"
-  s.add_development_dependency 'engine_cart', '~> 0.8.0'
+  s.add_development_dependency 'engine_cart', '~> 0.10'
```
cbeer: '~> 1.0'?

cbeer:
I'd be ok with dropping the rails 4.0 build. The alternative, I guess, is to try to pin the devise dependency for that build.

Issue #1543, opened 2016-10-04
Translations missing


mejackreed:
| Locale | Key | Value in other locales or source |
| --- | --- | --- |
| de | blacklight.search.facets.missing | en [Missing] |
| it | blacklight.search.facets.missing | en [Missing] |
| pt-BR | blacklight.search.facets.missing | en [Missing] |

jcoyne:
@andresmrm would you be able to help us out with the Portuguese?

jcoyne:
@lpugin Can you help us translate "missing" (as in: "this facet is missing from the Solr result") to Italian?

cbeer:
I think `blacklight.search.facets.missing` is a little more nuanced than that -- rather more like "if you click on this label to see all documents for this search that have no values for the facet field"

andresmrm:
Sure! PR sent!

lpugin:
will do asap

barmintor:
@simosacchi might be able to help with it?

simosacchi:
@jcoyne @barmintor In the context/example provides above "is missing" can be translated in Italian as "manca" or "è mancante" or "non compare". (i.e. the English sentence "this facet is missing from the Solr result", it Italian would be something like "questa faccetta non compare nei risultati di Solr" or "questa faccetta manca dai risultati di Solr" ). 

Ping me if this is not what you were looking for.

barmintor:
Fixed in https://github.com/projectblacklight/blacklight/commit/463914d5c15be5b3182918e73d629ca447eed3f5

PR #1546, opened 2016-10-10
help with #1543

andresmrm:
- pt-BR	blacklight.search.bookmarks.absent	en Bookmark

I think this button is supposed to add something to bookmarks, right?
I would translate as "Adicionar aos Favoritos" (Add to Bookmarks) or maybe "Favorite" or "Favoritar".

- pt-BR	blacklight.search.view_title	en View results as:

Not sure about what is expected from this one... What is the context?
Maybe it's: 'Ver resultados como: '

- pt-BR	blacklight.search.zero_results.search_everything	en try searching everything

Again, not sure about the context here.
Translated as:
'tente buscar em tudo'
but maybe it's:
'tente buscar tudo'

- pt-BR	blacklight.sms.errors.carrier.invalid	en You must enter a valid carrier

Translated as 'Você precisa **usar** uma operadora válida', that would be equivalent to 'You must **use** a valid carrier'. **enter** could also be translated to **entrar**, but I think that isn't very clear. Maybe as **digitar** (type), or **selecionar** (select), depending on context.

### DIFF
--- a/config/locales/blacklight.pt-BR.yml
+++ b/config/locales/blacklight.pt-BR.yml
@@ -127,12 +127,14 @@ pt-BR:
         title: '%{value}'
         author: ' por %{value}'
         url: 'Link: %{url}'
+      success: 'SMS Enviado'
       errors:
         to:
           invalid: 'Você precisa digitar um número de telefone válido'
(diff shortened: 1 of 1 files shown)


cbeer:
👏 that makes sense to me, and we can always improve on it later. Thanks!

PR #1527, opened 2016-09-20
Move helper methods to presenters

jcoyne:

PR #1566, opened 2016-10-28
Use i18n translations for search and sort fields

cbeer:


cbeer:
This is somewhat unsatisfying and uncovered significant opportunities for future refactoring.

jcoyne:
Can we merge #1527 first?

cbeer:
I'd really like to get this in first, I think.. I'm optimistic it'll be an easy backport to into a 6.x release, but after #1527, I think it'll be somewhat more painful. I've added a quick refactor into this PR that may make it more appealing.

PR #1577, opened 2016-11-03
Deprecate link_to_search configuration in favor of link_to_facet

### LINKED ISSUE #1574: Rename "link_to_search" to "link_to_facet"
We should rename the config variable, because it creates a link to a facet.

https://github.com/projectblacklight/blacklight/blob/master/app/presenters/blacklight/rendering/link_to_facet.rb#L5

- [ ] allow `link_to_facet`
- [ ] deprecate `link_to_search`
- [ ] update wiki documentation

cbeer:
fixes #1574

### DIFF
--- a/lib/blacklight/configuration/facet_field.rb
+++ b/lib/blacklight/configuration/facet_field.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 module Blacklight
   class Configuration::FacetField < Blacklight::Configuration::Field
+    extend Deprecation
+
     def normalize! blacklight_config = nil
       self.query.stringify_keys! if self.query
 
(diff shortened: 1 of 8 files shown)


lib/blacklight/configuration/facet_field.rb
```diff
@@ -9,6 +11,11 @@ def normalize! blacklight_config = nil
       self.if = self.show if self.if.nil?
       self.index_range = 'A'..'Z' if self.index_range == true
 
+      if self.link_to_facet
```
jcoyne: This block should be link_to_search, right?

cbeer: Ugh. Find and replace wasn't scoped as expected.

PR #1583, opened 2016-11-03
Move base views into the shared directory

jcoyne:


cbeer:
I can't think of an easy way to deprecate this in 6.x, can you? I figure there's enough other view + css changes that it won't be a big deal just to do it and document it in the release notes.

mejackreed:
Curious about the reasoning here... purely organizational?

cbeer:
Yep; someone noted this at the summit today.

mejackreed:
Great, is there a place where we are collecting breaking changes for the 7.0 release notes?

cbeer:
Not yet; I'll toss a `breaking-change` label on it to remind us later.

PR #1589, opened 2016-11-30
Fixes #1587 by calling to_sym before comparison, when needed

### LINKED ISSUE #1587: Opensearch configuration bug
When I tried to configure my opensearch to use a different title_field from the index view, I ran into trouble.
In my catalog_controller.rb I wrote:
```
    ...
   config.view['opensearch'] = { title_field: 'my_title_field' }
    ...
end
```
but blacklight doesn't seem to mind it.
After a lot of digging I found out that the OpenStruct class from which config inherits, turns all of it's keys into symbols, while  blacklight uses a string to get the opensearch configuration, in search_helper.rb in the method get_opensearch_response:
   `field ||= blacklight_config.view_config('opensearch').title_field`
In order to work, this line should be:
   `field ||= blacklight_config.view_config(:opensearch).title_field`
Thank you all for constructing blacklight! It's a pleasure to use and develop with it!

atz:
Fixes #1587

Includes cleanup of comments.

Condensation (ternary ops, `||`)  is specifically because Rubocop would not allow even a single line of code to be added to this class.

### DIFF
--- a/lib/blacklight/configuration.rb
+++ b/lib/blacklight/configuration.rb
@@ -19,15 +19,12 @@ module Blacklight
 
     include Fields
 
-    # Set up Blacklight::Configuration.default_values to contain
-    # the basic, required Blacklight fields
+    # Set up Blacklight::Configuration.default_values to contain the basic, required Blacklight fields
     class << self
       def default_values
         @default_values ||= begin
           {
-          ##
           # === Search request configuration
-          ##
           # HTTP method to use when making requests to solr; valid
           # values are :get and :post.
           http_method: :get,
@@ -37,10 +34,9 @@ module Blacklight
           default_solr_params: {},
           ##
           # === Single document request configuration
-          ##
           # The solr rqeuest handler to use when requesting only a single document
           document_solr_request_handler: nil,
-          # THe path to send single document requests to solr
+          # The path to send single document requests to solr
           document_solr_path: 'get',
           document_unique_id_param: :ids,
           # Default values of parameters to send when requesting a single document
@@ -61,7 +57,6 @@ module Blacklight
           connection_config: nil,
           ##
           # == Blacklight view configuration
-          ##
           navbar: OpenStructWithHashAccess.new(partials: {}),
           # General configuration for all views
           index: ViewConfig::Index.new(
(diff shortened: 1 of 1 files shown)


lib/blacklight/configuration.rb
```diff
@@ -288,15 +276,12 @@ def deep_copy
...
     # Get a view configuration for the given view type
     # including default values from the index configuration
+    # @param [Symbol,#to_sym] view_type
     def view_config(view_type)
-      if view_type == :show
-        self.index.merge self.show
-      else
-        self.index.merge view.fetch(view_type, {})
-      end
+      view_type = view_type.to_sym unless view_type.is_a? Symbol
```
atz: There is a comparison happening in this method, so logic would require it up front.  It certainly could be done at lower levels as well.

atz:
The revisions are revisions I would make anyway.  My comment was to explain why they appear *here* rather than separately.

lib/blacklight/configuration.rb
cbeer: By this, I simply meant the symbolification.

atz:
OK, I don't see anything worth impeding this over.

atz:
@projectblacklight/team-blacklight This branch includes fixes that unblock all other CI builds.

lib/blacklight/configuration.rb
mejackreed: If we do this here, then aren't necessitated to do this everywhere that may want to use `view_type`? Why not just pop this down as @cbeer suggests?

atz: @mejackreed Yes, it is is quite plausible that it is (already) required elsewhere, too. This PR addresses the concrete actual problem in #1587.  I support doing it at a lower level **also**. But that is not necessary to resolve 1587 AND is insufficient to allow consumers of `view_config` to use symbols and strings interchangeably.  Ticket that separately if you want me to work on it.

As I said in response to Chris, we cannot *only* do `to_sym` at the lower level if you want the comparison in the very next line to still be valid for both string and symbol arguments, that is to say, specifically, if we want to treat these lines as equivalent: 
```ruby
view_config('show')
view_config(:show)
```

I hear you saying that you want that: this provides it.

atz:
@mejackreed you say you aren't requesting changes, but github says you are.

atz:
@cbeer, I agree doing it at the lower level will catch more similar cases.  As I said, I agree with doing it there **also**.  That you are still talking about it as a "review" of _this_ change suggests you haven't read or understood my replies.  Unless you want these two lines to be different:
```ruby
view_config('show')
view_config(:show)
```
then you need to do it here, no matter what else you do.  Since this is the 3rd time I've explained it in this thread without any acknowledgement, I will proceed with merging this.

PR #1595, opened 2016-12-03
Blacklight::OpenStructWithHashAccess should expose a HashWithIndifferentAccess

cbeer:


cbeer:
Too bad about all the side-effects, though. I'm not sure it's worth it. It's probably easier just to declare configuration keys are symbols.

atz:
@cbeer: since we have/had apps using non-symbol keys, is it worth it detect those and warn/fail?  Do we need a deprecation release or can we just declare that was always the intent?

barmintor:
Reimplementing on master (or even using  HashWithIndifferentAccess) raises errors in presenters around valid keys in the interpolation helpers. I'm going to bump this to 8.x, since it's a more complicated change now than it may have once been.

PR #1599, opened 2017-01-03
Specify Blacklight logger for logging

tampakis:
When the action_view logger is disabled, Blacklight throws an error because the logger is set to nil.
(Anyone else try to suppress the dozens of "Rendered /blacklight-6.7.2/app/views/catalog/_bookmark_control.html.erb (0.6ms)" statements in the logs per request?)

### DIFF
--- a/lib/blacklight/solr/response.rb
+++ b/lib/blacklight/solr/response.rb
@@ -91,7 +91,7 @@ class Blacklight::Solr::Response < ActiveSupport::HashWithIndifferentAccess
         value.each { |v| force_to_utf8(v) }
       when String
         if value.encoding != Encoding::UTF_8
-          Rails.logger.warn "Found a non utf-8 value in Blacklight::Solr::Response. \"#{value}\" Encoding is #{value.encoding}"
+          Blacklight.logger.warn "Found a non utf-8 value in Blacklight::Solr::Response. \"#{value}\" Encoding is #{value.encoding}"
           value.dup.force_encoding('UTF-8')
         else
           value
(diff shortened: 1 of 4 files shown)


cbeer:
I'm curious -- how/why are you disabling the view logger?

tampakis:
We were interested in getting info-level logging for everything except the action_view because it prints out a line for each partial rendered on a page. I originally set the action_view logger to nil, which caused the application to error when calling the logging debug statements I changed in this PR.

However instantiating a new logger with a different log level seemed to be a more careful solution: https://github.com/pulibrary/orangelight/pull/815.

I think logging coming out of Blacklight should be consistent so I updated the PR to do so.

cbeer:
I agree that consistency is important and that it isn't clear when using `Blacklight.logger` is appropriate. My gut feeling is things in `./app` should probably use the Rails default `logger` to make it fit with the expected Rails behavior, while things in `./lib` should generally use `Blacklight.logger` as they might be used in other contexts.

jcoyne:
I agree with @cbeer, we want these messages in the action_view logger, not the top level rails logger.

lib/blacklight/solr/response.rb
```diff
@@ -91,7 +91,7 @@ def force_to_utf8(value)
         value.each { |v| force_to_utf8(v) }
       when String
         if value.encoding != Encoding::UTF_8
-          Rails.logger.warn "Found a non utf-8 value in Blacklight::Solr::Response. \"#{value}\" Encoding is #{value.encoding}"
+          Blacklight.logger.warn "Found a non utf-8 value in Blacklight::Solr::Response. \"#{value}\" Encoding is #{value.encoding}"
```
cbeer: Although this change looks good. Maybe there are other inconsistent cases too?

PR #1601, opened 2017-01-04
Use a valid version of rails for jruby. Remove duplicate matrix entry

jcoyne:


mejackreed:
Yet the jruby build is still failing?

PR #1609, opened 2017-01-09
Use a fallback for locating blacklight svg icons in production

cbeer:


app/models/blacklight/icon.rb
```diff
@@ -41,7 +41,7 @@ def file_source
     private
 
     def file
-      Rails.application.assets.find_asset(path)
+      (Rails.application.assets || ::Sprockets::Railtie.build_environment(Rails.application)).find_asset(path)
```
jcoyne: Can you write a comment about why we have this conditional here?

cbeer: 👌

PR #1610, opened 2017-01-09
Fix some styles for Bootstrap 4 alpha 6

jcoyne:


app/assets/stylesheets/blacklight/_facets.scss
```diff
@@ -12,9 +12,16 @@
   line-height: inherit;
 }
 
+.facet-limit {
+  margin-bottom: 5px;
+
+  .card-block {
+    padding: 1rem;
```
cbeer: Are there bootstrap variables we can use for these?

PR #1625, opened 2017-01-31
Refactor SearchHelper module to SearchService class

jcoyne:


app/controllers/concerns/blacklight/bookmarks.rb
```diff
@@ -136,6 +134,10 @@ def clear
 
   protected
 
+  def search_service
+    Blacklight::SearchService.new(blacklight_config)
```
jkeck: Would it be useful to make the search service class settable?  Just thinking about if an app or a gem might want to provide their own search service class to be used.

Totally understand if that seems premature though.  Fine with letting the use case reveal itself later.

cbeer: I think `BookmarksController < CatalogController`. Do we need to do this explicitly, or can it just be inherited?

app/services/blacklight/search_service.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    # TODO: Can this be private?
+    attr_reader :blacklight_config
+
+    # a solr query method
+    # @param [Hash] user_params ({}) the user provided parameters (e.g. query, facets, sort, etc)
+    # @yield [search_builder] optional block yields configured SearchBuilder, caller can modify or create new SearchBuilder to be used. Block should return SearchBuilder to be used.
+    # @return [Blacklight::Solr::Response] the solr response object
+    def search_results(user_params)
```
cbeer: Do you think it would make sense to add these `user_params` (also called `request_params`, or `paging_params`?) as an argument to the initializer?

app/services/blacklight/search_service.rb
```diff
@@ -0,0 +1,120 @@
...
+# SearchService returns search results from the repository
+module Blacklight
+  class SearchService
+    include Blacklight::RequestBuilders
+
+    def initialize(blacklight_config)
+      @blacklight_config = blacklight_config
+    end
+
+    # TODO: Can this be private?
```
cbeer: Should we address this TODO now?

app/services/blacklight/search_service.rb
```diff
@@ -0,0 +1,120 @@
...
+    ##
+    # The key to use to retrieve the grouped field to display
+    def grouped_key_for_results
+      blacklight_config.index.group
+    end
+
+    delegate :repository_class, to: :blacklight_config
+
+    def repository
+      repository_class.new(blacklight_config)
```
cbeer: This strikes me as awkward. Could we add it an accessor on `blacklight_config` instead?

cbeer: I think we also have some use cases for varying the `repository` based on user parameters (previously supported by overriding the method in the controller). I don't think this architecture necessarily precludes that (especially if we can pull the user params up to an attribute), but it would be good to think that through more.

app/services/blacklight/search_service.rb
jcoyne: This requires a change to the signature of `SearchBuilder`, so I'd prefer to make that a separate PR. Presently the `SearchBuilder` is delegating `blacklight_config` to it's passed in context which is an instance of `SearchService` at this time.

cbeer: 👌 Does it make sense to add that comment to the TODO, or do you think it's a near-term sort of change?

jcoyne: I think we can do it in the short term (before release)

app/services/blacklight/search_service.rb
```diff
@@ -0,0 +1,118 @@
...
+    # Get the solr response when retrieving only a single facet field
+    # @return [Blacklight::Solr::Response] the solr response
+    def facet_field_response(facet_field, extra_controller_params = {})
+      query = search_builder.with(user_params).facet(facet_field)
+      repository.search(query.merge(extra_controller_params))
+    end
+
+    # Get the previous and next document from a search result
+    # @return [Blacklight::Solr::Response, Array<Blacklight::SolrDocument>] the solr response and a list of the first and last document
+    def previous_and_next_documents_for_search(index, request_params, extra_controller_params = {})
```
cbeer: We're still passing `request_params` here. I think it's trickier, since we're not building these results in response to the current url params. That strikes me as potentially a higher-level failing. I wonder if there's anything we should do to call this out?

My other thought'd be do something like:

```ruby
search_service.user_params(ActiveSupport::HashWithIndifferentAccess.new(current_search_session.query_params).previous_and_next_documents_for_search(index)
```

(where `#user_params` creates a new `SearchService` using the provided params instead). At least that'd push the consistency a little higher, maybe?

atz: @cbeer do you intend this to be a blocker or just a point of coversation?  My inclination is to approve/merge.

cbeer: I would like to give @jcoyne a chance to respond first.

jcoyne: I think `previous_and_next_documents_for_search` *might* belong in a totally different service.  I don't like the idea of a `user_params` setter much.  That has too much mutable state for my liking, and it would only be used when invoking this one method.

PR #1632, opened 2017-02-23
Suppress the document tools if none are configured.

cbeer:


jcoyne:
@cbeer can this be backported to 6.x?

PR #1634, opened 2017-02-28
No need to style sr-only elements

jcoyne:


app/assets/stylesheets/blacklight/_constraints.scss
```diff
@@ -2,7 +2,7 @@
   @extend .mb-2;
 }
 
-span.constraints-label {
+.constraints-label {
```
cbeer: Is this only used to style this `.sr-only` element?

https://github.com/projectblacklight/blacklight/blob/e7d7cb883e4d24a5263797cac47a8e5fd5b603ca/app/views/catalog/_constraints.html.erb#L4

jcoyne: Yes, as far as I can tell.

cbeer: Can we just remove the style entirely? No reason to make some sr-only text bold, is there?

jcoyne: Yes, that's best. We didn't have sr-only in the 6 branch, so I didn't notice.

atz:
@jcoyne, @cbeer So we can close this or rework it?

PR #1637, opened 2017-03-02
Consistently differentiate between the user-facet facet key and the solr field name

cbeer:
Fixes https://github.com/projectblacklight/spotlight/pull/1762


jcoyne:
Let me know when you're done

cbeer:
Yep. I want to step back and see if there's a way to refactor things to avoid the same mistake.. and also audit the code to make sure we catch all the places.

cbeer:
I think this is ready. No obvious refactoring presented itself, so I just added some tests for this feature.

PR #1640, opened 2017-03-06
CatalogHelper#render_search_to_page_title: prettify empty results

dunn:
Previously if no constraints were provided the `<title>` would end up
with a dash at the beginning: `— #{application_name} Search Results`

### DIFF
--- a/app/helpers/blacklight/catalog_helper_behavior.rb
+++ b/app/helpers/blacklight/catalog_helper_behavior.rb
@@ -306,6 +306,7 @@ module Blacklight::CatalogHelperBehavior
       constraints += params['f'].to_unsafe_h.collect { |key, value| render_search_to_page_title_filter(key, value) }
     end
 
+    return 'All results' if constraints.empty?
     constraints.join(' / ')
   end
 
--- a/spec/helpers/catalog_helper_spec.rb
+++ b/spec/helpers/catalog_helper_spec.rb
@@ -199,7 +199,7 @@ describe CatalogHelper do
       allow(document).to receive_messages(has?: true)
       expect(helper.has_thumbnail? document).to be true
     end
-    
+
     it "does not have a thumbnail if the thumbnail_field is missing from the document" do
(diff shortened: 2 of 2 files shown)


dunn:
I guess the one problem with this is that if someone's customized `blacklight.search.page_title.title` to move `constraints`, this could look pretty weird (think `Results for “All results”`)

dunn:
@mejackreed also pointed out this would require work for internationalization too, so I'm gonna rethink it.

cbeer:
Ok, let us know if we can lend a hand. I think your approach seems reasonable, and hopefully it's just a matter of getting the right i18n keys in place.

You might consider extracting https://github.com/projectblacklight/blacklight/blob/e7d7cb883e4d24a5263797cac47a8e5fd5b603ca/app/views/catalog/_search_results.html.erb#L3 into a new helper method that uses different i18n keys for zero vs any constraints?

dunn:
If the result of that helper method is still used as the value for `constraints` in `search.page_title.title`, then we've got `constraints` playing a different role depending on whether it's zero or non-zero constraints.  With non-zero it's the constraints, but with zero it's a message about the lack of constraints, so we can't interpolate it anywhere; e.g., `Results for %{constraints}` only works when there's at least one constraint.

The easiest change I can think of is to just change the default value to be that, i.e.:
```yaml
    search:
      page_title: 
        title: 'Results for "%{constraints}" - %{application_name}'
```
Not perfect, but `Results for ""` is at least obvious about what it means.

cbeer:
What about giving the implementor full control over the label, maybe like:

```
search:
  page_title:
    title:
      with_constraints:  '%{constraints} - %{application_name} Search Results'
      without_constraints: 'All results - %{application_name}'
```

or something akin to the pluralization support (one vs many):

```
search:
  page_title:
    title:
      zero:  '%{constraints} - %{application_name} Search Results'
      any: 'All results - %{application_name}'
```

or some backwards-compatible variation on that?

dunn:
Either seems OK.  Is the way to make it backwards-compatible adding new keys, setting the old one to an empty string and deprecating it, and using the new keys unless the old one is set by the user?

PR #1649, opened 2017-03-27
Allow customization of <title> for searches with no constraints

dunn:
**WIP:** the non-English localization files need to be updated.

**************

Following up on #1640, and @cbeer's proposal in https://github.com/projectblacklight/blacklight/pull/1640#issuecomment-285477106, this deprecates `blacklight.search.page_title.title` and replaces it with two i18n keys:

- `blacklight.search.page_title.with_constraints`
- `blacklight.search.page_title.without_constraints`

This allows more natural strings for the `<title>` of all types of searches.


app/helpers/blacklight/catalog_helper_behavior.rb
```diff
@@ -289,6 +289,24 @@ def render_search_to_page_title_filter(facet, values)
...
+    end
+
+    return t('blacklight.search.page_title.without_constraints') unless query_has_constraints?(params)
+
+    t('blacklight.search.page_title.with_constraints',
+      :constraints => render_search_to_page_title(params),
+      :application_name => application_name)
+  end
+
   def render_search_to_page_title(params)
```
dunn: This method should really be called something like `render_search_constraints`, but renaming it would have to wait for a major version update.  Should I add a deprecation notice now?

dunn:
Fixing the rubocop stuff in #1650

PR #1656, opened 2017-03-31
[WIP] Pin to a working release of rubocop

jcoyne:


.solr_wrapper.yml
```diff
@@ -3,3 +3,4 @@
 collection:
   dir: solr/conf/
   name: blacklight-core
+version: 6.4.0
```
cbeer: 👎 . I'd like  usto spend a little more time today figuring out what is wrong.

jcoyne: This wasn't intended to be merged. It was an experiment to show that the problem was indeed Solr 6.5.0

atz: Do you want to close this PR or remove this edit?

PR #1662, opened 2017-04-05
Use rails style IndentationConsistency cop

jcoyne:


mejackreed:
Why? I'm 👎  b/c it already appears to be a consistent style.

jcoyne:
@mejackreed then we should set it to `EnforcedStyle: normal`, right?

jcoyne:
@mejackreed I think we should either merge this PR or https://github.com/projectblacklight/blacklight/pull/1663.  Let's pick a style and stick to it, not half-and-half.

mejackreed:
My personal preference is to be against large scale style only changes. I think its a good idea to have a style guide and enforce it, but if the tools constantly cause us to thrash our code for style, I'm more inclined to leave the todos or pin the version. (I realize this is a personal preference and potentially not the view of the Blacklight community).

jcoyne:
@mejackreed enforcing a style will potentially prevent this sort of thrashing, where I do it one way and chris does it another way.  If we pick something, we only need to make the change one time and forever after we never need to think about it.

mejackreed:
@jcoyne for sure! Can the style be adopted though w/out the wholesale changes? That is really the heart of my pain with these types of changes.

jcoyne:
@mejackreed I think that's the thing about rubocop, I don't think there is a way to use it just on changed lines.

jcoyne:
We merged #1663 instead.

PR #1663, opened 2017-04-05
Enforcing indentation consistency

jcoyne:


app/controllers/concerns/blacklight/controller.rb
```diff
@@ -52,134 +52,134 @@ def default_catalog_controller
...
+  def search_facet_url options = {}
+    opts = search_state.to_h.merge(action: "facet").merge(options).except(:page)
+    url_for opts
+  end
+  deprecation_deprecate search_facet_url: 'Use search_facet_path instead.'
 
-    def search_facet_url options = {}
-      opts = search_state.to_h.merge(action: "facet").merge(options).except(:page)
-      url_for opts
+  def search_facet_path(options = {})
```
atz: These are semantic changes mixed in here with your whitespace edits.  What gives?

jkeck: @atz what do you mean by semantic  changes?  I think the diff is perhaps misleading.   Take a look below.

atz: @jkeck: you're right, the diff highlighting in github made it look like the method name and implementation changed.

PR #1579, opened 2016-11-03
Add Rails migration compatibility version to migrations

cbeer:
Fixes #1576

PR #1672, opened 2017-04-24
Test on Rails 5.1.0.rc2

jcoyne:


cbeer:
I guess we need to chose between supporting Rails 4.2 and Rails 5.1, huh?

jcoyne:
@cbeer if we want to take on a lot of work with converting our migrations to erb templates and sniffing for Rails versions, we could do both.  However, I think this is a good place to make a break from 4.2.

jcoyne:
@cbeer here's the example of that https://github.com/plataformatec/devise/blob/master/lib/generators/active_record/templates/migration.rb#L1

jcoyne:
Moved this work to #1579

PR #1674, opened 2017-04-24
Disable RSpec monkeypatching

jcoyne:


atz:
Attempted resolution of conflict...

PR #1676, opened 2017-04-24
Sanitize the user provided sort parameter

### LINKED ISSUE #1673: Filter sort input values
When someone submits a query like `'sort'=>'/etc/passwd'`, this results in an invalid query to solr:
```
'error'=>{'msg'=>'Can\'t determine a Sort Order (asc or desc) in sort spec \'/etc/passwd\', pos=11','code'=>400}}
```

jcoyne:
Fixes #1673

### DIFF
--- a/lib/blacklight/search_builder.rb
+++ b/lib/blacklight/search_builder.rb
@@ -214,6 +214,10 @@ module Blacklight
       @facet
     end
 
+    # Decode the user provided 'sort' parameter into a sort string that can be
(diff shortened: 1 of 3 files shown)


jcoyne:
This should prevent errors in SearchWorks like https://app.honeybadger.io/projects/50022/faults/33368770

cbeer:
👍 I'm pretty sure this was just backwards-compatibility cruft we've kept alive.

PR #1681, opened 2017-04-27
Deprecate default_per_page helper

jcoyne:
It was doing the same thing the configuation was already doing.


cbeer:
This is a little more than simple deprecation -- it's dropping support altogether if implementors have overridden the helper method as we're no longer calling the old method at all. I'm not sure if there's a good way to be graceful, or if we should just call this out in the 7.x release notes?

jcoyne:
@cbeer I think just putting this in the release notes for 7.x would be acceptable.

PR #1687, opened 2017-04-27
Add 'display_control' option to view configs to allow them to toggle …

cbeer:
…on/off visibility in the view type widget

(Turns out `display` is a reserved word for ruby objects? I'm still not exactly sure where that comes from.)

### DIFF
--- a/app/helpers/blacklight/configuration_helper_behavior.rb
+++ b/app/helpers/blacklight/configuration_helper_behavior.rb
@@ -134,6 +134,13 @@ module Blacklight::ConfigurationHelperBehavior
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/views/catalog/_view_type_group.html.erb_spec.rb
```diff
@@ -37,6 +37,21 @@
     expect(rendered).to have_selector('.view-type-c', :text => 'c')
   end
 
+  it "hides views that are set to not display" do
```
jcoyne: I'd prefer that this is converted to a helper test as that's where all the logic resides.

PR #1692, opened 2017-04-28
Backport: Pass controller instance to SearchState

jcoyne:
This allows subclasses to access the url_helpers

### DIFF
--- a/app/controllers/concerns/blacklight/controller.rb
+++ b/app/controllers/concerns/blacklight/controller.rb
@@ -67,7 +67,15 @@ module Blacklight::Controller
 
     # @return [Blacklight::SearchState] a memoized instance of the parameter state.
     def search_state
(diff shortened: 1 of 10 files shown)


app/controllers/concerns/blacklight/controller.rb
```diff
@@ -67,7 +67,7 @@ def render_saved_searches?
 
     # @return [Blacklight::SearchState] a memoized instance of the parameter state.
     def search_state
-      @search_state ||= search_state_class.new(params, blacklight_config)
+      @search_state ||= search_state_class.new(params, blacklight_config, self)
```
cbeer: Ought this have an arity check to ensure backwards compatibility?

jcoyne: I've added one.

PR #1700, opened 2017-05-10
Adjust responsive layout

ggeisler:
This PR removes the sidebar column at the "small" viewport size, making the `sm` viewport layout similar to the `xs` layout. 

Rationale: Currently when you view the application at the small viewport size the sidebar column is shown but the facets are collapsed. Clicking on the expand icon simply shows the facets in the previously empty sidebar column. In other words, there's no reason to collapse the facets with this arrangement.

before-facet-toggle

One approach would be to remove the collapse behavior and just show the facets in the sidebar at the small viewport size. However, I wonder how useful the side-by-side arrangement actually is to a user at the small viewport size. Both the facets and the main content area are fairly squished at 577 to 767px:

before-sm

Alternatively, we could just use the same no-sidebar arrangement currently used in the `xs` viewport for the `sm` viewport as well. That's all this PR does (though someone more familiar with the layout decisions made in Blacklight should definitely make sure I'm not overlooking some negative ramifications of this approach). So with this PR the sidebar is not displayed until you get to the `md` viewport; until then the facets are shown in the expand/collapse section above the main content area:

after-facet-toggle

### DIFF
--- a/app/helpers/blacklight/layout_helper_behavior.rb
+++ b/app/helpers/blacklight/layout_helper_behavior.rb
@@ -22,14 +22,14 @@ module Blacklight
     # Classes used for sizing the main content of a Blacklight page
     # @return [String]
     def main_content_classes
-      'col-md-9 col-sm-8'
+      'col-md-9'
     end
 
     ##
     # Classes used for sizing the sidebar content of a Blacklight page
     # @return [String]
     def sidebar_classes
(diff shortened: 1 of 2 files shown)


cbeer:
Makes sense to me. I wonder if we need to do something better with the collapsed "Limit your search" header and toggle to be clear that something useful happens when you click it? For some reason, I felt more comfortable knowing only very small screens got to see that.

PR #1701, opened 2017-05-10
Prevent logo and toggle icon in topbar from overlapping

ggeisler:
At the smallest (`xs`) viewport size, the Blacklight logo and the toggle icon for expanding the utility links overlap:

before-logo toggler-overlap

This PR prevents that overlap by enabling the logo to align left and the toggle icon to align right:

after-facet-toggle

To do this I added a unique class to the `nav` element used for the topbar. I could also do it without adding that class (just using a `.navbar > .container` selector), which works fine for Blacklight, but I'm worried that approach could have unintended consequences for downstream applications that add additional `.navbar` elements with nested `.container`s. Seems safer to just add a class to the `nav` element so the CSS is scoped to that specific use of the `nav`.

### DIFF
--- a/app/assets/stylesheets/blacklight/_header.scss
+++ b/app/assets/stylesheets/blacklight/_header.scss
@@ -2,6 +2,13 @@
   Header bar at top (Bootstrap nav-bar)
 */
 
+.topbar > .container {
+  @media (max-width: breakpoint-max(xs)) {
+    margin-left: 0;
+    margin-right: 0;
+  }
+}
+
 .navbar-brand {  /* The main logo image for the Blacklight instance */
   @if $logo-image {
     background: transparent $logo-image no-repeat top left;
@@ -46,6 +53,6 @@
     // hide 'search' label at very small screens
     @media screen and (max-width: breakpoint-max(xs)) {
       @include sr-only();
-    }    
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


cbeer:
Weird. I wonder if we screwed up our navbar markup to cause this behavior?

ggeisler:
Sure @cbeer. I couldn't see anything obvious but you probably will have a better eye than me. It does seems like something that should just work correctly without my fix.

cbeer:
It wouldn't be the first time we screwed up something across bootstrap v4 alpha releases. I'll try to get to this yet this week.

ggeisler:
@cbeer Just checking to see if you've made any progress on investigating this issue? Not critical but it would be good to be able to close our associated ticket in ArcLight by the end of the work cycle next week.

cbeer:
I think we must be doing something wrong in https://github.com/projectblacklight/blacklight/blob/e11ea1afd1056ad11fda9ee7fb73e415f4c25e67/app/assets/stylesheets/blacklight/_header.scss#L15-L25. I wonder why we're messing around with background images instead of just using an `img` tag like the upstream docs suggest. @jkeck?

mejackreed:
I looked into it a bit this morning and couldn't figure it out. I tried the `img` tags and other approaches but ran into the same issue. Without a parent `.container` it looks fine on mobile, but doesn't work at large screen. /shrug

PR #1703, opened 2017-05-11
Deprecate facet_list_limit

jcoyne:


lib/blacklight/solr/search_builder_behavior.rb
```diff
@@ -185,6 +185,7 @@ def add_facet_paging_to_solr(solr_params)
       solr_params[:"facet.field"] = with_ex_local_param(facet_ex, facet_config.field)
 
       limit = if scope.respond_to?(:facet_list_limit)
+                Deprecation.warn(self, "The use of facet_list_limit is deprecated and will be removed in 7.0")
```
cbeer: In the deprecation message, can you also recommend using the `more_limit` option in the field configuration?

cbeer: (I also wonder if the hard-coded `20` below ought to be pushed to the engine configuration and we could recommend setting that as well)

PR #1646, opened 2017-03-07
Pass controller instance to SearchState

jcoyne:
This allows subclasses to access the url_helpers

Alternative approach to #1645

PR #1704, opened 2017-05-11
Change signature of SearchBuilder

jcoyne:


jcoyne:
This should be rebased after #1702 is merged.

jcoyne:
@cbeer I hear that, but I'm even less excited about having an object that looks like a Controller but is not, in order to use the SearchService.

Another approach is for the plugin to override `Catalog#search_service`. But then the plugin must have a custom subclass of SearchService, and of SearchBuilder.  That seems unnecessarily complex to me.

jcoyne:
@cbeer how about creating a parameter object that encapsulates blacklight_config and may be subclassed to also hold `current_ability`?

cbeer:
Fixed by #1646, probably?

jcoyne:
@cbeer Uh, it's been a long time... probably?

PR #1712, opened 2017-05-31
Adds an explicit `#empty?` method for the GroupResponse

### LINKED ISSUE #1711: Grouping logic not working
The logic to show a grouping result is not working.

To illustrate, for a given search result:

https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/_search_results.html.erb#L20-L26

Calls `@response.empty?` which will eventually call `total` on `Blacklight::Solr::Response::GroupResponse`. This logic does not seem to evaluate correctly. Evaluates `total` on `@response` not on this instance.

mejackreed:
This is needed as the metaprogramming method missing calls `#empty`
on the `@response` object which in turn has the incorrect scope to
correctly evaluate the total.

Closes #1711

### DIFF
--- a/lib/blacklight/solr/response/group_response.rb
+++ b/lib/blacklight/solr/response/group_response.rb
@@ -32,6 +32,13 @@ class Blacklight::Solr::Response::GroupResponse
     params[:start].to_s.to_i
   end
 
+  ##
+  # Relying on a fallback (method missing) to @response is problematic as it
+  # will not evaluate the correct `total` method.
+  def empty?
+    total.zero?
+  end
+
   def method_missing meth, *args, &block
     if response.respond_to? meth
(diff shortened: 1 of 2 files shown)


lib/blacklight/solr/response/group_response.rb
```diff
@@ -32,6 +32,14 @@ def start
     params[:start].to_s.to_i
   end
 
+  ##
+  # Relying on an included super method fails here as it will not evaluate the
```
cbeer: I don't understand what you mean by "an included super method"?

mejackreed: Yeah that wasn't clear at all was it! I've updated the comment, let me know if there is better wording you prefer

PR #1713, opened 2017-05-31
use correct local variable for label_for_search_field and sort_field_label

drh-stanford:
This PR corrects a `NameError` when these methods are called -- it cannot find the local variable/method `field` in these 2 methods.

```
NameError:
       undefined local variable or method `field' for #<#<Class:0x007fe047e32770>:0x007fe048f58270>
       Did you mean?  fields
```


cbeer:
Could you write a test for this case (which was apparently uncovered before?)

drh-stanford:
Ok, I've added tests for those 2 methods.

PR #1736, opened 2017-08-06
Remove Saved Searches feature

### LINKED ISSUE #520: Saved Searches UX concern

cbeer:
Fixes #520 

The feature was buried in the UX and not more useful than e.g. browser bookmarks since our blacklight search urls are persistent URLs anyway.


mark-dce:
@cbeer @jcoyne @jkeck Can anyone tell me if the actual tracking was removed from the application or just the UI for saved searches.  We're seeing a really huge `searches` table in one of our Hyrax applications, and as best I can tell, we don't want it or need it.  Is there an easy way for us to turn off search history tracking in an application that depends on Blacklight?

jkeck:
@mark-dce this PR doesn't stop tracking all-together.  However; @mbklein submitted #1720 which can probably solve your problem.  I don't have a good implementation to point to, but effectively you can configure a crawler detector that could just always return `true` so that all traffic is viewed as a crawler and no searches are serialized.

PR #1742, opened 2017-08-11
Use 'ch' units for determining the width of the facet counts

cbeer:


app/assets/javascripts/blacklight/facet_load.js
```diff
@@ -9,11 +9,11 @@
 
     $('ul.facet-values, ul.pivot-facet').each(function(){
       var longest = $(this).find('span.facet-count').sort(longer).first();
-      var clone = longest.clone()
-        .css('visibility','hidden').css('width', 'auto');
-      $('body').append(clone);
-      $(this).find('.facet-count').first().width(clone.width());
-      clone.remove();
+      
+      if (longest && longest.textContent) {
```
jkeck: `textContent` is the native API I believe?  So we would need either `longest[0].textContent` or use jQuery's `longest.text()`?

jcoyne: I'd prefer we begin to move away from being coupled to jQuery, so I much prefer the former.

cbeer: Thanks (I guess with our test data, it makes no visual difference whether this JS executes or not). I just use `[0]` instead of `.first()`and it should all work out, I think?

PR #1747, opened 2017-08-18
Add marc_display field to the default fl list

cbeer:


jcoyne:
One test needs to be adjusted.

PR #1751, opened 2017-08-25
Remove option of passing false as the second argument to thumbnail_tag

jcoyne:
It's much more clear to the reader what the intention of this code is:

```ruby
thumbnail_tag({}, suppress_link: true)
```
as compared to the more obtuse:
```ruby
thumbnail_tag({}, false)
```

Deprecated in 6.x in #1752


cbeer:
Do you want to deprecate it in `master`, or back in `release-6.x` so we can remove it entirely in master?

jcoyne:
@cbeer I can do `release-6.x`, I was unsure if we were planning more releases.

PR #1753, opened 2017-08-27
Javascript style fixes

jcoyne:
This fixes a number of jshint warnings

### DIFF
--- a/app/assets/javascripts/blacklight/autofocus.js
+++ b/app/assets/javascripts/blacklight/autofocus.js
@@ -1,6 +1,6 @@
 //= require blacklight/core
 (function($) {
-  Blacklight.do_search_autofocus_fallback = function() {
+  Blacklight.doSearchAutofocusFallback = function() {
     if (typeof Modernizer != "undefined") {
       if (Modernizr.autofocus) {
         return;
@@ -11,6 +11,6 @@
   }
 
   Blacklight.onLoad(function() {
-    Blacklight.do_search_autofocus_fallback();
+    Blacklight.doSearchAutofocusFallback();
   });
 })(jQuery);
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/blacklight/autofocus.js
```diff
@@ -1,7 +1,7 @@
 //= require blacklight/core
 (function($) {
-  Blacklight.do_search_autofocus_fallback = function() {
-    if (typeof Modernizer != "undefined") {
+  Blacklight.doSearchAutofocusFallback = function() {
```
mejackreed: Should we try to notify users with console warnings for 6.x deprecation messages for the method renames?

cbeer: I guess we could do something like:

```
Blacklight.doSearchAutofocusFallback = function() {
if(typeof Blacklight.do_search_autofocus_fallback == "function") {
  console.log("DEPRECATED")
  return Blacklight.do_search_autofocus_fallback();
}
...
}
```

Right?

mejackreed: I was thinking something like:

```javascript
Blacklight.do_search_autofocus_fallback = function() {
  console.warn("DEPRECATED: Blacklight.do_search_autofocus_fallback is deprecated and will be removed in version 7.0 This method has been renamed to Blacklight.doSearchAutofocusFallback")
  Blacklight.doSearchAutofocusFallback();
})
```

cbeer: I thought about that, but `do_search_autofocus_fallback` is a method that people may have overridden. I suppose we might need both, in case someone is calling those methods directly?

mejackreed: Err.. yeah I guess that could be problematic.

jcoyne: So, we can do that in a backport?

PR #1755, opened 2017-08-27
Move instance variables to search_session

jcoyne:
Also removed `@search_context_response` which was not used.

### DIFF
--- a/app/controllers/concerns/blacklight/search_context.rb
+++ b/app/controllers/concerns/blacklight/search_context.rb
@@ -111,9 +111,7 @@ module Blacklight::SearchContext
       response, documents = search_service.previous_and_next_documents_for_search index, search_state.reset(current_search_session.query_params).to_hash
 
       search_session['total'] = response.total
-      @search_context_response = response
-      @previous_document = documents.first
-      @next_document = documents.last
+      { prev: documents.first, next: documents.last }
     end
   rescue Blacklight::Exceptions::InvalidRequest => e
     logger.warn "Unable to setup next and previous documents: #{e}"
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/blacklight/search_context.rb
```diff
@@ -111,9 +111,8 @@ def setup_next_and_previous_documents
       response, documents = search_service.previous_and_next_documents_for_search index, search_state.reset(current_search_session.query_params).to_hash
 
       search_session['total'] = response.total
-      @search_context_response = response
-      @previous_document = documents.first
-      @next_document = documents.last
+      search_session['prev'] = documents.first
```
cbeer: I'm a little surprised this works. Do you know if we are creating serialized blobs in the session store every time we do this?

jcoyne: Crap, I didn't realize this was the object in session. I'm going to rework this.

app/controllers/concerns/blacklight/search_context.rb
```diff
@@ -111,9 +111,7 @@ def setup_next_and_previous_documents
       response, documents = search_service.previous_and_next_documents_for_search index, search_state.reset(current_search_session.query_params).to_hash
 
       search_session['total'] = response.total
-      @search_context_response = response
-      @previous_document = documents.first
```
cbeer: I guess we should backport some deprecation for this?

jcoyne: I suppose we could make these deprecated instance variables in a backport.

cbeer: 👍

PR #1756, opened 2017-08-29
ShowPresenter#html_title should pass a no_html argument

jcoyne:
This is a hint to the rendering pipeline, that html should not be
used for joining.  This is not being use by blacklight itself, but
will be used by other projects.

See https://github.com/sul-dlss/dlme/issues/202

### DIFF
--- a/app/presenters/blacklight/show_presenter.rb
+++ b/app/presenters/blacklight/show_presenter.rb
@@ -35,9 +35,9 @@ module Blacklight
         fields = Array.wrap(view_config.html_title_field)
         f = fields.detect { |field| document.has? field }
         f ||= 'id'
-        field_values(field_config(f))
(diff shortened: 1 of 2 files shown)


app/presenters/blacklight/show_presenter.rb
```diff
@@ -35,9 +35,9 @@ def html_title
         fields = Array.wrap(view_config.html_title_field)
         f = fields.detect { |field| document.has? field }
         f ||= 'id'
-        field_values(field_config(f))
+        field_values(field_config(f), no_html: true)
       else
-        heading
+        heading(no_html: true)
```
cbeer: This bothers me a little bit because it isn't actually used in blacklight.. It almost makes me wonder if we're appropriately using `field_values` here at all. I'd like to put a little more thought into it before we ship this.

PR #1761, opened 2017-08-30
Remove unused modal form code

jcoyne:


tampakis:
The email and sms forms are submitted via modals and use this code.

PR #1768, opened 2017-09-01
Allow semantics to be defined from multiple stored field names

mejackreed:
This change is meant to allow for semantics to be combined from multiple
fields.

If accepted, I will backport to 6.x

### DIFF
--- a/app/models/concerns/blacklight/document/semantic_fields.rb
+++ b/app/models/concerns/blacklight/document/semantic_fields.rb
@@ -34,12 +34,14 @@ module Blacklight::Document
     # but extensions should call super and modify hash returned, to avoid
     # unintentionally erasing values provided by other extensions.
     def to_semantic_values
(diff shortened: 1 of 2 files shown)


app/models/concerns/blacklight/document/semantic_fields.rb
```diff
@@ -34,12 +34,14 @@ def field_semantics
...
-        value = self[field_name]
+      @semantic_value_hash ||= self.class.field_semantics.each_with_object(Hash.new([])) do |(key, field_names), hash|
+        ##
+        # Handles single string field_name or an array of field_names
+        value = Array.wrap(field_names).map { |field_name| self[field_name] }.flatten.compact
 
         # Make single and multi-values all arrays, so clients
         # don't have to know.
-        hash[key] = Array.wrap(value) unless value.nil?
+        hash[key] = Array.wrap(value) unless value.empty?
```
jcoyne: Do we still need to `Array.wrap(value)`?  Should we be not assigning this if value is empty?

PR #1770, opened 2017-09-01
Models should extend ApplicationRecord

jcoyne:


cbeer:
Is `ApplicationRecord` required now or just conventional? Do we need to provide a default implementation or something? Or make sure the blacklight install generator checks for it?

jcoyne:
@cbeer it's certainly the convention. Given the text here: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-models-now-inherit-from-applicationrecord-by-default it sounds much stronger than that:

> In Rails 4.2, an Active Record model inherits from ActiveRecord::Base. In Rails 5.0, all models inherit from ApplicationRecord.
> 
> ApplicationRecord is a new superclass for all app models, analogous to app controllers subclassing ApplicationController instead of ActionController::Base. This gives apps a single spot to configure app-wide model behavior.
> 
> When upgrading from Rails 4.2 to Rails 5.0, you need to create an application_record.rb file in app/models/

PR #1779, opened 2017-09-08
Pass the document object to the path helpers referenced in…

jkeck:
…session_tracking_path so that to_param is used instead of the id directly.

If this accepted I suppose I'll backport this as well as I need it in a 6.x release.

### DIFF
--- a/spec/helpers/blacklight/url_helper_behavior_spec.rb
+++ b/spec/helpers/blacklight/url_helper_behavior_spec.rb
@@ -241,7 +241,7 @@ RSpec.describe Blacklight::UrlHelperBehavior do
     end
 
     it "converts the counter parameter into a data- attribute" do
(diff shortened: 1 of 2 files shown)


spec/helpers/blacklight/url_helper_behavior_spec.rb
```diff
@@ -241,7 +241,7 @@
     end
 
     it "converts the counter parameter into a data- attribute" do
-      allow(helper).to receive(:track_test_path).with(hash_including(id: '123456', counter: 5)).and_return('tracking url')
+      allow(helper).to receive(:track_test_path).with(hash_including(id: an_instance_of(SolrDocument), counter: 5)).and_return('tracking url')
```
cbeer: I'm not sure it's important, but this could also be `have_attributes(id: '123456')` if we want to make sure we're getting the expected object?

jkeck: Cool, I hadn't seen `have_attributes` before.

PR #1796, opened 2017-10-09
Backport solr 7 schema changes to 6.x

### LINKED ISSUE #1795: Quickstart Guide no longer works because of solr 7 incompatibility issue
The latest solr from solr_wrapper is now 7.0. which does not support `defaultSearchField` and `solrQueryParser` which are defined in the default solr schema.

Following the quickstart guide leads to an exception when trying to start solr_wrapper.

Would it be preferable to pin Blacklight to solr 6.x or update schema to remove?

bibliotechy:
Remove deprecated defaultOperator and defaultSearchField solr configurations

Fixes #1795

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -16,10 +16,10 @@
  limitations under the License.
 -->
 
-<!--  
+<!--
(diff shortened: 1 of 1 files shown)


mejackreed:
👍

PR #1804, opened 2017-10-19
Hungarian translation for Blacklight.

JoshBone:
Hi,

As a result of our talks in Blacklight Summit the Hungarian translation is ready. 

Best,
Josh


mejackreed:
👏 bravo

PR #1806, opened 2017-11-14
added Chinese version

yuxisteem:
Hi, I've just created the Simplified Chinese version of the locale file. Please let me know if there are any changes needed. Thanks.

### DIFF
--- a/config/locales/blacklight.zh.yml
+++ b/config/locales/blacklight.zh.yml
@@ -1,0 +1,212 @@
+en:
+  views:
+    pagination:
+      first: '&laquo; 首页'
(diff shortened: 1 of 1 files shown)


config/locales/blacklight.zh.yml
```diff
@@ -0,0 +1,212 @@
+en:
```
cbeer: Thanks @yuxisteem -- this looks great, and just needs one small change before we merge. The `en` here on line 1 should be `zh` to match the locale.

yuxisteem: Sorry for this, I've got it changed.

PR #1807, opened 2017-11-15
Move the solr configuration into the generator templates

jcoyne:


mejackreed:
Curious about the reasoning here for this?

jcoyne:
This makes it clear that these templates are used by the generator and puts them in the expected location for templates.

PR #1820, opened 2018-01-05
Update query to work with Solr 7.2

jcoyne:
See the first item in:
https://lucene.apache.org/solr/7_2_0/changes/Changes.html#v7.2.0.upgrade_notes

This probably needs to be backported.

### DIFF
--- a/spec/features/search_filters_spec.rb
+++ b/spec/features/search_filters_spec.rb
@@ -64,7 +64,7 @@ RSpec.describe "Facets" do
       expect(page).to have_selector("span.selected", :text => "Tibetan")
       expect(page).to have_selector("span.facet-count.selected", :text => "1")
     end
-    within(".blacklight-pub_date") do 
+    within(".blacklight-pub_date") do
       expect(page).to have_selector("span.selected", :text => "2004")
       expect(page).to have_selector("span.facet-count.selected", :text => "1")
     end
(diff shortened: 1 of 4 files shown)


jcoyne:
It's still failing the did_you_mean_spec and search_spec which first started failing in https://travis-ci.org/projectblacklight/blacklight/builds/319183477

jcoyne:
We should probably merge #1821 until we figure out what else needs fixing here.

spec/models/blacklight/solr/search_builder_spec.rb
```diff
@@ -240,6 +240,14 @@
     describe "solr parameters for a field search from config (subject)" do
       let(:user_params) { subject_search_params }
 
+      before do
+        # pre solr-7.2-style configuration with local params
```
jcoyne: I'm not sure what this comment is telling me.  Is it "You can set..."?

spec/features/search_filters_spec.rb
```diff
@@ -187,16 +187,18 @@
     visit root_path
     within(".blacklight-subject_topic_facet") do
       expect(page).not_to have_selector(".card-block", :visible => true)
-      click_link "Topic"
+      find(".card-header").click
```
jcoyne: How is Solr 7.2 affecting the front end here?

cbeer: No idea, but I'm getting test failures on the branch without it. Testing in the browser works fine.

spec/models/blacklight/solr/search_builder_spec.rb
cbeer: The tests in this block expect the solr local params, which don't work at all in Solr 7.2, so we have to set that configuration in this unit test to test the feature.

PR #1827, opened 2018-01-18
Update to Bootstrap 4.0

cbeer:


jcoyne:
You can get around these errors by adding something like this: https://github.com/samvera/hyrax/pull/2511/files

cbeer:
Yep, I have that over in #1826.

PR #1829, opened 2018-01-26
[WIP] failing test for #label_for_search_field where no search fields are c…

mejackreed:
…onfigured

This is a failing test (and hopefully at some point a fix) for `label_for_search_field` when no search fields are configured. I'm not sure what the expected behavior should be.

### DIFF
--- a/app/helpers/blacklight/configuration_helper_behavior.rb
+++ b/app/helpers/blacklight/configuration_helper_behavior.rb
@@ -5,9 +5,9 @@ module Blacklight::ConfigurationHelperBehavior
 
   ##
   # Index fields to display for a type of document
-  # 
+  #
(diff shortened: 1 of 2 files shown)


skng5:
On further examination, this test won't work. We should test for the `Keyword` label. The problem is in `helper.label_for_search_field(nil))` which returns a hash instead of the default `Keyword` label.

jkeck:
@skng5 I think the intention was to have the test fail, and the TODO here is to write the code to make the test pass.  Will let @mejackreed confirm.

skng5:
@mejackreed @jkeck The translation path to the default label is doesn't correspond to anything in the translation file. I've changed it to what we want.

skng5:
Undetermined desired behavior.

PR #1835, opened 2018-02-14
add h1 to brand bar plus utility class to maintain spacing

starsplatter:


jcoyne:
Thanks for this @starsplatter.  This definitely fixes the bug of the missing `h1`.  However, we have some concern that the brand is not an appropriate header.  I wonder if we should have "Search Results" be `.sr-only` and make that an `h1`.   We'd like some time to do some research on this.

Looks like this was removed here: https://github.com/projectblacklight/blacklight/commit/3b8084cba27ba05c558d7a393662370354a46d78#diff-fa98ac5287321c38ba6a0864b3d15d29R31

Issue #1838, opened 2018-02-20
Solr 7.2+ compatibility: defType and LocalParams parsing


magibney:
As of Solr 7.2, parsing of LocalParams in solr are only supported when defType (the default parser) is "lucene" or "func" (see [Solr 7.2 upgrade notes](https://lucene.apache.org/solr/7_2_1/changes/Changes.html#v7.2.0.upgrade_notes) and [SOLR-11501](https://issues.apache.org/jira/browse/SOLR-11501)). 

This change will break Blacklight support for LocalParams for per-search-field rails-side configuration.

magibney:
Prefer to explicitly update existing blacklight configs to conform to new localParams parsing constraints. See #2020.

PR #1839, opened 2018-02-20
support localParams with solr-7.2 (see SOLR-11501)

magibney:
One way to extend support for LocalParams in blacklight with Solr 7.2+ ([issue 1838](https://github.com/projectblacklight/blacklight/issues/1838)) would be: at the point of addition of `solr_local_params`, to grab the query parser out of the top-level defType param (as set in defaults), replace it with "lucene" at the top level, and specify the query parser in the localParams block instead.


jcoyne:
Please add a test and fix the lint concerns raised by Rubocop.

magibney:
@jcoyne, does this look ok? The remaining rubocop complaint is about the overall module length, which I imagine is outside of the scope of this issue/PR. Sorry it took me a while to get around to fixing the rest of the lint concerns.

jcoyne:
@magibney Sorry for the delay. This looks good, but can you add a rubocop exemption to the `.rubocop_todo.yml` for the module lenght error?  That should get the test suite pass the rubocop checks.

magibney:
@jcoyne, sorry for the delay ... does this look workable? Thanks!

jcoyne:
Code looks good, but can you add a test please?

barmintor:
One way to get at what should be represented as a unit test (and what's missing): In this patch, there are 6 possible states of defType and local params, but only one tested in [blacklight/solr/search_builder_spec.rb](https://github.com/projectblacklight/blacklight/blob/master/spec/models/blacklight/solr/search_builder_spec.rb) under `describe "for :solr_local_parameters config"...`

What should be added there for maintainers to understand whether this behavior is correct in the future?

dkinzer:
@barmintor I couldn't add the spec directly to that branch so I created a new PR with the spec commit off of that branch.  I had to merge master into it and it is currently failing the lint check do to too many lines in some files (not sure what to do about that).

Also based I'm not sure the function is working as expected based on my tests there are some funky outcomes that seem counter intuitive to me.  I added some comments directly on the spec PR #2018.

The spec passes locally as written, though.

magibney:
Prefer to explicitly update existing blacklight configs to conform to new localParams parsing constraints. See #2020.

PR #1845, opened 2018-03-22
Do not assume spelling is available.

dkinzer:
Solr is not always the repository so it's not safe to assume that the
spelling field is available in the response.

This commit short circuits the respose when spelling is not available
otherwise an exception is thrown when trying to get words from nil.

### DIFF
--- a/app/helpers/blacklight/blacklight_helper_behavior.rb
+++ b/app/helpers/blacklight/blacklight_helper_behavior.rb
@@ -110,6 +110,7 @@ module Blacklight::BlacklightHelperBehavior
   # @param [Blacklight::Solr::Response] response
   # @return [Boolean]
   def should_show_spellcheck_suggestions? response
+    # The spelling response field may be missing from non solr repositories.
     response.total <= spell_check_max &&
       !response.spelling.nil? &&
       response.spelling.words.any?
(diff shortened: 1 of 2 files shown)


cbeer:
I'm surprised that `spelling` is nil-able; see https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/solr/response/spelling.rb#L8

Are you using a non-solr response adapter?

dkinzer:
@cbeer correct, we are developing a primo-central response adapter, and the primo API does not provide a spellcheck service like solr does.  That's why I ran into this issue.

app/helpers/blacklight/blacklight_helper_behavior.rb
```diff
@@ -110,7 +110,9 @@ def document_has_value? document, field_config
   # @param [Blacklight::Solr::Response] response
   # @return [Boolean]
   def should_show_spellcheck_suggestions? response
-    response.total <= spell_check_max && response.spelling.words.any?
+    response.total <= spell_check_max &&
+      !response.spelling.nil? &&
```
jcoyne: Please add a comment that states when this happens. (e.g. not using the Solr repository adapter)

dkinzer: will do.

PR #1847, opened 2018-03-28
Use new data-turbolinks attribute instead of the old…

### LINKED ISSUE #1846: Facets closing immediately with clicking heading-text in facet on demo site
This is only an issue when clicking on the facet heading text itself, not the chevron or the card header.

borked-facets

jkeck:
…data-no-turbolink attribute.

Fixes #1846 

https://github.com/turbolinks/turbolinks#disabling-turbolinks-on-specific-links


cbeer:
What versions of turbolinks does this support?

mejackreed:
I'm thinking its `~ 5.0`. Do we need to support something previous? (I'm trying to remember)

PR #1849, opened 2018-03-29
add a project solr_wrapper for developer happiness

mejackreed:
So a Blacklight dev can run solr separately from the rake tasks.


camillevilla:
@mejackreed developer happiness++

PR #1857, opened 2018-03-30
Remove the application name from an h1, and bump up page titles (and …

cbeer:
…other headers) as appropriate

### DIFF
--- a/app/views/catalog/_did_you_mean.html.erb
+++ b/app/views/catalog/_did_you_mean.html.erb
@@ -1,5 +1,5 @@
 <% if should_show_spellcheck_suggestions? @response %>
   <div id="spell">
-    <h4 class="suggest"><em><%= t('blacklight.did_you_mean', :options => safe_join(@response.spelling.words.map { |word| link_to_query(word) }, " #{t('blacklight.or')} ")).html_safe %></em></h4>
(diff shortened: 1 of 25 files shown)


app/views/catalog/_did_you_mean.html.erb
```diff
@@ -1,5 +1,6 @@
 <% if should_show_spellcheck_suggestions? @response %>
   <div id="spell">
-    <h4 class="suggest"><em><%= t('blacklight.did_you_mean', :options => safe_join(@response.spelling.words.map { |word| link_to_query(word) }, " #{t('blacklight.or')} ")).html_safe %></em></h4>
+    <!-- TODO -->
```
mejackreed: What is the todo here?

cbeer: Garbage.

mejackreed:
I'm not sure if this should be a follow on but the search results seem a little off?

screen shot 2018-03-30 at 4 15 15 pm

Bookmarks, History, and Home are fine.

mejackreed:
So my comment here is more about the fact that the result documents have a `h2` level, same as the "Search results" heading. I'm thinking that the results should be `h3`s. What do you think?

ggeisler:
FWIW, when working on DLME I think we did pay some attention to trying to get the heading levels correct and result items are `h3`s there:

-_digital_library_of_the_middle_east_-_dlme_search_results

PR #1862, opened 2018-04-09
Remove deprecated method: discard_flash_if_xhr

jgondron:
This discard flash method is beginning to cause problems with todays integration of rails and turbolinks. Ex: https://github.com/samvera/hyrax/issues/2592. Would like to remove this from 7.x. 

Related: #1863


jeremyf:
@jcoyne Can you restart this job https://travis-ci.org/projectblacklight/blacklight/jobs/364115047

jcoyne:
@jeremyf I restarted the build but it still failed. Perhaps a rebase?

jcoyne:
This also causes an issue with API mode, where there is no flash:
```
NameError (undefined local variable or method `flash' for #<CatalogController:0x00007fa45e014230>):
```

jcoyne:
Rebased into #1896

PR #1863, opened 2018-04-09
Deprecate method: discard_flash_if_xhr

jgondron:
Discarding flash on xhr requests will be deprecated in release 7. Added a deprecation message to the discard_flash_if_xhr method of blacklight/controller and provided instructions on how to either continue the behavior with 7.x or skip it now in 6.x.

### DIFF
--- a/app/controllers/concerns/blacklight/controller.rb
+++ b/app/controllers/concerns/blacklight/controller.rb
@@ -133,6 +133,10 @@ module Blacklight::Controller
     def discard_flash_if_xhr
       flash.discard if request.xhr?
     end
+    deprecation_deprecate discard_flash_if_xhr: "Discarding flash messages on XHR requests is deprecated.
+      If you wish to continue this behavior, add this method to your ApplicationController with an
+      after_action :discard_flash_if_xhr filter. To disable this behavior now and remove this warning, add
+      a skip_after_action :discard_flash_if_xhr to your ApplicationController."
 
     ##
     #


app/controllers/concerns/blacklight/controller.rb
```diff
@@ -133,6 +133,7 @@ def current_or_guest_user
     def discard_flash_if_xhr
       flash.discard if request.xhr?
     end
+    deprecation_deprecate :discard_flash_if_xhr
```
jcoyne: It seems like this is going to give people a deprecation message that they can't do anything about.   I'm not sure I understand why we'd do this.

jgondron: Yeah. Since it's completely going away, not being replaced by anything, I wasn't sure what to put. I'll add something. Feel free to help me with the wording on this.

app/controllers/concerns/blacklight/controller.rb
```diff
@@ -133,6 +133,9 @@ def current_or_guest_user
     def discard_flash_if_xhr
       flash.discard if request.xhr?
     end
+    deprecation_deprecate discard_flash_if_xhr: "Discarding flash messages on XHR requests is deprecated.
```
jcoyne: What if you don't want to continue with the behavior, but you just want the deprecation to go away? Should we add `skip_after_action :discard_flash_if_xhr`?

jgondron: Yep, you would add a skip. Added these instructions to the warning.

PR #1867, opened 2018-04-19
Backport jquery3 fixes to 6.10.x 

mejackreed:
Can we ship a 6.10.2 with the jquery3 compatibility?


jcoyne:
I'm seeing errors:
```
 ERROR: Error CREATEing SolrCore 'blacklight-core': Unable to create core [blacklight-core] Caused by: Setting defaultSearchField in schema not supported since Solr 7
```

mejackreed:
Yep. I'm going to spend my efforts to get an update to 6.14.x rather than fix this at the moment.

PR #1873, opened 2018-05-03
Low Contrast Errors from Bootstrap 4

mtribone:
Overrides the primary, success, info, and gray variables from Bootstrap. Updates the colors for page links and buttons for the primary, success and info colors in case they are used out-of-the-box by users. There are some false positives for the facet headers because they inherit from the heading–all good.

*Before shot with some false positives for the facets*

screen shot 2018-05-01 at 1 06 12 pm

screen shot 2018-05-03 at 11 55 44 am

*After shot with same false positives*

screen shot 2018-05-03 at 11 53 42 am

screen shot 2018-05-03 at 11 52 51 am

### DIFF
--- a/app/assets/stylesheets/blacklight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/blacklight/_bootstrap_overrides.scss
@@ -21,7 +21,31 @@
   }
 }
 
-
 .page-link {
   white-space: nowrap;
+}
+
+// Accessibility overrides for low contrast errors
+$primary: #006cdf;
+$success: #238f3c;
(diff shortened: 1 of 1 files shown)


cbeer:
I'm 👎 on this; I'd rather we let local implementations override bootstrap as appropriate instead of foisting it on them even if it's a good idea.

jcoyne:
I agree. This doesn't seem to be a Blacklight issue.  If you don't think the defaults are good, then your issue is with bootstrap itself.  It is not Blacklight's role to pick the colors.

PR #1876, opened 2018-05-11
Normalize the bookmark_controller.delete method.

dkinzer:
Currently the `bookmark_controller.create` method allows the user to
pass an array of bookmarks via the parameters with a format of
`[{ id: foo, document_type: bar }]`.

This is useful because theoretically we can bookmark records that are
not of type SolrDocument.  Likewise the bookmark model does not make any
assumptions about document_type other than providing a default if one is
not given.

Contrary to these methods, the `bookmark_controller.delete` is not
implemented in a consistent fashion to its counterpart.

`bookmark_controller.delete` works on only one bookmark at a time and
its document type can only be derived from `blacklight_config`. This is
unfortunate because `blacklight_config` is copied from the
`CatalogController` and the document type is already set to be
`SolrDocument`.

This commit updates `bookmark_controller.delete` to work in
a complimentary manner to `bookmark_controller.create`.

The use case of this change is to allow for being able to bookmark non
SolrDocuments in an environment where there are multiple backends not
just of type Solr.

## Additionally
* This change also fixes bookmark creation via bookmarks params not working in Rails when strong parameters is enabled.

### DIFF
--- a/app/controllers/concerns/blacklight/bookmarks.rb
+++ b/app/controllers/concerns/blacklight/bookmarks.rb
@@ -70,7 +70,7 @@ module Blacklight::Bookmarks
   # is simpler.
   def create
     @bookmarks = if params[:bookmarks]
-                   params[:bookmarks]
+                   permit_bookmarks[:bookmarks]
                  else
                    [{ document_id: params[:id], document_type: blacklight_config.document_model.to_s }]
                  end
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/blacklight/bookmarks.rb
```diff
@@ -102,9 +102,19 @@ def create
   # Beware, :id is the Solr document_id, not the actual Bookmark id.
   # idempotent, as DELETE is supposed to be.
   def destroy
-    bookmark = current_or_guest_user.bookmarks.find_by(document_id: params[:id], document_type: blacklight_config.document_model.to_s)
```
mejackreed: I'm not sure if I'm reading this correctly. Does this mean that an API call could now potentially destroy an un-related users Bookmark?

mejackreed: I see the change below now...

jcoyne:
Can we merge #1877 first?  That keeps the rubocop stuff out of this PR.

PR #1880, opened 2018-05-29
Fix blacklight:server task fails with missing file.

dkinzer:
`rake blacklight:server` fails with

```
Failed to execute solr create:
Specified configuration directory blacklight/solr/conf not found!

. Further information may be available in
/var/folders/l8/8ljsh0p103d9m4gg3866jlhh0000gp
```

I've updated the task to be the same as the ci version.  This fixed the
issue for me.

### DIFF
--- a/tasks/blacklight.rake
+++ b/tasks/blacklight.rake
@@ -52,8 +52,8 @@ namespace :blacklight do
       Rake::Task['engine_cart:generate'].invoke
     end
 
-    SolrWrapper.wrap(port: '8983') do |solr|
-      solr.with_collection(name: 'blacklight-core', dir: File.join(File.expand_path("..", File.dirname(__FILE__)), "solr", "conf")) do
+    SolrWrapper.wrap(version: '7.1.0', port: '8983') do |solr|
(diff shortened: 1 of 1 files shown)


tasks/blacklight.rake
```diff
@@ -52,8 +52,8 @@ namespace :blacklight do
       Rake::Task['engine_cart:generate'].invoke
     end
 
-    SolrWrapper.wrap(port: '8983') do |solr|
-      solr.with_collection(name: 'blacklight-core', dir: File.join(File.expand_path("..", File.dirname(__FILE__)), "solr", "conf")) do
+    SolrWrapper.wrap(version: '7.1.0', port: '8983') do |solr|
```
jcoyne: Is this parameter change required?

dkinzer: Not sure, probably not.  I included to be consistent with the CI setup.  See line 16 above.

jcoyne:
@dkinzer thanks for submitting this. I've picked up this idea in https://github.com/projectblacklight/blacklight/pull/1881 and moved the dir to a const and moved some of the configs to the solr_wrapper.yml

Issue #1818, opened 2017-12-20
Search box on homepage not functional in Safari


mejackreed:
This was originally reported by @drh-stanford in https://github.com/sul-dlss/earthworks/issues/317 . We are seeing the same problem upstream in Blacklight `master`

PR #1884, opened 2018-05-30
[WIP] Add autofocus.js to the javascript package

jcoyne:
It seems that it was overlooked in https://github.com/projectblacklight/blacklight/commit/6a61628e7c3aaaa756d48d0cb1a6b47701765470

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "[![Build Status](https://travis-ci.org/projectblacklight/blacklight.png?branch=master)](https://travis-ci.org/projectblacklight/blacklight) [![Gem Version](https://badge.fury.io/rb/blacklight.png)](http://badge.fury.io/rb/blacklight) [![Coverage Status](https://coveralls.io/repos/github/projectblacklight/blacklight/badge.svg?branch=master)](https://coveralls.io/github/projectblacklight/blacklight?branch=master)",
   "main": "app/assets/javascripts/blacklight",
   "scripts": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
...
   "main": "app/assets/javascripts/blacklight",
   "scripts": {
-    "js-compile-bundle": "shx cat app/javascript/blacklight/core.js app/javascript/blacklight/autocomplete.js app/javascript/blacklight/bookmark_toggle.js app/javascript/blacklight/checkbox_submit.js app/javascript/blacklight/collapsable.js app/javascript/blacklight/facet_load.js app/javascript/blacklight/modal.js app/javascript/blacklight/search_context.js | shx sed \"s/^(import|export).*//\" | babel --filename app/javascript/blacklight/blacklight.js > app/assets/javascripts/blacklight/blacklight.js"
+    "js-compile-bundle": "shx cat app/javascript/blacklight/core.js app/javascript/blacklight/autocomplete.js app/javascript/blacklight/autofocus.js app/javascript/blacklight/bookmark_toggle.js app/javascript/blacklight/checkbox_submit.js app/javascript/blacklight/collapsable.js app/javascript/blacklight/facet_load.js app/javascript/blacklight/modal.js app/javascript/blacklight/search_context.js | shx sed \"s/^(import|export).*//\" | babel --filename app/javascript/blacklight/blacklight.js > app/assets/javascripts/blacklight/blacklight.js"
```
dkinzer: You should probably add autofocus file last or at least test this order in Safari to make sure this bug this not return ... #1818

jcoyne:
I don't think we need this polyfill anymore. Autofocus is supported on all browsers except iOS safari (where autofocus would be a usability concern). https://caniuse.com/#feat=autofocus

PR #1887, opened 2018-06-04
Support Rails API mode

jcoyne:
Don't call methods that aren't available.


jcoyne:
@jkeck I added a few more commits after your approval.  Still looking good?

PR #1897, opened 2018-06-06
Use the latest version of solr

jcoyne:


cbeer:
Now that solr_wrapper has some logic to choose the latest version, maybe we can omit the version entirely?

PR #1910, opened 2018-06-26
Remove hardcoding of SolrDocument in url_for_document

jcoyne:
I want to open Blacklight to the possibility of being backed by a search engine other than Solr (e.g. ElasticSearch), so we don’t want to encode `SolrDocument` if the document comes from something other than Solr.


jcoyne:
@cbeer have you pondered?

barmintor:
Rebased on main, and opened a PR to try to address @cbeer's change request.

PR #1911, opened 2018-06-26
Avoid unnecessary instantiation of data object

jcoyne:
This makes it possible to use with Elasticsearch due to:
https://github.com/elastic/elasticsearch-rails/issues/804

### DIFF
--- a/app/controllers/concerns/blacklight/catalog.rb
+++ b/app/controllers/concerns/blacklight/catalog.rb
@@ -65,7 +65,7 @@ module Blacklight::Catalog
       path = uri.query ? "#{uri.path}?#{uri.query}" : uri.path
       redirect_to path, status: 303
     else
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -65,7 +65,7 @@ def track
       path = uri.query ? "#{uri.path}?#{uri.query}" : uri.path
       redirect_to path, status: 303
     else
-      redirect_to blacklight_config.document_model.new(id: params[:id]), status: 303
+      redirect_to({ action: :show, id: params[:id] }, status: 303)
```
mjgiarlo: It's not clear to me that the code you removed is not necessary. Can you explain?

jcoyne: It's not necessary, because the tests show the intended behavior is unchanged.

PR #1912, opened 2018-06-26
Pin selenium-webdriver to avoid DevToolsActivePort file doesn't exist…

jcoyne:
…' error

https://github.com/SeleniumHQ/selenium/issues/6049
https://bugs.chromium.org/p/chromedriver/issues/detail?id=2473


mjgiarlo:
This change looks fine to me, though pinning to patch-level releases is suboptimal. We can revisit that when the underlying issue allows the dependency to be more loosely pinned.

PR #1913, opened 2018-06-27
Move suggest handling into the repository

jcoyne:
This allows the repository to abstract how suggestion handling happens

### DIFF
--- a/app/models/blacklight/suggest_search.rb
+++ b/app/models/blacklight/suggest_search.rb
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
+
 module Blacklight
   class SuggestSearch
     attr_reader :request_params, :repository
 
     ##
     # @param [Hash] params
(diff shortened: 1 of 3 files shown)


app/models/blacklight/suggest_search.rb
```diff
@@ -15,20 +17,7 @@ def initialize(params, repository)
...
-    # @return [RSolr::HashWithResponse]
-    def suggest_results
-      repository.connection.send_and_receive(suggest_handler_path, params: request_params)
-    end
-
-    ##
-    # @return [String]
-    def suggest_handler_path
-      repository.blacklight_config.autocomplete_path
+      respository.suggest_results(request_params)
```
cjcolvar: typo: `respository`

jcoyne: Thanks Chris. Fixed.

PR #1914, opened 2018-06-27
Extract reflect_fields method to the repository

jcoyne:
This allows the repository to implement reflect_fields and allows us to
remove a conditional.

### DIFF
--- a/lib/blacklight/configuration/fields.rb
+++ b/lib/blacklight/configuration/fields.rb
@@ -5,6 +5,8 @@ module Blacklight
     # solr fields configuration
     module Fields
       extend ActiveSupport::Concern
+      extend Deprecation
+      self.deprecation_horizon = 'blacklight version 8.0.0'
 
       module ClassMethods
         # Add a configuration block for a collection of solr fields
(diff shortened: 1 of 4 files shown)


lib/blacklight/configuration/fields.rb
```diff
@@ -137,17 +120,35 @@ def add_blacklight_field config_key, *args, &block
...
           return nil
         end
 
         @table[:luke_fields] ||= Rails.cache.fetch("blacklight_configuration/admin/luke", expires_in: 1.hour) do
           begin
-            if repository_class <= Blacklight::Solr::Repository
-              repository = repository_class.new(self)
-              repository.send_and_receive('admin/luke', params: { fl: '*', 'json.nl' => 'map' })['fields']
-            end
+            repository = repository_class.new(self)
```
mjgiarlo: Can you safely remove the condition around this?

jcoyne: The tests seem to pass.  I can think of no reason why we can't require repository implementations to implement this interface.

mjgiarlo: Looks like @cbeer last touched this code, here: https://github.com/projectblacklight/blacklight/commit/d75ff8594463f3c078ea0b0e51fc48ed15bef70e

Thoughts on that, Chris?

lib/blacklight/configuration/fields.rb
```diff
@@ -137,17 +120,35 @@ def add_blacklight_field config_key, *args, &block
...
+          config.key = field
+          if self[config_key.pluralize][config.key]
+            self[config_key.pluralize][config.key] = config.merge(self[config_key.pluralize][config.key])
+          else
+            add_blacklight_field(config_key, config, &block)
+          end
+        end
+      end
+
       def luke_fields
```
cbeer: Should we rename this method?

lib/blacklight/configuration/fields.rb
cbeer: I don't think it's a problem; I don't remember what the luke response looks like, but if we think something similar will work cross-repository, I don't see a problem with it.

PR #1930, opened 2018-07-13
Add a routing spec for SearchHistoryController

jcoyne:


barmintor:
This is a weird set of failures and successes - why wouldn't the routes be loaded in Rails 5.1.x?

barmintor (approved):
Some test is better than no test!

PR #1941, opened 2018-07-19
Add labels to facets and fields in json response

jcoyne:


app/views/catalog/index.json.jbuilder
```diff
@@ -13,19 +12,26 @@ end
...
     json.id document.id
     json.attributes do
       doc_presenter = index_presenter(document)
 
       index_fields(document).each do |field_name, field|
-        if should_render_index_field? document, field
-          json.set! field_name, doc_presenter.field_value(field_name)
+        next unless should_render_index_field? document, field
+        json.set!(field_name) do
+          json.id "#{document_url}##{field_name}"
```
jkeck: Just curious if there would be any use to having the field name as an attribute instead of (or in addition to) being interpolated with the document URL.  It'll be easy enough to strip it if it was needed, but just thinking a bit about migration path for those using this API and were using the field name in some way.

jcoyne: The whole move to JSON-API was part of the 7.0 release, so changing what we're returning there, prior to a release is not a problem I want to worry about too much.

PR #1947, opened 2018-07-20
size() is deprecated in jQuery 1.8, use length instead

cdmo:
I'm probably not following the right etiquette here, but size() was deprecated and in newer versions of jquery this breaks rendering


mejackreed:
Does this need to be rebuilt for the npm package?

jcoyne:
Yes. I'll do that.

cdmo:
That was fast! Thanks.

PR #1964, opened 2018-08-19
Add a test for url_for_document, especially for a case that uses rail…

cbeer:
…s polymorphic routing magic.

Fixes #1910

### DIFF
--- a/lib/blacklight/search_state.rb
+++ b/lib/blacklight/search_state.rb
@@ -46,9 +46,7 @@ module Blacklight
     # to provide more interesting routing to
     # documents
     def url_for_document(doc, options = {})
-      if respond_to?(:blacklight_config) &&
-          blacklight_config.show.route &&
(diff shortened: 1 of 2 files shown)


lib/blacklight/search_state.rb
```diff
@@ -46,9 +46,7 @@ def reset(params = nil)
     # to provide more interesting routing to
     # documents
     def url_for_document(doc, options = {})
-      if respond_to?(:blacklight_config) &&
-          blacklight_config.show.route &&
-          (!doc.respond_to?(:to_model) || doc.to_model.is_a?(SolrDocument))
+      if blacklight_config.show.route || (doc.respond_to?(:to_model) && doc == doc.to_model)
```
jcoyne: I think this `||` must still be an `&&`. Because the tests fail due to `blacklight_config.show.route` being nil

barmintor:
On closer inspection, these tests are all present in `main` - @cbeer if I'm wrong there please let me know, but I think this is closeable.

PR #1966, opened 2018-08-28
Webpacker

cdmo:
Putting this out there as a starting point really for getting webpacker to work inside Blacklight. I'm pairing this with an update I've added to https://github.com/projectblacklight/blacklight/wiki/Using-Webpacker-to-compile-javascript-assets. 

More would have to be done in terms of clearing out the old asset pipeline, like removing the Action Cable stuff and it's derivatives and a lot of other pruning of gems that help in the asset pipeline, but these changes will allow Blacklight 7 to run with webpacker compiling, generating zero errors (typeahead working, modals rendering properly, expand/collapse stuff all good, etc). 

The changes to `base.html.erb` are a little over-zealous probably, all that really matters is changing out the ` <%= javascript_include_tag "application" %>`. I'm not exactly sure why my `base.html.erb` is so different, maybe because we (Penn State) are also upgrading to rails 5.2 for our project, so perhaps this is a result of that.

### DIFF
--- a/app/views/catalog/_home_text.html.erb
+++ b/app/views/catalog/_home_text.html.erb
@@ -34,11 +34,3 @@
   <h2 class='card-header collapsed collapse-toggle' data-toggle="collapse" data-target="#about-content"><a href="/rails/info/properties">About your application&rsquo;s environment</a></h2>
   <div id="about-content" class="card-body collapse"></div>
 </div>
-
-<script>
-  Blacklight.onLoad(function() {
-    $('#about .card-header').one('click', function() {
-      $($(this).data('target')).load($(this).find('a').attr('href'));
-    });
-  });
-</script>
--- a/app/views/layouts/blacklight/base.html.erb
+++ b/app/views/layouts/blacklight/base.html.erb
@@ -1,35 +1,43 @@
 <!DOCTYPE html>
 <html lang="en" class="no-js">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
 
-    <!-- Internet Explorer use the highest version available -->
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <!-- Mobile viewport optimization h5bp.com/ad -->
+  <meta name="HandheldFriendly" content="True">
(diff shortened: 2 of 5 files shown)


app/views/catalog/_home_text.html.erb
```diff
@@ -33,12 +33,4 @@
 <div id="about" class="card">
   <h2 class='card-header collapsed collapse-toggle' data-toggle="collapse" data-target="#about-content"><a href="/rails/info/properties">About your application&rsquo;s environment</a></h2>
   <div id="about-content" class="card-body collapse"></div>
-</div>
-
-<script>
-  Blacklight.onLoad(function() {
```
jcoyne: I'm curious about why this needed to be removed.

app/views/layouts/blacklight/base.html.erb
```diff
@@ -1,35 +1,43 @@
...
-
-    <div class="row">
-      <%= content_for?(:content) ? yield(:content) : yield %>
-    </div>
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+
+  <!-- Mobile viewport optimization h5bp.com/ad -->
+  <meta name="HandheldFriendly" content="True">
```
jcoyne: This appears to revert some commits like https://github.com/projectblacklight/blacklight/commit/42a0fca01bb3a9511a1d92c748dfceeaab31cdcc#diff-01a8fe6697b36c0d2663ca12f0fe9502

app/views/catalog/_home_text.html.erb
cdmo: jQuery wasn't loaded yet: https://git.psu.edu/i-tech/psulib_blacklight/issues/75. This was only a problem on the homepage.

cdmo:
@jrochkind brought up a good point on Slack when I mentioned this PR there. Right now I am treating the `blacklight-frontend` package as just a link to the main blacklight GH repo. This is because the latest tag on the `blacklight-frontend` in yarn packages still has deprecation issues that break execution of js. Maybe a new tag could be made for that package? Alternatively, another thought, remove the package altogether and just pull in the Blacklight javascript from the blacklight gem?

Also, I should point out, that there is still one outstanding bug that I've noticed. Un-bookmarking hits an error. But it appears to be a the Ruby-level, no JS.

jcoyne:
I'm not sure the BL community is ready for webpacker to be a requirement in version 7.  My intentions are to make it an optional (shall we say "experimental"?) feature for this release.

cdmo:
Sounds good to me.

barmintor:
We want to come back to this when Rails forces our hand, and divide this up into PRs for the different changes here.

PR #1970, opened 2018-09-13
Add aria attributes to pagination for improved accessibility

adamjarling:
Adds `aria-` attributes to pagination items.  Work coming from this Hyrax issue:  https://github.com/samvera/hyrax/issues/3138

@no-reply suggested we make PRs to blacklight, so any and all feedback welcome.  Or if there's a better way to re-factor any of the proposed changes in this PR.   The blacklight code is great by the way, super clean and organized having a first look around.

pagination-aria

### DIFF
--- a/app/views/catalog/_results_pagination.html.erb
+++ b/app/views/catalog/_results_pagination.html.erb
@@ -1,9 +1,9 @@
 <% if show_pagination? and @response.total_pages > 1 %>
  <div class="row record-padding">
   <div class="col-md-12">
-    <div class="pagination">
+    <nav class="pagination" role="region" aria-label="<%= t('views.pagination.aria.container_label') %>">
       <%= paginate @response, :outer_window => 2, :theme => 'blacklight' %>
-    </div>
+    </nav>
   </div>
  </div>
 <% end %>
--- a/app/views/kaminari/blacklight/_first_page.html.erb
+++ b/app/views/kaminari/blacklight/_first_page.html.erb
@@ -7,5 +7,5 @@
     remote:        data-remote
 -%>
 <li class="page-item <%= 'disabled' if current_page.first? %>">
-  <%= link_to_unless current_page.first?, raw(t 'views.pagination.first'), url, :remote => remote, class: 'page-link' %>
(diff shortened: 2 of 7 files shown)


barmintor:
I think this is failing because we're missing translations for the new aria keys in i18n for `[es, fr, de, it, pt-BR]`. Do we have a source for those translations?

app/views/catalog/_results_pagination.html.erb
```diff
@@ -1,9 +1,9 @@
 <% if show_pagination? and @response.total_pages > 1 %>
  <div class="row record-padding">
   <div class="col-md-12">
-    <div class="pagination">
+    <nav class="pagination" role="region" aria-label="<%= t('views.pagination.aria.container_label') %>">
```
mejackreed: I've seen a few web resources that specify using `role="navigation"` for major groups like this. Is there somewhere I can learn more about  the rationale for `role="region"`?

I've been looking at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role and http://www.a11ymatters.com/pattern/pagination/

mtribone: @mejackreed I believe the feedback from the consultants stated that the pagination is "secondary" navigation and not "main" navigation, so adding the role of region was to fill in for IE–me thinks-and the spotty support of `<nav>` with assistive tech with that browser. 

I'll go back and dig up the report, but I believe the idea is that if every grouping of links has `role=navigation` then nothing stands out as more important than the other. You know? What's the hierarchy of navigation on the page? menu bar > facets > pagination?

barmintor:
Bumping the translation issue here- tests need to pass. It might be easier to solicit them from a branch in this repository?

app/views/catalog/_results_pagination.html.erb
barmintor: Is this a relevant ref: https://www.w3.org/WAI/GL/wiki/Using_the_region_role_to_identify_a_region_of_the_page

mejackreed: Sounds good. Thanks for the context and understanding @mtribone

barmintor:
Commits merged in #1991, thanks @adamjarling !

Issue #1898, opened 2018-06-07
API response of attributes should have the human readable label.


jcoyne:
Currently a response attributes looks like:

```json
         "attributes" : {
            "id" : "2008308175",
            "isbn_ssim" : [
               "8186470336"
            ],
            "pub_date_ssim" : [
               "2002"
            ],
            "lc_1letter_ssim" : [
               "B - Philosophy, Psychology, Religion"
            ],
            "author_tsim" : [
               "Ṅag-dbaṅ-blo-bzaṅ-rgya-mtsho, Dalai Lama V, 1617-1682"
            ],
            "score" : 1,
            "subject_addl_ssim" : [
               "Prayers and devotions",
               "India Prayers and devotions"
            ],
         }
```

I propose we change this to:

```json
         "attributes": {
            "id" : {
                "label": "id",
                "value": "2008308175"
            },
            "isbn_ssim": {
               "label": "ISBN",
               "value": [
               "8186470336"
               ]
            },
            "pub_date_ssim" : {
               "label": "Publication Date", 
               "value": [
                   "2002"
                ],
            }
         }
```

jcoyne:
Fixed by #1979

PR #1979, opened 2018-10-04
Add labels to fields in the json show response

jcoyne:
This gives the show view partity with the index view.
See https://github.com/projectblacklight/blacklight/pull/1941


jcoyne:
Fixes #1898

Issue #1974, opened 2018-10-02
Error on `bundle exec rake blacklight:server`


dkinzer:
Running this command on latest master commit I get the following error locally:

```
rake aborted!p.md5: |=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=--| 0% ( ETA: ??:??:?? )
SolrWrapper::SolrWrapperError: Unable to download solr from http://www.us.apache.org/dist/lucene/solr/7.5.0/solr-7.5.0.zip.md5
404 Not Found: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL /dist/lucene/solr/7.5.0/solr-7.5.0.zip.md5 was not found on this server.</p>
</body></html>
tasks/blacklight.rake:57:in `block (2 levels) in <top (required)>'
/Users/dkinzer/.rbenv/versions/2.4.4/bin/bundle:23:in `load'
/Users/dkinzer/.rbenv/versions/2.4.4/bin/bundle:23:in `<main>'

Caused by:
OpenURI::HTTPError: 404 Not Found
```

jcoyne:
It looks like you might have an out of date version of solr_wrapper.  Can you run `gem install solr_wrapper` and then try again?

dkinzer:
hmm that didn't work but running `bundle update solr_wrapper` got it a bit farther field this time.  The new error is:

```
Failed to execute solr create: 
Specified configuration directory /Users/dkinzer/projects/blacklight/solr/conf not found!

. Further information may be available in /var/folders/l8/8ljsh0p103d9m4gg3866jlhh0000gp/T/solr-7.5.0/logs
/Users/dkinzer/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/solr_wrapper-2.0.0/lib/solr_wrapper/instance.rb:321:in `exec'
/Users/dkinzer/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/solr_wrapper-2.0.0/lib/solr_wrapper/instance.rb:158:in `create'
/Users/dkinzer/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/solr_wrapper-2.0.0/lib/solr_wrapper/instanc
...
```

[...]

PR #1981, opened 2018-10-05
Add sym link to .internal/solr dir.

dkinzer:
REF projectblacklight/blacklight#1974

Fixes issue with ./solr/conf dir not found per instructions set in
`.internal_test_app/.solr_wrapper.yml`

### DIFF
--- a/solr
+++ b/solr
@@ -1,0 +1,1 @@
+.internal_test_app/solr


jcoyne:
I don't want to accept this here, because so far you are the only person who has been able to produce this problem. This might be a workaround that you keep locally in your own workspace.

PR #1982, opened 2018-10-05
Avoid unnecessary lookup of field config

jcoyne:
This also ensures the fields that are presented are the appropriate ones for the specific object type. Previously when it was looking up the field by name, it only searched in the top level config, not the per-object type config.


cbeer:
This will also need a deprecation warning in the 6.x branch.

jcoyne:
@cbeer that's a good idea. Another thought is we may re-add passing a string and deprecate that behavior in this PR. Which is preferable?

PR #1985, opened 2018-10-16
Removed commented options that are no longer available in api

mejackreed:


jcoyne:
can you rebuild the distribution javascript too, then your changes will be in sync. `npm run js-compile-bundle` should do it.

mejackreed:
Ah thanks for the reminder on that. Rebuilt and pushed.

PR #1989, opened 2018-10-23
Do not add facet limit unless it is known.

dkinzer:
REF: #1990 

This commit updates the code that adds facet limits from user supplied
parameters so that facets are only added if they have been configured
as facet fields.  This is done to allow for multiple faceted searches to
be defined in the same URL while still keeping the facets separate.

### DIFF
--- a/spec/features/search_spec.rb
+++ b/spec/features/search_spec.rb
@@ -115,9 +115,4 @@ RSpec.describe "Search Page" do
     expect(page).to have_content "Welcome!"
     expect(page).not_to have_selector "#q[value='history']"
   end
-
-  it "sanitizes searches with unconfigured facet parameters" do
-    visit root_path f: { missing_s: [1] }
-    expect(page).not_to have_text "Missing S"
-  end
 end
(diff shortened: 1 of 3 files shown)


cdmo:
@dkinzer this is just sanity checking then, right? If the facet isn't defined, i.e. a bad config/typo in facet setting, then don't add it to the query?

dkinzer:
@cdmo effectively yes.  If we are passing some facets over URL params the app shouldn't just pass these to solr backend if they are not actually defined / configured at the controller level.

barmintor:
I think these changes are effectively in-place via other refactors, but I have rebased against `main` to check. The specs might still be useful to verify!

spec/features/search_spec.rb
```diff
@@ -115,4 +115,9 @@
     expect(page).to have_content "Welcome!"
     expect(page).not_to have_selector "#q[value='history']"
   end
+
+  it "sanitizes searches with unconfigured facet parameters" do
+    visit root_path f: { missing_s: [1] }
+    expect(page).not_to have_text "Missing S"
```
barmintor, line 121: I think this is a useful thing to check, but I'm wary of "page doesn't have text" tests - high likelihood of false passes. I wonder if we could now re-implement this as a ConstraintsComponent test?

dkinzer: Sounds good.

barmintor (approved):
Adding specs to verify facet params are sanitized before passing to solr

barmintor:
Waiting on CI to merge, thanks for this work @dkinzer

PR #2005, opened 2018-10-31
Remove unnecessary route

jcoyne:
This was already defined by `lib/blacklight/routes/searchable.rb`

### DIFF
--- a/app/helpers/blacklight/url_helper_behavior.rb
+++ b/app/helpers/blacklight/url_helper_behavior.rb
@@ -72,7 +72,7 @@ module Blacklight::UrlHelperBehavior
     if respond_to?(controller_tracking_method)
       send(controller_tracking_method, params.merge(id: document))
     else
-      blacklight.track_search_context_path(params.merge(id: document))
+      main_app.track_catalog_path(params.merge(id: document))
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,5 +2,4 @@
 Blacklight::Engine.routes.draw do
   get "search_history",             to: "search_history#index",   as: "search_history"
   delete "search_history/clear",       to: "search_history#clear",   as: "clear_search_history"
-  post "/catalog/:id/track", to: 'catalog#track', as: 'track_search_context'
 end


barmintor:
@jcoyne I think this change looks good, but it's raising the question of where the Searchable routes are tested? I might be missing them, so I'll spend a minute more looking.

barmintor:
Some testing in blacklight/spec/controllers/catalog_controller_spec.rb

jcoyne:
@barmintor I suspect so, but I also suspect it was broken before this too as that route was explicitly to `/catalog`

barmintor:
Is it possible that this route is not actually redundant/unnecessary? When the `:searchable` concern is effective, you don't fall into that block (your helper should respond to `"track_#{controller_name}_path"`). If you don't (like the BookmarksController), you still need a path and you don't have that routing concern to provide it.

barmintor:
@jcoyne that is, it might be wrong or deceptive (especially when you don't name your BL controller Catalog, so it's just a free standing broken route).

app/helpers/blacklight/url_helper_behavior.rb
```diff
@@ -72,7 +72,7 @@ def session_tracking_path document, params = {}
     if respond_to?(controller_tracking_method)
       send(controller_tracking_method, params.merge(id: document))
```
barmintor: Should this message be sent to `main_app` as well? cf `respond_to?` guard.

app/helpers/blacklight/url_helper_behavior.rb
```diff
@@ -72,7 +72,7 @@ def session_tracking_path document, params = {}
...
     else
-      blacklight.track_search_context_path(params.merge(id: document))
+      main_app.track_catalog_path(params.merge(id: document))
```
barmintor: Should there be a log message here warning that falling back to an unsafe default?

dkinzer: Yeah, it would be nice if these track path worked with more than just catalog... we usually have to add these after the fact in order to make our build not fail (i.e. track for articles vs catalog etc.)

bibliotechy: I think that logging a message about the controller_tracking_method not being defined and falling back on the unsafe default, perhaps with a pointer to some docs, would be a good addition here.

barmintor:
Closed via #2031

Issue #1868, opened 2018-04-23
Vanilla install of `7.0.0.rc1` issues


cdmo:
I'm using a Mac 10.12.6.

Just wanted to share my experience in attempting a vanilla install of the rc1 release for Blacklight 7. I ran into a few problems, most I could fix:

1. The `add_devise_to_users.rb` migration failed
1. Hit a routing error: `undefined method `add_show_tools_partial' for CatalogController:Class`
1. Hit scss errors on unknown variables
1. `unknown field 'marc_display` on `rake solr:marc:index_test_data` 

I solved 1 by just removing the migration (sidestepping this issue entirely)

I solved 2 by changing `CatalogController`:

```
# frozen_string_literal: true
class CatalogController < ApplicationController

  include Blacklight::Catalog
  include Blacklight::DefaultComponentConfiguration # this is what I added
  include Blacklight::Marc::Catalog
```

Solved the third issue by just commenting out the unknown variables in `blacklight.scss`

The 4th one I haven't been able to solve. Seems like it should be working, but isn't. Running `rake solr:marc:index_test_data` the following error is hit for all documents:

```
2018-04-23T15:08:25-04:00 ERROR Could not add record 2008308175/2008308175 at source file position 25: Solr error response: 400: {
  "responseHeader":{
    "status":400,
    "QTime":0},
  "error":{
    "metadata":[
      "error-class","org.apache.solr.common.SolrException",
      "root-error-class","org.apache.solr.common.SolrException"],
    "msg":"ERROR: [doc=2008308175] unknown field 'marc_display'",
    "code":400}}
```

EDIT: It looks like the field doesn't exist in the out-of-the-box `schema.xml` file doesn't have any mention of that `marc_display` field. When I use the `schema.xml` from the old 6.x branch, it will index but then Blacklight hits an error `Blacklight::Exceptions::InvalidRequest in CatalogController#index` ...

erosas87:
Do you need to install marc? I don't know if it's supported with v7.0.0.rc1, I would recommend trying to install it without that

cdmo:
@erosas87 Are you referring to the `solr-marc` gem? So, to explain, I was following the quick start guide, which has solr-marc as a dependency, but perhaps the intention is to update the quick start guide later or maybe abandon it.

cdmo:
Sorry I didn't see the release notes about 

`  include Blacklight::DefaultComponentConfiguration # this is what I added`

My bad!

[...]

PR #2030, opened 2018-11-12
bump blacklight_marc

cdmo:
bumping the version of blacklight_marc

connected to #1868 

cc @barmintor

### DIFF
--- a/lib/generators/blacklight/install_generator.rb
+++ b/lib/generators/blacklight/install_generator.rb
@@ -88,7 +88,7 @@ EOF
       return unless options[:marc]
 
       blacklight_marc = String.new('blacklight-marc')
-      gem blacklight_marc, '~> 6.1'
+      gem blacklight_marc, '= 7.0.0.rc1'
 
       Bundler.with_clean_env do
         run "bundle install"


lib/generators/blacklight/install_generator.rb
```diff
@@ -88,7 +88,7 @@ def generate_blacklight_marc_demo
       return unless options[:marc]
 
       blacklight_marc = String.new('blacklight-marc')
-      gem blacklight_marc, '~> 6.1'
+      gem blacklight_marc, '= 7.0.0.rc1'
```
jcoyne: should this be `>= 7.0.0.a`?

cdmo: Like "pre-release or greater of this dot release," right? Looking around on the internet, I can't seem to find documentation on how to specify this, the closest thing I could find was https://github.com/bundler/bundler/issues/1198#issuecomment-1228741.  These seem like they're all possible, good options?

`= 7.0.0.rc1` (I know this one works)
`>= 7.0.0.a`
`>= 7.0.0.beta`
`~> 7.0.0.a`
`~> 7.0.0.beta`

jcoyne: Okay, how about ` < 8`?  I'm thinking we don't want to have to revisit this line once we release newer versions of blacklight-marc.

PR #2039, opened 2018-12-13
Test with Ruby 2.6

cbeer:


jcoyne:
It appears that Thor needs to be updated to work with bundler 2.0 (which ships with Ruby 2.6)

jcoyne:
@cbeer can you update this to use the final release?

PR #2040, opened 2018-12-14
Replace coveralls with codeclimate

cbeer:


jcoyne:
I don't know. Codeclimate says our code is no good. 😆 https://codeclimate.com/github/projectblacklight/blacklight

PR #2061, opened 2019-01-24
Move `User#to_s` from generator to `Blacklight::User`

no-reply:
This method had been injected directly into the application's `User`, but inheritance seems like the stronger approach.

### DIFF
--- a/app/models/concerns/blacklight/user.rb
+++ b/app/models/concerns/blacklight/user.rb
@@ -5,6 +5,8 @@ module Blacklight::User
   # SEE ALSO:  The lib/blacklight/generator/user_generator.rb class for where this
   # is generated into the hosting application.
   included do
+    class_attribute :string_display_key
+
     has_many :bookmarks, dependent: :destroy, as: :user
     has_many :searches,  dependent: :destroy, as: :user
   end
@@ -34,17 +36,5 @@ module Blacklight::User
     return send(string_display_key) if respond_to?(string_display_key)
 
     super
-  end
-
-  module ClassMethods
-    def string_display_key(attr_name = nil)
-      return @string_display_key if attr_name.blank?
-
-      @string_display_key = attr_name
-    end
-
-    def string_display_key_unless(attr_name)
-      @string_display_key || string_display_key(attr_name)
-    end
   end
 end
--- a/app/models/concerns/blacklight/user.rb
+++ b/app/models/concerns/blacklight/user.rb
@@ -26,4 +26,10 @@ module Blacklight::User
   def existing_bookmark_for(document)
     bookmarks_for_documents([document]).first
   end
+
+  ##
+  # @return [String] a user-displayable login/identifier for the user account
+  def to_s
+    email
+  end
 end
--- a/app/models/concerns/blacklight/user.rb
+++ b/app/models/concerns/blacklight/user.rb
@@ -30,8 +30,21 @@ module Blacklight::User
   ##
   # @return [String] a user-displayable login/identifier for the user account
   def to_s
-    return email if respond_to?(:email)
+    string_display_key = self.class.string_display_key
+    return send(string_display_key) if respond_to?(string_display_key)
 
     super
+  end
+
+  module ClassMethods
+    def string_display_key(attr_name = nil)
+      return @string_display_key if attr_name.blank?
+
+      @string_display_key = attr_name
+    end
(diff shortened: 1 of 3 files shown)


app/models/concerns/blacklight/user.rb
```diff
@@ -28,4 +28,10 @@ def document_is_bookmarked?(document)
   def existing_bookmark_for(document)
     bookmarks_for_documents([document]).first
   end
+
+  ##
+  # @return [String] a user-displayable login/identifier for the user account
+  def to_s
+    email
```
cbeer: I think `email` is something we get from devise, and I think the goal of putting it in the generator was not to tie others to that particular implementation?

cbeer (dismissed):

app/models/concerns/blacklight/user.rb
no-reply: Does inheriting it tie them to that implementation any more than generating it into their code?

The problem I'm encountering with this approach is that it's awkward to introduce an alternate implementation anywhere other than the application.

no-reply: see, e.g.: [`Blacklight::AccessControls::User#user_key`](https://github.com/projectblacklight/blacklight-access_controls/blob/master/lib/blacklight/access_controls/user.rb#L16-L21) and `Hyrax::User#display_name`. Arguably, both of these should provide an alternate default `#to_s` implementation, but they can't achieve this without editing files in the application's tree.

no-reply:
@cbeer: what you you think about this with the changes?

barmintor:
Rebased on `main`

barmintor:
I think @no-reply is making a valuable observation here, but I think @cbeer is justifiably reluctant to have some of the Devise-specific stuff move out of the generator and into the mixin. I'm adding a commit that I think splits the difference to see if we can move this conversation forward for BL 8 - it should permit downstream app's to have a hook for easy overrides, and be backwards compatible with apps that have overridden `#to_s` in the past, or kept the previously generated method.

Thoughts?

app/models/concerns/blacklight/user.rb
```diff
@@ -33,8 +34,21 @@ def existing_bookmark_for(document)
...
   def to_s
-    return email if respond_to?(:email)
+    string_display_key = self.class.string_display_key
+    return send(string_display_key) if respond_to?(string_display_key)
 
     super
   end
+
+  module ClassMethods
+    def string_display_key(attr_name = nil)
```
jcoyne, line 44: Can you use `class_attribute :string_display_key, default: :email` ?   That means we don't need all this code.

Aside from that I think this approach works.

jcoyne (changes requested):

app/models/concerns/blacklight/user.rb
barmintor: I think part of @cbeer's reservation initially is that the `:email` attribute only exists if you have set up Devise, which is not strictly required. But I think you're right about using the class attribute - I could just have the generated code from Devise setup be `string_display_key ||= :email` (we don't want to assign unconditionally because the generated code shouldn't stomp on mixin assignments). I'll make that change!

lib/generators/blacklight/user_generator.rb
```diff
@@ -36,12 +36,10 @@ def generate_devise_assets
...
       # add the #to_s to the model.
       insert_into_file("app/models/#{model_name}.rb", before: /end(\n| )*$/) do
-        "\n  # Method added by Blacklight; Blacklight uses #to_s on your\n" \
+        "\n  # Configuration added by Blacklight; Blacklight::User uses a method key on your\n" \
           "  # user class to get a user-displayable login/identifier for\n" \
           "  # the account.\n" \
-          "  def to_s\n" \
-          "    email\n" \
-          "  end\n"
+          "  self.string_display_key = :email unless string_display_key?\n"
```
jcoyne, line 42: Could you use `self.string_display_key ||= :email` ?

jcoyne (changes requested):

lib/generators/blacklight/user_generator.rb
barmintor: Sold!

jcoyne (approved):

no-reply:
this solution suits me. thanks for reviving this @barmintor

PR #2063, opened 2019-02-01
Improve performance of Search.delete_old_searches

### LINKED ISSUE #2062: Search.delete_old_searches should use batches to reduce memory consumption
[Search.delete_old_searches](https://github.com/projectblacklight/blacklight/blob/master/app/models/search.rb#L13) calls destroy_all, which [loads all relations into memory](https://github.com/rails/rails/issues/22510) and can cause problems with even moderate search volume.

Pending an upstream fix (that issue was opened in 2015…), we can workaround this with manual batching, e.g.:

```ruby
where(['created_at < ? AND user_id IS NULL', Time.zone.today - days_old]).
  in_batches{|batch| batch.destroy_all}
```

danschmidt5189:
Closes #2062

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -15,6 +15,6 @@ class Search < ApplicationRecord
     raise ArgumentError, 'days_old is expected to be a number' unless days_old.is_a?(Numeric)
     raise ArgumentError, 'days_old is expected to be greater than 0' if days_old <= 0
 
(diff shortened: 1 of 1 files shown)


danschmidt5189:
@jcoyne Updated to use delete_all. I noted the potential problems in the commit message.

jcoyne:
Thanks @danschmidt5189. Great contribution!

PR #2064, opened 2019-02-04
Add merge_facet method to include in solr response.

dkinzer:
This commit adds the method `merge_facet` which comes in handy when
creating a federated facets that merges results from multiple sources.

This seems to be a missing method since it is already possible to add
a :url_method for configuring facet fields, which suggest it's already
possible to potentially federate a facet field.


jcoyne:
I don't understand why someone would choose to use this. Can you give your use case?

dkinzer:
We use this to merge some non solr "facets" into the facet .. i.e.
 
Library_Search

PR #2065, opened 2019-02-04
Rewrite the gemfile to require a version of sqlite3 that works with o…

cbeer:
…lder versions of rails

### DIFF
--- a/spec/test_app_templates/lib/generators/test_app_generator.rb
+++ b/spec/test_app_templates/lib/generators/test_app_generator.rb
@@ -4,6 +4,17 @@ require 'rails/generators'
 class TestAppGenerator < Rails::Generators::Base
   source_root File.expand_path('../../../test_app_templates', __dir__)
 
+  def fix_sqlite3_version_requirement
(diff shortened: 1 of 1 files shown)


no-reply:
@cbeer this seems like it's not doing what's intended.

spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -4,6 +4,17 @@
...
   source_root File.expand_path('../../../test_app_templates', __dir__)
 
+  def fix_sqlite3_version_requirement
+    return unless Gem.loaded_specs['rails'].version.to_s <= '5.2.2'
+
+    # Hack for https://github.com/rails/rails/issues/35153
+    gemfile = File.expand_path('Gemfile')
+    IO.write(gemfile, File.open(gemfile) do |f|
+      text = f.read
+      text.gsub(/^gem "sqlite3"$/, 'gem "sqlite3", "~> 1.3.6"')
```
eliotjordan: Should be `/^gem 'sqlite3'$/`.

PR #2073, opened 2019-02-25
Update the front-end package to use bootstrap 4.3.1

jcoyne:
Fixes CVE-2019-8331


app/assets/javascripts/blacklight/blacklight.js
```diff
@@ -28,16 +27,15 @@ Blacklight = function () {
       return listeners;
     }
   };
-}();
-
-// turbolinks triggers page:load events on page transition
+}(); // turbolinks triggers page:load events on page transition
```
cbeer: What's the relationship between `app/assets/javascripts` and `app/javascripts`? It seems like you've (or someone?) made some changes to one that aren't reflected in the other?

jcoyne: `app/assets/javascripts` is the compiled assets from `app/javascripts`. In a pure JS project the former would be `dist` and the latter `src`

jcoyne: I think the whitespace difference is from the updated version of babel.

cbeer: got it; it's made some really weird choices about where comments go, i guess.

PR #2083, opened 2019-03-07
Encapsulate the blacklight config

jcoyne:
All values set on Blacklight::Engine.config are actually stored
in a class variable on Railtie::Configuration. So these pollute
that namespace.  All other railties are encapsulating their configs
onto a single entry in that scope.  This change makes Blacklight do
likewise.

This probably needs to be a major revision

### DIFF
--- a/app/controllers/concerns/blacklight/bookmarks.rb
+++ b/app/controllers/concerns/blacklight/bookmarks.rb
@@ -16,7 +16,7 @@ module Blacklight::Bookmarks
     before_action :verify_user
 
     blacklight_config.track_search_session = false
-    blacklight_config.http_method = Blacklight::Engine.config.bookmarks_http_method
+    blacklight_config.http_method = Blacklight::Engine.config.blacklight.bookmarks_http_method
     blacklight_config.add_results_collection_tool(:clear_bookmarks_widget)
 
     blacklight_config.show.document_actions[:bookmark].if = false if blacklight_config.show.document_actions[:bookmark]
--- a/app/controllers/concerns/blacklight/catalog.rb
+++ b/app/controllers/concerns/blacklight/catalog.rb
@@ -257,13 +257,13 @@ module Blacklight::Catalog
   end
 
   def sms_mappings
-    Blacklight::Engine.config.sms_mappings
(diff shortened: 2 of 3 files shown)


bess:
+1 but it seems like we should talk about when to merge this, if we think it's a major version increment. Would it make sense to mark this as a Blacklight 8 feature?

barmintor:
Now that the 8.0.0.alpha tag is in [main](https://github.com/projectblacklight/blacklight/commit/a82e70484ba89ea84380a213dee89c3cbd3acb95), can we revive this PR?

barmintor:
Rebasing for CI before merge.

barmintor:
The OpenStruct linter rule was added in the interim of this PR being opened. There's three approaches available to us:

1. Use OpenStructWithHashAccess here, and remove it (or not) along with the rest of the usages of that class
2. Add to `rubocop_todo`
3. Disable the `Style/OpenStructUse` linter rule.

I think there's an argument for disabling the rule rather than adding yet another `todo`, but I'm also sympathetic to using OpenStructWithHashAccess and dealing with that (or not!) as a single task in the future.

tpendragon:
https://github.com/rubocop/rubocop/issues/10206 is where they added this, and it makes a good point that Ruby 3 basically says "quit it."

I'm for option 1, and actually removing that class one day (although I suspect it's gonna be hard :( )

barmintor:
Sold @tpendragon. I'll merge after CI.

cbeer:
@jcoyne  or @barmintor, can you create a PR against the release-7.x with appropriate deprecations and a migration path for this change?

jcoyne:
@cbeer I wouldn't know how to do that as `Blacklight::Engine.config` is pointing at the `Railtie::Configuration`. Since that's not a class we control, we can't add deprecations to it.

barmintor:
It's true that it might be awkward, but I created https://github.com/projectblacklight/blacklight/issues/2607 to try to work through it @cbeer

barmintor:
Ok, I have a proposed deprecation up in https://github.com/projectblacklight/blacklight/pull/2610

Spoiler: It _was_ kind of awkward.

PR #2084, opened 2019-03-13
Accept everything as the default routing constraint

jcoyne:


lib/blacklight/engine.rb
```diff
@@ -45,7 +45,7 @@ class Engine < Rails::Engine
 
     config.routes = OpenStruct.new
     # Set identifier_constraint to enforce a format for the document identifiers
-    config.routes.identifier_constraint = /[\w-]+/
+    config.routes.identifier_constraint = /.+/
```
cbeer: Won't this break anyone using URLs with extensions?

jcoyne: Explain what you mean by that?

jcoyne: Keep in mind this is only for the "track" route: https://github.com/projectblacklight/blacklight/blob/v6.20.0/config/routes.rb#L10

barmintor:
Added 2 routing specs to verify per @cbeer's review that format in the URL is not interfered with. Merging to `release-6.x`

PR #2085, opened 2019-03-15
support multiple flash messages of each type

dunn:
this seems to be the standard way for handling it: https://stackoverflow.com/questions/23321623/for-rails-app-how-to-show-multiple-flashnotice-in-html


jcoyne:
@dunn does Blacklight itself ever set more than one message at a time?

dunn:
@jcoyne I don't know, but as it's an engine it should be flexible with regard to the views and controllers that the application provides.

PR #2089, opened 2019-03-28
Fix a field name for author in the example schema

jcoyne:
The current value is not a dynamic field and causes an error in Solr 8


jcoyne:
I'm merging this because it's less failing (API build works at least) than master.

PR #2104, opened 2019-05-23
this class appears to be not in use in anymore

cdmo:
Small potatoes, but this appears to be a small bit of extra code that might be worth removing


cdmo:
I guess I was thinking it was a bit odd that _this one element_ has a class attached, but others don't have extra classes with no style attached. The rest of the codebase seems to assume you'll override the view (or add classes through a helper method like `LayoutHelper::sidebar_classes`). The naming convention of the class is a bit odd too compared to all the other classes/ids I've encountered so far. Just seemed like a thing that got missed, but if it's needed/wanted, of course just close the PR. Thanks.

PR #2106, opened 2019-06-03
Forcing 'counter' to integer to avoid XSS

ta1986:
Appscan claims it successfully embedded a script through 'counter' of Previous/Next links. When users click on modified Previous/Next links, the script will be executed.
To avoid unwanted scripts execute when page loads, 'counter' should be sanitized.

### DIFF
--- a/app/helpers/blacklight/catalog_helper_behavior.rb
+++ b/app/helpers/blacklight/catalog_helper_behavior.rb
@@ -102,7 +102,7 @@ module Blacklight::CatalogHelperBehavior
   # @see #page_entries_info
   # @return [String]
   def item_page_entry_info
-    t('blacklight.search.entry_pagination_info.other', current: number_with_delimiter(search_session['counter']),
+    t('blacklight.search.entry_pagination_info.other', current: number_with_delimiter(search_session['counter'].to_i),
                                                        total: number_with_delimiter(search_session['total']),
                                                        count: search_session['total'].to_i).html_safe
   end


jcoyne:
@ta1986 can you add an integration test to illustrate how this will fix the problem?

ta1986:
Hmm, I don't have a good test yet. Issue can be reproduced by:
1. Do a search:
https://demo.projectblacklight.org/?search_field=all_fields

2. Inspect -> Replace in the HTML:
/catalog/2009002583/track?counter=1
With:
/catalog/2009002583/track?counter=<script>alert(123)</script>

3. Now click on the first item to see the 'alert'

jcoyne:
@ta1986 how would an attacker mutate that link to be anything other than the default value?

ta1986:
@jcoyne You are right. It's a false alarm. Attackers are unable to mutate that URL.
However, the fix helps Blacklight bypass IBM Appscan and Fortify security scans in my institution.

jcoyne:
@ta1986 Okay, can you please add a comment in the code that explains that, and add a unit test test for the helper?

barmintor:
@ta1986 thanks for bringing this issue to the community's attention. While this code can no longer be merged, you should know that a similar guard is in place in the [current code](https://github.com/projectblacklight/blacklight/blob/fd33c74a378d164a571174fa0054c25b70ef09d1/app/components/blacklight/search_context_component.rb#L45). If you see this issue resurface in a modern release of Blacklight, please open a new issue.

PR #2117, opened 2019-06-27
Return a default presenter instead of nil.

dkinzer:
The `presenter` helper method should return a default presenter instead of
nil.  Otherwise an error occurs when you have a new action that is
not one of "show", "citation", or "index".

### DIFF
--- a/app/helpers/blacklight/blacklight_helper_behavior.rb
+++ b/app/helpers/blacklight/blacklight_helper_behavior.rb
@@ -260,7 +260,7 @@ module Blacklight::BlacklightHelperBehavior
     case action_name
     when 'show', 'citation'
       show_presenter(document)
-    when 'index'
+    else
       index_presenter(document)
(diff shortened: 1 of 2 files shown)


jcoyne:
I don't understand why you'd want an IndexPresenter for something that isn't an index view.  Can you tell me why that would be what you want?

dkinzer:
So in this particular case I'm rendering something via ajax using some index view partials.  So it works, but I guess my problem is more with the fact that we are returning nil and then later in some views assuming the this command does not return nil.

I'm not sure why this wasn't failing for me when we were on BL-6 but now it is.

I think at the very least this method should not be returning nil ever.

Maybe there should be way for an action to register a specific presenter?

If you think this is not useful  I can close this PR and just continue to override in our instance.

bess:
I think this looks like a helpful change and I'd like to merge it.

PR #2119, opened 2019-07-11
RecordMailer: clean up formatting

dunn:
## add an extra linebreak between records
Screen Shot 2019-07-10 at 6 55 38 PM

## remove array cruft from record title
Screen Shot 2019-07-10 at 6 55 45 PM

### DIFF
--- a/app/models/record_mailer.rb
+++ b/app/models/record_mailer.rb
@@ -7,7 +7,10 @@ class RecordMailer < ActionMailer::Base
             rescue
               I18n.t('blacklight.email.text.default_title')
             end
-    subject = I18n.t('blacklight.email.text.subject', count: documents.length, title: title)
+
+    subject = I18n.t('blacklight.email.text.subject',
+                     count: documents.length,
+                     title: Array(title).first)
 
     @documents      = documents
     @message        = details[:message]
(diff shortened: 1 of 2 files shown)


app/models/record_mailer.rb
```diff
@@ -7,7 +7,7 @@ def email_record(documents, details, url_gen_params)
             rescue
               I18n.t('blacklight.email.text.default_title')
             end
-    subject = I18n.t('blacklight.email.text.subject', count: documents.length, title: title)
+    subject = I18n.t('blacklight.email.text.subject', count: documents.length, title: title.first)
```
jkeck: Is this guaranteed to be an array?

dunn: I think yes, unless the user application overrides the title field to be single-valued?

jkeck: Yeah, but they could do that theoretically do that I believe, yeah? (and with something like the canonical title of an item, that might be more common).

dunn: Yeah, that's true.  I'll take another pass at this.

dunn: `Array(title).first` should work.

PR #2128, opened 2019-09-05
Remove jquery dependency from search_context.js

jcoyne:
Bootstrap 5 is removing the jquery dependency, so this gets Blacklight moving towards not requiring that dependency.

### DIFF
--- a/app/assets/javascripts/blacklight/blacklight.js
+++ b/app/assets/javascripts/blacklight/blacklight.js
@@ -431,54 +431,68 @@ Blacklight.onLoad(function () {
   Blacklight.modal.setupModal();
 });
 
-(function ($) {
-  Blacklight.doSearchContextBehavior = function () {
-    if (typeof Blacklight.do_search_context_behavior == 'function') {
-      console.warn("do_search_context_behavior is deprecated. Use doSearchContextBehavior instead.");
-      return Blacklight.do_search_context_behavior();
-    }
+Blacklight.doSearchContextBehavior = function () {
+  if (typeof Blacklight.do_search_context_behavior == 'function') {
+    console.warn("do_search_context_behavior is deprecated. Use doSearchContextBehavior instead.");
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/blacklight/blacklight.js
```diff
@@ -432,54 +432,68 @@ Blacklight.onLoad(function () {
   Blacklight.modal.setupModal();
 });
 
-(function ($) {
-  Blacklight.doSearchContextBehavior = function () {
-    if (typeof Blacklight.do_search_context_behavior == 'function') {
-      console.warn("do_search_context_behavior is deprecated. Use doSearchContextBehavior instead.");
-      return Blacklight.do_search_context_behavior();
-    }
+Blacklight.doSearchContextBehavior = function () {
```
jcoyne: Note: this code is the compiled distribution. The source is in `app/javascript/`

app/javascript/blacklight/search_context.js
```diff
@@ -1,49 +1,67 @@
...
+  document.body.appendChild(form)
+  form.querySelector('[type="submit"]').click()
+
+  event.preventDefault()
+  event.stopPropagation()
+  event.stopImmediatePropagation()
+};
+
+Blacklight.onLoad(function() {
+    Blacklight.doSearchContextBehavior();
```
mejackreed:
```suggestion
  Blacklight.doSearchContextBehavior();
```

jcoyne: Isn't this the job of jslint/jshint/hound?

jcoyne: Seriously, why isn't hound catching this?  It's in the .jshint config.

mejackreed: I'm thinking we should move away from jshint and hound and to an updated eslint config that just gets raised on CI. Hound / jshint hasn't been that reliable

PR #2129, opened 2019-09-05
Remove jQuery dependecy from facet_load.js

jcoyne:
Bootstrap 5 is dropping jquery, so this aligns with that move.


app/assets/javascripts/blacklight/blacklight.js
```diff
@@ -229,29 +229,29 @@ Blacklight.onLoad(function () {
 /*global Blacklight */
 
 
-(function ($) {
-  'use strict';
+'use strict';
```
jcoyne: Note: this code is the compiled distribution. The source is in `app/javascript/`

PR #2140, opened 2019-09-19
Enable new functionality for pagination information on grouped results

mejackreed:
Enables https://github.com/projectblacklight/arclight/issues/699 and replacing approach in https://github.com/projectblacklight/arclight/pull/775.

A few things about this pull request. At first I aimed to target removal of `page_entries_info` in a Blacklight 8 release by removing our custom i18n keys and just using Kaminari's. This would have worked except for the two remaining Blacklight needs:
 - We need to change the math for grouped results
 - We need to provide delimited number strings

We could accmmodate either one of those I think independently, but together I couldn't figure out a way to successfully push that to the `Blacklight::Solr::Response` in a sane way.

I've hopefully simplified this method a bit so that future Blacklight maintainers have an easier time. 🤞

### DIFF
--- a/app/helpers/blacklight/catalog_helper_behavior.rb
+++ b/app/helpers/blacklight/catalog_helper_behavior.rb
@@ -31,8 +31,11 @@ module Blacklight::CatalogHelperBehavior
 
   ##
   # Override the Kaminari page_entries_info helper with our own, blacklight-aware
-  # implementation.
-  # Displays the "showing X through Y of N" message.
+  # implementation. Why do we have to do this?
+  #  - We need custom counting information for grouped results
+  #  - We need to provide number_with_delimiter strings to i18n keys
+  # If we didn't have to do either one of these, we could get away with removing
+  # this entirely.
   #
   # @param [RSolr::Resource] collection (or other Kaminari-compatible objects)
   # @return [String]
--- a/app/helpers/blacklight/catalog_helper_behavior.rb
+++ b/app/helpers/blacklight/catalog_helper_behavior.rb
@@ -46,7 +46,7 @@ module Blacklight::CatalogHelperBehavior
                  elsif collection.respond_to?(:model_name) && !collection.model_name.nil? # AR, Blacklight::PaginationMethods
                    collection.model_name.human.downcase
                  else
-                   t('blacklight.entry_name.default')
+                   collection.entry_name(count: collection.size)
                  end
 
     entry_name = entry_name.pluralize unless collection.total_count == 1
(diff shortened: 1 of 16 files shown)


lib/blacklight/solr/response/group_response.rb
```diff
@@ -39,6 +39,17 @@ def empty?
     total.zero?
   end
 
+  ##
+  # Overridden from Blacklight::Solr::Response::PaginationMethods to support
+  # grouped key specific i18n keys. `key` is the field being grouped
+  def entry_name(options)
+    begin
+      I18n.t("blacklight.entry_name.grouped.#{key}", raise: true)
```
cbeer: I think this can be collapsed using I18n fallbacks (and its pluralization methods?), something like:

```ruby
I18n.t("blacklight.entry_name.grouped.#{key}", default: 'blacklight.entry_name.grouped.default', count: options[:count])
```

mejackreed: 👍 I opted to leave the count out as that would require additional keys. Just using the rails helper for pluralization.

app/helpers/blacklight/catalog_helper_behavior.rb
```diff
@@ -31,22 +31,21 @@ def json_api_link_tag(options = {})
...
   # @return [String]
-  def page_entries_info(collection, options = {})
+  def page_entries_info(collection, entry_name: nil)
     return unless show_pagination? collection
 
-    entry_name = if options[:entry_name]
-                   options[:entry_name]
-                 elsif collection.respond_to? :model  # DataMapper
-                   collection.model.model_name.human.downcase
-                 elsif collection.respond_to?(:model_name) && !collection.model_name.nil? # AR, Blacklight::PaginationMethods
```
dkinzer: This change is breaking our site which uses both solr and non solr documents.   because collection.model_name.downcase throws an error when collection.model_name is nil.

app/helpers/blacklight/catalog_helper_behavior.rb
```diff
@@ -31,22 +31,21 @@ def json_api_link_tag(options = {})
...
-                   collection.model_name.human.downcase
+    entry_name = if entry_name
+                   entry_name.pluralize(collection.size, I18n.locale)
                  else
-                   t('blacklight.entry_name.default')
+                   collection.entry_name(count: collection.size).downcase
```
dkinzer: Actually our issue is when `collection.entry_name` returns nil we get an error.  We pushed up a PR to fix this.

PR #2141, opened 2019-09-20
Prevent the click on the facet heading from jumping to the anchor

jcoyne:
We just want the panel to open close without moving the viewport.

Related to #2132


cdmo:
Just want to link this issue to the PR I put in https://github.com/projectblacklight/blacklight/pull/2137

I _believe_ that the collapse.js is supposed to be doing this in the first place, but isn't working either (I could be wrong on that!).

Regardless though, if you use buttons you don't have to worry about links jumping the page down.

bess:
@jcoyne and @mejackreed Can this PR be closed? Was it indeed superseded by #2137?

mejackreed:
Yes.. thanks for following up @bess

PR #2145, opened 2019-09-20
Add jump links to pages

### LINKED ISSUE #2144: Add jump links to pages
@mejackreed pointed out that [Arclight issue #840](https://github.com/projectblacklight/arclight/issues/840) could just as well be done here.

Quoting that ticket body:

----
Example implementation here: https://searchworks.stanford.edu - I copied this HTML and CSS from there and adjusted it a bit for visual style; but if there's a better way to implement this, feel free.

- [x] **Add a skip link section as the first div in the `<body>` tag.**

   ```
   <div id="skip-link">
       <a href="#search_field" class="element-invisible element-focusable" data-turbolinks="false">Skip to search</a>
    <a href="#main-container" class="element-invisible element-focusable" data-turbolinks="false">Skip to main content</a>
   </div>
   ```
"Skip to search" and "Skip to main content" should be present on all pages. 

- [x] **Style the links so they are invisible until they receive focus via tab key.**

   ```
   #skip-link .element-invisible {
       position: absolute !important;
       clip: rect(1px, 1px, 1px, 1px);
       overflow: hidden;
       height: 1px;
   }
   #skip-link {
       position: absolute;
       top: 5px;
       left: 10px;
       z-index: 1;
   }
   #skip-link .element-invisible.element-focusable:active, #skip-link .element-invisible.element-focusable:focus {
       position: static !important;
       clip: auto;
       overflow: visible;
       height: auto;
   }
   #skip-link a:focus {
       background-color: white;
       border-radius: 0 0 0.2rem 0.2rem;
       padding: .5rem 1rem;
   }
   ```

- [x] **Add a third link on search results pages**

   `<a href="#documents" class="element-invisible element-focusable" data-turbolinks="false">Skip to first result</a>`

estelendur:
Closes #2144 

See my concern in the comments there:
> The search bar has autofocus on "catalog-like index page[s] without query or facet parameters", including the front page, which means that the skip links are not the first thing in the tab order. What should we do with this? Remove autofocus? Leave it? Something more complicated?

Resolves [Arclight #840](https://github.com/projectblacklight/arclight/issues/840).


app/views/catalog/_search_results.html.erb
```diff
@@ -7,6 +7,10 @@
   <%= json_api_link_tag %>
 <% end %>
 
+<% content_for(:skip_links) do -%>
+  <a href="#documents" class="element-invisible element-focusable" data-turbolinks="false">Skip to first result</a>
```
jkeck: I believe we could use `link_to` here (although not super important).  Also, we should internationalize this value. 

I'm not 100% sure what our process is for adding new UI strings these days (e.g. skip the specs that will fail if we add a new key to only one language, use Google Translate to translate the value for us, some other option).

app/views/layouts/blacklight/base.html.erb
```diff
@@ -17,19 +17,24 @@
     <%= content_for(:head) %>
   </head>
   <body class="<%= render_body_class %>">
-  <%= render partial: 'shared/header_navbar' %>
+    <div id="skip-link">
+      <a href="#search_field" class="element-invisible element-focusable" data-turbolinks="false">Skip to search</a>
+       <a href="#main-container" class="element-invisible element-focusable" data-turbolinks="false">Skip to main content</a>
```
jkeck: Same comment here wrt `link_to` and internationalization.

app/assets/stylesheets/blacklight/_header.scss
```diff
@@ -53,3 +53,31 @@
...
+  .element-invisible.element-focusable:active,
+  .element-invisible.element-focusable:focus {
+    position: static !important;
+    clip: auto;
+    overflow: visible;
+    height: auto;
+  }
+
+  a:focus {
+    background-color: white;
```
jkeck: Can we use the variable `$white` here instead of the hardcoded color?

app/assets/stylesheets/blacklight/_header.scss
```diff
@@ -53,3 +53,31 @@
...
+    border-radius: 0 0 0.2rem 0.2rem;
+    padding: .5rem 1rem;
```
jkeck: I believe we can take advantage of some Bootstrap utility classes for [border radius](https://getbootstrap.com/docs/4.0/utilities/borders/#border-radius) and [spacing](https://getbootstrap.com/docs/4.3/utilities/spacing/#notation) instead of hardcoding them here.

app/views/catalog/_search_results.html.erb
mejackreed: I've been using Google Translate 🤷‍♂

app/assets/stylesheets/blacklight/_header.scss
estelendur: Is it possible to use those Bootstrap classes in the scss file, or do I just add them to each skip link in the views?

estelendur: Eh, I just added them to each skip link. There's only three right now.

app/views/catalog/_search_results.html.erb
```diff
@@ -7,6 +7,10 @@
   <%= json_api_link_tag %>
 <% end %>
 
+<% content_for(:skip_links) do -%>
+    <%= link_to t('blacklight.skip_links.first_result'), "#documents", class: "element-invisible element-focusable rounded-bottom py-2 px-3", data: {turbolinks: "false"} %>
```
cbeer:
```suggestion
    <%= link_to t('blacklight.skip_links.first_result'), '#documents', class: 'element-invisible element-focusable rounded-bottom py-2 px-3', data: { turbolinks: false } %>
```

PR #2153, opened 2019-09-24
Enable a configuration for skipping a saved search that uses the requ…

mejackreed:
…est and params object

Very similar to approach for crawler, but semantically a bit different?

What do others think?

### DIFF
--- a/app/controllers/concerns/blacklight/search_context.rb
+++ b/app/controllers/concerns/blacklight/search_context.rb
@@ -38,7 +38,7 @@ module Blacklight::SearchContext
   end
 
   def find_search_session
-    if agent_is_crawler?
+    if agent_is_crawler? || skip_session_tracking?
       nil
     elsif params[:search_context].present?
       find_or_initialize_search_session_from_params JSON.parse(params[:search_context])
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/blacklight/search_context.rb
```diff
@@ -38,7 +38,7 @@ def set_current_search_session
   end
 
   def find_search_session
-    if agent_is_crawler?
+    if agent_is_crawler? || skip_session_tracking?
```
cbeer: Should we go ahead and push `agent_is_crawler?` into the `skip_session_tracking?` method?

app/controllers/concerns/blacklight/search_context.rb
```diff
@@ -78,6 +78,13 @@ def agent_is_crawler?
     crawler_proc.call(request)
   end
 
+  def skip_session_tracking?
```
cbeer: I haven't look at what `track_search_session` is, but what if we let it be a boolean and/or a lambda?

app/controllers/concerns/blacklight/search_context.rb
cbeer: Or, should this check get pushed up to `record_search_parameters` or `before_action :set_current_search_session, if: ...` around line 15?

app/controllers/concerns/blacklight/search_context.rb
mejackreed: Yeah that might be a better extendable approach. I'll look into that.

app/controllers/concerns/blacklight/search_context.rb
mejackreed: 👍

barmintor:
Not sure if people still want this, but it was in an awkward state of failing tests/requested changes/approved/needed rebase.

I've rebased and put the fixups originally requested in. Do we still want this, or can we close this PR?

PR #2154, opened 2019-09-26
[WIP] Basic overridable schema.org json-ld support

mejackreed:
The idea here is that others would use the document rendering chain to put custom schema.org json-ld support.

Tried figuring out how presenters might help us.. I couldn't.

### DIFF
--- a/app/views/catalog/_schema_org.html.erb
+++ b/app/views/catalog/_schema_org.html.erb
@@ -1,0 +1,5 @@
+<script type="application/ld+json">
+  {
+    "@context": <%= document.itemtype %>,
+  }
+</script>


app/views/catalog/_schema_org.html.erb
```diff
@@ -0,0 +1,5 @@
+<script type="application/ld+json">
+  {
+    "@context": <%= document.itemtype %>,
```
jcoyne: Shouldn't the `@context` be `http://schema.org` and the `@type` should be `document.itemtype`

mejackreed: Yes most definitely Dropping a WIP on this.

barmintor:
This PR is abandoned, but linked from #2703 if someone want to take it up on modern Blacklight.

PR #2157, opened 2019-10-02
Add enhanced API for Blacklight::Icons handling cases for semantic an…

### LINKED ISSUE #2150: Proposal for svg icon perception by screen readers
This issue is based off of discussion in https://github.com/projectblacklight/arclight/issues/866 and a [preexisting implementation in GeoBlacklight](https://github.com/geoblacklight/geoblacklight/pull/840).

The overall goal here is to make svg icons perceivable and more accessible in a few different ways.

## Background
As mentioned a few times in the Blacklight, GeoBlacklight communities, taking a look at [FontAwesome's guide on accessible icons](https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility) is useful.

We need to provide an API that can accommodate:
 - Semantic icons - icon has semantic meaning
 - Decorative icon - an icon that is only decorative

### Considerations
 - We also need to account for that SVG's will not come with the embedded accessibility needs. 
 - This should be i18n friendly
 - Configurable for based off a situation

## Semantic Icons Needs
From https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility#svg-semantic
 - `role="image"`
 - `title` tag in `svg`
 - `aria-label="[iconname i18n]"` 

## Decorative Icon Needs
From https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility#svg-semantic
 - `role="image"`
 - `aria-hidden` already supported in https://github.com/projectblacklight/blacklight/pull/2126

## Proposal
 - [ ] Enable support of [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) to be added to an svg with a lookup path using i18n
 - [ ] Enable `title` element to also be configurable via `i18n` and `Blacklight::Icon` api.
 - [ ] Add role customization , default to `image`

## Controversy
There seems to be some divergence around usage of `aria-label` or adding additional markup to support `aria-labelled-by` (which is what GeoBlacklight ended up doing). I would support going with a simpler `aria-label` approach if it is supported by major screen readers and accomplishes our needs.

mejackreed:
…d decorative icons fixes #2150

~~I'd like some verification from @jvine on one last item before :shipit: ~~

Update: We went with the `aria-labelled-by` + `title` approach here.

### DIFF
--- a/app/models/blacklight/icon.rb
+++ b/app/models/blacklight/icon.rb
@@ -2,22 +2,39 @@
 
 module Blacklight
   class Icon
-    attr_reader :icon_name
+    attr_reader :icon_name, :aria_hidden, :label, :role
     ##
     # @param [String, Symbol] icon_name
     # @param [Hash] options
     # @param [String] classes additional classes separated by a string
-    def initialize(icon_name, classes: '', aria_hidden: false)
+    # @param [Boolean] aria_hidden include aria_hidden attribute
+    # @param [Boolean] label include <title> and aria-label as part of svg
+    # @param [String] role role attribute to be included in svg
+    def initialize(icon_name, classes: '', aria_hidden: false, label: true, role: 'image')
       @icon_name = icon_name
       @classes = classes
       @aria_hidden = aria_hidden
+      @label = label
+      @role = role
     end
 
     ##
-    # Returns the raw source, but you could extend this to add additional attributes
+    # Returns an updated version of the svg source
     # @return [String]
     def svg
-      file_source
+      svg = ng_xml.at_xpath('svg')
+      svg['role'] = role
+      svg['aria-labelled-by'] = unique_id if label
+      svg.add_child("<title id='#{unique_id}'>#{icon_label}</title>") if label
+      ng_xml.to_xml
+    end
+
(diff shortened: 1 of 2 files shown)


app/models/blacklight/icon.rb
```diff
@@ -2,30 +2,43 @@
...
+    attr_reader :icon_name, :aria_hidden, :label, :role
     ##
     # @param [String, Symbol] icon_name
     # @param [Hash] options
     # @param [String] classes additional classes separated by a string
-    def initialize(icon_name, classes: '', aria_hidden: false)
+    # @param [Boolean] aria_hidden include aria_hidden attribute
+    # @param [Boolean] label include <title> and aria-label as part of svg
+    # @param [String] role role attribute to be included in svg
+    def initialize(icon_name, classes: '', aria_hidden: false, label: true, role: 'image')
```
cbeer: could we just pass through e.g. `aria: { hidden: true }` and pass all those properties (whatever they may be) through? Or even `html_options: {}`?

PR #2163, opened 2019-10-03
Support for rails 6

jcoyne:


spec/views/catalog/_show_tools.html.erb_spec.rb
```diff
@@ -15,7 +15,11 @@
     let(:document_actions) { blacklight_config.show.document_actions }
 
     it "renders a document action" do
-      allow(view).to receive(:some_action_solr_document_path).with(document).and_return "x"
+      if Rails.version < '6.0.0'
+        allow(view).to receive(:some_action_solr_document_path).with(document).and_return "x"
+      else
+        allow(view).to receive(:some_action_solr_document_path).with(document, only_path: true).and_return "x"
```
cjcolvar: Is it important to strictly test the argument `only_path: true`?  If not, then I think these two cases could be condensed into `allow(view).to receive(:some_action_solr_document_path).with(document, any_args).and_return "x"`

app/views/catalog/_constraints_element.html.erb
```diff
@@ -13,7 +13,8 @@
       <span class="filter-name"><%= label %></span>
     <% end %>
     <% unless value.blank? %>
-      <%= content_tag :span, value, class: 'filter-value', title: value %>
+      <% title = value.gsub(/<|>/, '') # the title could be html %>
```
cjcolvar: If this is for stripping html tags would it be better to use the standard `strip_tags`?  Or does `content_tag` already do that stripping and you're trying to partially avoid it?

blacklight.gemspec
```diff
@@ -25,15 +25,15 @@ Gem::Specification.new do |s|
...
+  s.add_dependency "rails", '>= 5.1', '< 7'
   s.add_dependency "globalid"
   s.add_dependency "jbuilder", '~> 2.7'
   s.add_dependency "kaminari", ">= 0.15" # the pagination (page 1,2,3, etc..) of our search results
   s.add_dependency "deprecation"
 
   s.add_development_dependency "rsolr", ">= 1.0.6", "< 3"  # Library for interacting with rSolr.
   s.add_development_dependency "solr_wrapper"
-  s.add_development_dependency "rspec-rails", "~> 3.5"
+  s.add_development_dependency "rspec-rails", "~> 4.0.0.beta2"
```
cjcolvar: Should we wait for the official 4.0.0 release?  Maybe it won't be too long from now (https://github.com/rspec/rspec-rails/pull/2071#issuecomment-532519693)?

app/views/catalog/_constraints_element.html.erb
jcoyne: `strip_tags` is exactly what is needed here. Thanks.

spec/views/catalog/_show_tools.html.erb_spec.rb
jcoyne: I think your suggestion for `any_args` is good. I'll use that.

blacklight.gemspec
jcoyne: I'm not sure this is a big concern of mine since this is just a development dependency and it works for us.

PR #2164, opened 2019-10-03
Refresh CSRF tokens when updating resubmittable checkbox

### LINKED ISSUE #2002: Bookmark change for a guest "Can't verify CSRF token authenticity"
Blacklight 7, Rails 5.2

A not-logged in User (Guest) does a search
Guest clicks a checkbox to bookmark
Guest clicks the same checkbox to un-bookmark
An alert message is displayed saying "error"

The application log looks like this:

```
Started DELETE "/bookmarks/10023235" for 127.0.0.1 at 2018-10-31 09:47:13 -0400
Processing by BookmarksController#destroy as JSON
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"2oNI6L6OmHe90HTjoH4ZcQmASQCpuuR3wwKRMW8VVUytQE+hwo3vVfbd58zcQdlWv2XbtELma+fbT7ZMTzEnfQ==", "id"=>"10023235"}
Can't verify CSRF token authenticity.
Completed 422 Unprocessable Entity in 1ms (ActiveRecord: 0.0ms)



ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken):
```

If you add the following line to the `bookmarks.rb` it goes away

`skip_before_action :verify_authenticity_token` or `protect_from_forgery unless: -> { request.format.json? }`

As in 

```rb
    copy_blacklight_config_from(CatalogController)

    skip_before_action :verify_authenticity_token ## or protect_from_forgery unless: -> { request.format.json? }
    before_action :verify_user
```

If I am understanding correctly, `protect_from_forgery unless: -> { request.format.json? }` is slightly safer and intended for API-based json formatted requests.

This has come up in a bunch of places like GitHub and StackOverflow, [here's one place](https://github.com/psu-libraries/psulib_blacklight/issues/71). Seems like a [change in Rails 5.2](https://github.com/rails/rails/issues/29193).

One weird thing is that you can un-bookmark from the My Bookmarks page, no problem.

I'm looking into a method that is better than just turning it off.

estelendur:
Co-authored-by: Christina Chortaria <actspatial@gmail.com>

Why does checkbox_submit.js exist when actually the site is using blacklight.js?

Fixes #2002.

Also fixes [Arclight #769](https://github.com/projectblacklight/arclight/issues/769).

### DIFF
--- a/app/assets/javascripts/blacklight/blacklight.js
+++ b/app/assets/javascripts/blacklight/blacklight.js
@@ -156,6 +156,7 @@ Blacklight.onLoad(function () {
         checkbox.prop('checked', state);
         label.toggleClass('checked', state);
 
+        Rails.refreshCSRFTokens();
         if (state) {
           //Set the Rails hidden field that fakes an HTTP verb
           //properly for current state action.
--- a/app/javascript/blacklight/checkbox_submit.js
+++ b/app/javascript/blacklight/checkbox_submit.js
@@ -69,6 +69,7 @@
         function updateStateFor(state) {
             checkbox.prop('checked', state);
             label.toggleClass('checked', state);
+            Rails.refreshCSRFTokens();
             if (state) {
                //Set the Rails hidden field that fakes an HTTP verb
                //properly for current state action.


jcoyne:
I can't duplicate this issue.  So I'm hesitant to say we need a fix for it.  

>Why does checkbox_submit.js exist when actually the site is using blacklight.js?

checkbox_submit.js is one component, which gets rolled up into blacklight.js

estelendur:
We are consistently experiencing this issue in Arclight, which I think uses bookmarks a bit differently than Blacklight. I actually don't know why it's extremely reproducible on the Arclight demo but doesn't seem to be reproducible on the Blacklight demo. What version of Rails is the Blacklight demo using?

We could probably move the change to Arclight, in theory, though I'm not actually sure how to sanely monkeypatch front-end JS.

So in order to see changes to checkbox_submit.js in my internal test app locally, I would have needed to rebuild the engine cart or rerun some rake task, I guess?

jcoyne:
@estelendur my current theory is this is a difference between rails-ujs and jquery_ujs (current used by demo.projectblacklight.org) I will do further investigation as soon as we can get the site to redeploy (it's currently having trouble).

jcoyne:
So, that theory is a bust as both sites are now using rails-ujs. I am seeing that demo.projectblacklight.org is setting X-CSRF-Token headers but arclight is not.

mejackreed:
I played around with this a bit this morning. I found a difference between arclight-demo and blacklight-demo which I think has an impact. 

Adding the jquery require before the rails-ujs seems to make a difference (at least in my local testing)

```diff
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
+//= require jquery3
 //= require rails-ujs
 //= require activestorage
 //= require turbolinks
 //
 // Required by Blacklight
-//= require jquery
 //= require popper
 // Twitter Typeahead for autocomplete
```

That being said.. I'm not sure what the right solution here is then. Updating this? https://github.com/projectblacklight/blacklight/blob/master/lib/generators/blacklight/assets_generator.rb#L12-L35

jcoyne:
Yeah, we should update that assets block anyway (to use jquery3 and not to add jquery if it's already there.)

mejackreed:
I'd probably say yes we still need this. I'm noticing odd cases where this still has problems and I can't fully understand is. Anyone else seeing this?

cdmo:
I'm kinda hesitant too to be honest, to include a fix for something that we don't know for sure is a problem. The details that matter in the scenario that describes the problem are (to me) what is your asset pipline? And what are your dependencies for the front end? Our issue (the original issue reported on the PR) was due 100% to not including or starting rails_ujs in our webpacker-only asset pipeline.

I haven't looked at out-of-the-box Blacklight 7.3.0 with default asset pipeline, that would be the important scenario in my opinion, and if it isn't broken there, then I wonder if adding a fix would be better done in a downstream implementation that has decided to change defaults. Just my 2 cents.

If this is included though, it would seem to me that adding a spec would be a good idea.

barmintor:
Changes elsewhere in the codebase suggest that the related issue (#2002) was fixed in https://github.com/projectblacklight/blacklight/pull/2171/commits/5841b59b8040a7a58a0c1d988aba6a7a0e304613, which was fixed in v7.2.0, and the arclight demo was updated to pull that dependency in in this commit: https://github.com/sul-dlss/arclight-demo/commit/5ab0b3a9d82c37d90b55b4720967fdab50d21704

I am going to close this PR, but if @estelendur can still observe the bug in Blacklight post-7.2, we should reopen and create a new issue.

PR #2165, opened 2019-10-03
Handle text input rounding based on the number of search fields.

jkeck:
Bootstrap should handle this for us, but the markup that the typeahead library injects ends up messing that up.

To be honest, I don't know if this is worth it or not, but the borders have been bugging me.  An alternative would be to put some right margin on the fielded search drop down and have the left of the search input always be rounded and the right always be non-rounded (but that would remove the single-input appearance of the drop-down, text input, button we currently have though)

## Before
single-field-before

---

many-fields-before

## After
_Note: There is really no change here w/o a fielded search drop-down_
single-field-after

---

many-fields-after

### DIFF
--- a/app/views/catalog/_search_form.html.erb
+++ b/app/views/catalog/_search_form.html.erb
@@ -15,7 +15,7 @@
     <% end %>
 
     <label for="q" class="sr-only"><%= t('blacklight.search.form.search.label') %></label>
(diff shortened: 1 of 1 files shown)


ggeisler:
@jkeck I like your solution, looks good.

mejackreed:
++

PR #2168, opened 2019-10-05
Create a component for Facet::List

jcoyne:
Depends on #2164 & #2166 

I'm thinking of a component based approach for the next major version of blacklight.  I'm curious as to what you all think.


magibney:
Hi @jcoyne! I think I'm interested in learning more about this, and possibly providing feedback. Would you be able to create some (or point to some existing) documentation describing the high level goals of this change -- motivating use case(s), what types of use cases this might enable, design goals, tradeoffs, etc.?

Some background on the perspective I'm bringing: I've done some work (and considered doing other work) with facets, that might benefit from a more easily customized approach: browse/cross-references-over-facets, Database-A-Z-list-over-facets, JSON-facet-API stuff (incl. facet sort by correlation (SKG/relatedness), etc.).

At a minimum this background leaves me somewhat familiar with this part of the blacklight code, but I might also be interested to see whether this change could facilitate flexible/sustainable use of facets in different ways ...

jcoyne:
@magibney currently Blacklight uses an approach where we have a number of templates that depend on a number of helper methods.  These methods are more or less all in one namespace and it's very hard to understand exactly what the dependencies are between views and helpers and between helper methods and other helper methods.  I believe that a component based architecture addresses these shortcomings, so methods are packaged in relation to the views that need them.  This PR uses actionview-component (https://github.com/github/actionview-component) which will be shipped as part of Rails 6.1.

magibney:
Thanks, @jcoyne. So would it be accurate to say that this change is about restructuring to support a more modular, more clearly-scoped architecture? Put another way: there's no concrete _functional_ motivating use case beyond "organize code in a more rational way"? I certainly support that goal, and have had my share of confusion arising from the loose namespacing that you mention. 

Do you foresee there being implications for extensibility/flexibility in dealing with facets, or it's really just a question of shuffling code around to make sure that existing functionality is placed in the proper scope?

jcoyne:
@magibney yeah, there are no functional changes here, it's just an organization change.  That said, it would have implications on what adopters may have done to override views and helper methods, so this would probably target a new major version.

cdmo:
I saw this concept via the [RailsConf 2019 - Rethinking the View Layer with Components by Joel Hawksley](https://www.youtube.com/watch?v=y5Z5a6QdA-M] not too long ago, it looks really interesting. Treating Views as "first class citizens" in the Rails world instead of a kind of loose end.

dkinzer:
I'm confused about how these components works.  Will they replace views altogether.  Will adopting components require all current overrides to views to be chucked and reworked?

PR #2173, opened 2019-10-09
Update bundler

jcoyne:
I'm hoping this fixes a problem with bundler unable to make a lockfile now that sprockets 4 has been released


jcoyne:
Nope.

PR #1996, opened 2018-10-30
Enable the rendering pipeline to be configured per-instance

jcoyne:

PR #2179, opened 2019-10-10
Revert "Enable the rendering pipeline to be configured per-instance"

bess:
Reverts projectblacklight/blacklight#1996

This breaks the build. I should have re-run things before I merged it, sorry!


jcoyne:
Fixed in #2180

Issue #2184, opened 2019-10-10
Remove Vagrantfile


jcoyne:
Nobody is keeping this updated.

mejackreed:
Fixed in #2243

PR #2186, opened 2019-10-12
Add docker-compose configuration for local development.

dkinzer:
REF: [BL Maintenance breakout](https://docs.google.com/document/d/1CgguDDxQjE9SvPb-yWBc4-94nrIQLxFpRSVC54TEQLI/edit#heading=h.ak6eb06zpomj)

This change  adds a docker-compose.yml and Docker definitions to set up
a local blacklight development environment.

In order to bring up an instance of the project locally  the developer will
need to install docker and then run:

`docker-compose up`

from the root of the project.

The solr instance will be available at http://localhost:8983
The blacklight app instance will be available at http://localhost:3000

Note that the current definition continues to use the enginecart gem for
setting up the instance as this makes sense for development of a rails
engine application.

Also note that this change does not add docker to the internal project that gets generated.

This commit also removes an unused Vagrant file.


.docker/solr/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM solr:7.7
```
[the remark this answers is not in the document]
jcoyne: I'm not sure we need a separate Dockerfile for solr.  I believe we can handle this with docker-compose. e.g.: https://github.com/sul-dlss/argo/blob/master/docker-compose.yml#L50-L56

bin/start.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+bundle install
```
jcoyne: This is already happening in the Docker file.  Why is it necessary again?

bin/start.sh
```diff
@@ -0,0 +1,5 @@
...
+rm tmp/pids/server.pid 2> /dev/null
```
jcoyne: we don't have a tmp/pids directory.  What is this trying to do?

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3.7"
+
+services:
+  app:
+    build:
+      context: .
+      dockerfile: .docker/app/Dockerfile
+    stdin_open: true
```
jcoyne: Why?

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    tty: true
```
jcoyne: Why?

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    volumes:
+      - .:/app
+    depends_on:
+      - solr
+    ports:
+      - "3000:3000"
+    environment:
+      SOLR_URL: "http://solr:8983/solr/blacklight"
+      LC_ALL: "C.UTF-8"
```
jcoyne: I don't understand why this is necessary.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    entrypoint:
+      - /app/bin/start.sh
+
+  solr:
+    build:
+      context: .
+      dockerfile: .docker/solr/Dockerfile
+    image: tulibraries/solr:6.6.1
```
jcoyne: this doesn't seem right.

.docker/app/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ruby:2.6.5 as production
```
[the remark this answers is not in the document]
dkinzer: Yeah, this is confusing because I stopped using the tags  and forgot to remove them.  I'll update that.  Thanks.  I'll also update to use ruby 2.6

.docker/app/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN \
+      wget -qO- https://deb.nodesource.com/setup_9.x | bash - && \
+      wget -qO- https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+      echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+      apt-get update -qq && \
+      apt-get install -y --force-yes --no-install-recommends \
+      nodejs yarn build-essential libpq-dev \
+      vim ctags tmux \
```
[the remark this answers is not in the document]
dkinzer: Sometimes it's necessary to go inside the container and futz around... having and editor to make changes while your inside the container is good.  tmux is just useful if you start to do something complicated inside the container that requires multiple shells.  I'll remove ctags.

.docker/solr/Dockerfile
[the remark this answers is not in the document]
dkinzer: I'm using 7 because that's what were' using in travis.

The separate docker file is in order to create the collection (not sure how to do that from just the docker-compose step).

bin/start.sh
dkinzer: Oops... this should be `.internal_test_app/pids/server.pid`...

docker-compose.yml
dkinzer: Without this you wont be able to do a `binding.pry`  that connects to terminal output (it just wont respond to your keys)

docker-compose.yml
dkinzer: Same as above .

docker-compose.yml
dkinzer: Yup.. this is wrong.

bin/start.sh
dkinzer: I was getting a gem not found error without it... but maybe I just doing something not right.

.docker/app/Dockerfile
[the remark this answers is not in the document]
dkinzer: the app container doesn't contain an editor so if you ever have to go in to edit something on the file (for debugging purposes) you'll have to install an editor.

tmux is there for convenience in the case you have to go into the app container and work with multiple shells.

bin/start.sh
dkinzer: OK, so after removing it just now locally... the error I get without it is 

```
app_1   | Could not find rake-13.0.0 in any of the sources
app_1   | Run `bundle install` to install missing gems.
```

docker-compose.yml
dkinzer: I can't remember why I ever did that so I'm removing it too.

docker-compose.yml
dkinzer: Here is some better reference:  http://www.chris-kelly.net/2016/07/25/debugging-rails-with-pry-within-a-docker-container/

bin/start.sh
dkinzer: Actually, it's not already happening in the dockerfile. And, it's happening here so that it gets called every time we start the container.

cdmo:
Covers #2184

.docker/solr/Dockerfile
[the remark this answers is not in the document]
jcoyne: The lines I linked above show how to do this.

.docker/app/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ruby:2.6
+RUN \
+      wget -qO- https://deb.nodesource.com/setup_9.x | bash - && \
+      wget -qO- https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+      echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+      apt-get update -qq && \
+      apt-get install -y --force-yes --no-install-recommends \
+      nodejs yarn build-essential libpq-dev \
+      vim tmux \
```
jcoyne: Do we need this?

tasks/blacklight.rake
```diff
@@ -42,11 +42,17 @@ namespace :blacklight do
     if File.exist? EngineCart.destination
       within_test_app do
         system "bundle update"
+
+        system "bundle exec rake blacklight:index:seed" if ENV["DEVELOPMENT_ENVIRONMENT"] == "docker"
+
+        system "bundle exec rails s #{args[:rails_server_args]}"
```
jcoyne: This is a duplicate of what is happening on line 61

dkinzer: No, @jcoyne the step inside line 61 is happening within a SolrWrapper block.  We are skipping that block when running `docker-compose up`

.docker/solr/Dockerfile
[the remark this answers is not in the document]
dkinzer: OK, I'll update that.

.docker/app/Dockerfile
dkinzer: vim and tmux?   Yes. We need that in development mode.

dkinzer: i.e. for convenience ... this is just for development mode...

tasks/blacklight.rake
jcoyne: But what about when you aren't in docker-compose?

.docker/app/Dockerfile
jcoyne: I disagree pretty strongly.  Docker should only run a single process.  If you need to edit, you can mount the volume and edit locally.

dkinzer: @jcoyne editing mounted files in the container is not the use case.  It's when you need to edit a file that is not part of the mount . .i.e. one of the gems or any other file in the container you need to access that is not accessible via the mount.

tasks/blacklight.rake
dkinzer: The docker-comopse session sets `DEVELOPMENT_ENVIRONMENT: "docker"` environment variable which changes the behavior of this script so that it skps SolrWrapper (because it's not required) and seeds its per line 46 above.

dkinzer: Actually, I see your point... I'll update it.

.docker/app/Dockerfile
dkinzer: @jcoyne following up on this.  I followed up to your earlier objection with a clarifying statement but I haven't heard back.  I don't know if it's because you still object to the change or because you have not gotten around to it.  I can remove the tmux and vim if it's still objectionable to have them there despite my clarifying statement on my reasoning (I can always add them myself if I ever need them).

jcoyne: Sorry to have dropped this.  Yes, I don't want to have vim and tmux.

mjgiarlo: My experience is similar to @jcoyne's: I haven't added dev tools like vim and tmux to `Dockerfile` in other projects. I might be in favor of moving forward without them, and if that proves annoying going forward, we can revisit this?

dkinzer: Thanks @jcoyne, @mjgiarlo, it's been removed.

mjgiarlo:
I am only intermittently active in the Blacklight community, so I am unsure of the norms around merging PRs. We now have three approvals. Do other folks want to review? Or should we merge this and iterate on improving our usage of docker for local dev & testing? 

cc: @cdmo @jrochkind @cbeer @mejackreed @camillevilla @jkeck

jkeck:
Is there a plan to have this documented elsewhere besides this PR?  If this were to be merged, it would be important, IMO, to ensure the documentation is updated ASAP.

One of the issues from merging Docker support in previously is that work was done, but no (or very little) documentation was updated to reflect the changes.  My hope is that we don't end up in the same scenario here w/o a strong commitment to updating documentation as well.

mjgiarlo:
> Is there a plan to have this documented elsewhere besides this PR? If this were to be merged, it would be important, IMO, to ensure the documentation is updated ASAP.
> 
> One of the issues from merging Docker support in previously is that work was done, but no (or very little) documentation was updated to reflect the changes. My hope is that we don't end up in the same scenario here w/o a strong commitment to updating documentation as well.

A plan? Gosh, I don't know. How does the Blacklight community come together to do planning? 

I'm happy to give this some cycles later in the week, though. Specifically: to pull down the latest in `master`, read through the README, and see if the documentation tells the truth and judge how easy the whole process to run the test suite and get a working instance up and running. As part of that, happy also to put in follow-up documentation PRs.

If we're really not feeling ready to commit to a docker-based approach as a group of committers, I'm OK moving forward with solr_wrapper as long as there's a plan to make that a community-based project to help maintain it.

jkeck:
Well, there hasn't traditionally been a "plan" in this regard AFAIK.  Typically (at least in my experience) folks have included documentation with/or close to introducing changes that necessitated it.

I appreciate your offer to backfill that documentation @mjgiarlo!  I would be happy to help w/ that in any way I can (sanity check the steps w/ you, whatever).

mjgiarlo:
@jkeck cool! I'll be in touch later this week. What do folks think about this PR? Merge? Needs more review?

mjgiarlo:
I chatted with @jkeck earlier today and he supports this moving forward, so it seems we have sufficient support to move this PR forward. Thank you for your work and your patience, @dkinzer! merging now.

cdmo:
Hey @dkinzer did you see this issue? https://github.com/projectblacklight/blacklight/issues/2231

I mostly point this out to help along with what I thought was an important point by @jkeck - adding documentation, like updating the quick start guide or making a separate guide like "How to Contribute Code to Blacklight." [I'm not sure](https://github.com/projectblacklight/blacklight/pull/2243#issuecomment-585896925) something like this has existed (?).

Also, I just cloned master and ran `docker-compose up` and got:

```
➜  blacklight git:(master) docker-compose up
Building app
ERROR: Dockerfile parse error line 9: unknown instruction: &&
```

Did anyone else try this locally? I think there's a backslash missing the end of line 8.

dkinzer:
Thanks @cdmo I hadn't seen that issue.  I made a last minute change to the docker compose PR and I neglected to try it locally before pushing up.  Checking now.

dkinzer:
It's still working for me locally because the images are already built, so I wouldn't have caught the issue.

mjgiarlo:
@cdmo :speech_balloon: 
> I mostly point this out to help along with what I thought was an important point by @jkeck - adding documentation, like updating the quick start guide or making a separate guide like "How to Contribute Code to Blacklight." [I'm not sure](https://github.com/projectblacklight/blacklight/pull/2243#issuecomment-585896925) something like this has existed (?). EDIT: maybe the [contributing guidelines](https://github.com/projectblacklight/blacklight/blob/master/CONTRIBUTING.md) [](/projectblacklight/blacklight/blob/HEAD@{2020-03-13T03:06:56Z}/CONTRIBUTING.md) are the closest thing, and they could just be beefed up? Or the wiki page that the contributing guidelines point to?

In whipping up the docker-compose PR I'm working on now, I thought I would need to update the Quickstart, but having read over it... I'm not sure it needs updating if we're primarily using docker + -compose for local development of Blacklight itself and for CI. Unless folks are wanting Blacklight to ship docker-compose artifacts (instead of solr_wrapper) to new blacklight apps, which I don't think I've heard but I will admit I'm a newcomer to this thread.

> Also, I just cloned master and ran `docker-compose up` and got:
> 
> ```
> ➜  blacklight git:(master) docker-compose up
> Building app
> ERROR: Dockerfile parse error line 9: unknown instruction: &&
> ```
> 
> Did anyone else try this locally? I think there's a backslash missing the end of line 8.

I could patch this in my in-progress PR unless someone else beats me to it.

PR #2204, opened 2019-11-07
Push methods up to the DocumentPresenter

cbeer:


app/presenters/blacklight/document_presenter.rb
```diff
@@ -38,6 +50,22 @@ def display_type(base_name = nil, default: nil)
       display_type
     end
 
+    ##
+    # Render the index field label for a document
```
jcoyne: remove `index` here.

cbeer:
```suggestion
    # Render the field label for a document
```

PR #2211, opened 2019-11-14
Use Event#currentTarget to get the actual link instead of the thing t…

cbeer:
…hat was clicked on.

Fixes https://github.com/sul-dlss/dlme/issues/686

### DIFF
--- a/app/assets/javascripts/blacklight/blacklight.js
+++ b/app/assets/javascripts/blacklight/blacklight.js
@@ -441,7 +441,7 @@ Blacklight.doSearchContextBehavior = function () {
   const nodes = Array.prototype.slice.call(elements);
   nodes.forEach(function (element) {
     element.addEventListener('click', function (e) {
-      Blacklight.handleSearchContextMethod.call(e.target, e);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/blacklight/blacklight.js
```diff
@@ -441,7 +441,7 @@ Blacklight.doSearchContextBehavior = function () {
   const nodes = Array.prototype.slice.call(elements);
   nodes.forEach(function (element) {
     element.addEventListener('click', function (e) {
-      Blacklight.handleSearchContextMethod.call(e.target, e);
+      Blacklight.handleSearchContextMethod.call(e.currentTarget, e);
```
mejackreed: Can we do this here: https://github.com/projectblacklight/blacklight/blob/edca682fcf34c949ade403bc756d5df89b6d7cd4/app/javascript/blacklight/search_context.js#L13 and then use the build thing?

```sh
$ npm run js-compile-bundle
```

cbeer: Ah, thanks. I forgot that's how we're doing things now.

PR #2237, opened 2020-01-31
Move field labels into the field presenter and pass through the value…

cbeer:
… count so i18n could pluralize labels as appropriate


app/presenters/blacklight/document_presenter.rb
```diff
@@ -68,41 +72,28 @@ def thumbnail
...
-      FieldPresenter.new(view_context, document, field_config, options).render
+      field_value(field_config, options)
     end
+    deprecation_deprecate field_values: 'Use #field_value'
 
     def retrieve_values(field_config)
-      FieldRetriever.new(document, field_config).fetch
+      field_presenter(field_config).values
+    end
+    deprecation_deprecate retrieve_values: 'Use FieldPresenter#valuees'
```
jcoyne: typo

PR #2238, opened 2020-02-07
Move "Start Over" link into a partial

jkeck:
… (but default to the current non-html value).

### DIFF
--- a/app/views/catalog/_constraints.html.erb
+++ b/app/views/catalog/_constraints.html.erb
@@ -2,7 +2,7 @@
   <div id="appliedParams" class="clearfix constraints-container">
     <h2 class="sr-only"><%= t('blacklight.search.search_constraints_header') %></h2>
 
-    <%=link_to t('blacklight.search.start_over'), start_over_path, class: "catalog_startOverLink btn btn-primary" %>
+    <%= render 'start_over' %>
     <span class="constraints-label sr-only"><%= t('blacklight.search.filters.title') %></span>
     <%= render_constraints(params) %>
   </div>
--- a/app/views/catalog/_start_over.html.erb
+++ b/app/views/catalog/_start_over.html.erb
@@ -1,0 +1,1 @@
+<%= link_to t('blacklight.search.start_over'), start_over_path, class: "catalog_startOverLink btn btn-primary" %>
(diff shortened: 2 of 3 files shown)


app/views/catalog/_constraints.html.erb
```diff
@@ -2,7 +2,7 @@
   <div id="appliedParams" class="clearfix constraints-container">
     <h2 class="sr-only"><%= t('blacklight.search.search_constraints_header') %></h2>
 
-    <%=link_to t('blacklight.search.start_over'), start_over_path, class: "catalog_startOverLink btn btn-primary" %>
+    <%=link_to t('blacklight.search.start_over_html', text: t('blacklight.search.start_over'), default: :'blacklight.search.start_over'), start_over_path, class: "catalog_startOverLink btn btn-primary" %>
```
jkeck: This doesn't seem like an ideal way to do this, but not sure w/ our i18n-tasks setup how to accomplish this sort of thing.

I could also switch all the locales to use `_html` and then `default` back to the original key (in case it was overridden downstream), but wasn't sure which way to go.

jcoyne: Another approach would be to put "start over" in a partial or helper, and then it can be customized as necessary.  How would that approach be better or worse than the one you have proposed here?

jkeck: I had a vague recollection of this being in a partial previously (but I could be completely making that up).  I think that would be a reasonable approach here if folks found it acceptable.  I'll update this PR to include an approach w/ a partial shortly.

mejackreed: I'd be good with this `*_html` approach, as I think we are already using that pattern in other places. No strong feelings thought, but seems like its good to be able to include additional html in these types of locales.

jcoyne: I think that if you're putting html in the locale file, you're really stretching the purpose of i18n.

jkeck: I think exploring the partial option is a decent approach, at least from the backwards compatibility aspect (and the `_html` i18n key that effectively references it's default, and is a default makes me feel particularly uncomfortable about what I've done here).

Plainly switching over to `_html` would be the easiest approach (and i18n supports this by returning `html_safe` content when you do that, so at least there is support upstream), however this is going to be a PITA for any downstream consumer.  This was my attempt to get around that, but the complexity of the solution feels like it outweighs the benefits.

Issue #2231, opened 2020-01-21
Blacklight development flow; running test suite 


ewlarson:
Hi everyone,

I have a pretty simple, but important, PR for GeoBlacklight we'd like to submit upstream to Blacklight instead: https://github.com/geoblacklight/geoblacklight/pull/865

However! I don't know how to run Blacklight locally anymore. Am I to launch a docker instance of Solr? Also, I can't run the test suite via `bundle exec rake ci` now. How has that changed? The old ci task is still in master... Feeling lost here. Any new documentation I should read?

Once I'm back up and running, I can submit a PR. Thanks for the help!

cdmo:
Here's my experience for what it's worth because I just had to do some of this to get a PR into blacklight_advanced_search.

```sh
git clone git@github.com:projectblacklight/blacklight.git
bundle
bundle exec rake
```

At this point there are errors related to solr not being there. I don't have the time at the moment to see the cause of that (also I rarely use solr_wrapper except for working on blacklight core gems), but a workaround is to go into the `.internal_test_app` directory and run `solr_wrapper`. Then go back and run `bundle exec rake` from the main blacklight directory.

The seed then runs ok and it finds solr. However, locally I have a number of failing specs related to sprockets 4.0.0 that all look like (see https://github.com/projectblacklight/blacklight_advanced_search/pull/109#issuecomment-574712087 where I hit this there too):

```
  71) Search Page shows vernacular (Linked 880) and call number
      Failure/Error: //= link_tree ../images

      ActionView::Template::Error:
        link_tree argument must be a directory
      # ./.internal_test_app/app/assets/config/manifest.js:1
      # ------------------
      # --- Caused by: ---
      # Sprockets::ArgumentError:
      #   link_tree argument must be a directory
      #   ./.internal_test_app/app/assets/config/manifest.js:1
```

I'm not sure exactly why this is happening locally for me, but I have a feeling it's related to my version of ruby locally. I know that Travis CI doesn't encounter these errors.

```
➜  blacklight git:(master) ruby --version
ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin18]
```

mejackreed:
Related to the changes in https://github.com/projectblacklight/blacklight/pull/2012#issuecomment-540750149

I think the @cdmo work around should work.

1. `cd` to `.internal_test_app`
1. Run `solr_wrapper` (if you have solr_wrapper installed globally)
1. run `rake` from top level directory

[...]

PR #2243, opened 2020-02-13
Removes Vagrantfile

cdmo:
This has been talked about a number of times, I'm proposing we remove the `Vagrantfile` which is old, outdated and unsupported.


jronallo:
What's the current recommendation for getting set up as a developer for doing development on blacklight?

mejackreed:
I've found that Go Rails usually covers everything we need https://gorails.com/setup . Maybe something like that is worth linking in the wiki?

mejackreed:
@cdmo would you mind also 🔥 https://github.com/projectblacklight/blacklight/blob/master/provision.sh ?

cdmo:
Great catch @mejackreed, thanks.

@jronallo I don't think there is a "current recommendation for getting set up as a developer for doing development on blacklight." Was there at one point? In #2231 this was brought up to some degree, but a full on: _if you are developing a feature for project blacklight use X_ doesn't seem to exist afaik.

cdmo:
@mejackreed I updated the `.gitignore` too

PR #2253, opened 2020-03-03
Provide an icon label context into Blacklight::Icon

### LINKED ISSUE #2248: Future Blacklight::Icon considerations
With the `<title>` was added to the SVG (w/ a ID) in #2157 a potential unintended consequence is it's now relatively easy to add duplicate IDs to the page by having an icon show up more than once.  (We're currently seeing this in Spotlight where we the search icon can show up more than once on a page).  There are currently ways to get around this (e.g. don't use the `blacklight_icon` helper which caches the icon markup), but thought I would file this ticket in case any future work happens in `Blacklight::Icon` land that could meet these use cases.

Also, in this case it might be nice to be able to supply different labeling to this same icon (w/o having to physically duplicate the svg file in your repo).

mejackreed:
Fixes #2248. This allows someone to more surgically change the i18n
labels for icon usage in different places throughout a Blacklight
application.

Inspired by need in https://github.com/sul-dlss/exhibits/issues/1774

Additionally, `blacklight_icon` takes an `options` argument which is used as a cache key. This could be useful for someone trying to generate different title/id attributes used by icons in multiple places.

### DIFF
--- a/app/models/blacklight/icon.rb
+++ b/app/models/blacklight/icon.rb
@@ -2,7 +2,7 @@
 
 module Blacklight
   class Icon
-    attr_reader :icon_name, :aria_hidden, :label, :role
+    attr_reader :icon_name, :aria_hidden, :label, :role, :additional_options
     ##
     # @param [String, Symbol] icon_name
     # @param [Hash] options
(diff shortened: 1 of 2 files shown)


app/models/blacklight/icon.rb
```diff
@@ -2,20 +2,22 @@
...
+    attr_reader :icon_name, :aria_hidden, :label, :role, :additional_options
     ##
     # @param [String, Symbol] icon_name
     # @param [Hash] options
     # @param [String] classes additional classes separated by a string
     # @param [Boolean] aria_hidden include aria_hidden attribute
     # @param [Boolean] label include <title> and aria-label as part of svg
     # @param [String] role role attribute to be included in svg
-    def initialize(icon_name, classes: '', aria_hidden: false, label: true, role: 'img')
+    # @param [Hash] additional_options the way forward instead of named arguments
```
mejackreed: The future intent here so we don't keep adding named args

spec/models/blacklight/icon_spec.rb
```diff
@@ -35,6 +35,14 @@
           .not_to have_css 'svg[aria-labelledby^="bl-icon-search-"]'
       end
     end
+    context ' with a label context' do
+      subject { described_class.new(:search, classes: 'awesome', aria_hidden: true, additional_options: { label_context: 'foo' }) }
+      it 'adds title' do
+        puts subject.svg
```
jcoyne: Cruft?

mejackreed: Nice catch.. 🔥

PR #2254, opened 2020-03-03
Update travis build matrix

cbeer:


.travis.yml
```diff
@@ -36,7 +35,7 @@ matrix:
 
 before_install:
   - gem update --system
-  - gem install bundler
+  - gem install bundler -v 1.17
```
jcoyne: why do we need this version? Can you add a comment explaining?

PR #2260, opened 2020-03-13
Adds probot/stale configuration to repository for stale issues.

cdmo:
I will not be disappointed if this is rejected/not merged (not sure it should really!), but, I noticed stalebot messages in a repo I use ([more on that here](https://www.jeffgeerling.com/blog/2020/enabling-stale-issue-bot-on-my-github-repositories)) and thought about how it might be useful for blacklight. With stalebot stale issues are automatically closed after a period of inactivity. Just something to consider maybe.  https://probot.github.io/apps/stale/

If there is interest, I do think it might be worth adding a bit more to the explanation messaging

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,52 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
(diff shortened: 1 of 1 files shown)


mejackreed:
I'm personally ok with leaving issues opened around for a while, but I recognize that others like issues a bit more trimmed up.

It might be a little jarring to some so if we go this route we just need to communicate what the issue closing means.

cdmo:
It does leave a 30 day grace period after the initial 90 days of inactivity. Having so many opened issues and PRs kinda makes for a lot of "noise." But yeah, it'd be a big change for sure.

cdmo:
Ironically this is getting stale :)

PR #2263, opened 2020-03-13
Use docker-compose to run Solr in CI

mjgiarlo:
Fixes #2195
Fixes #2231 
Closes #2256 

# Summary

In this way, we are using docker-compose for development and testing of the Blacklight gem itself, not for downstream purposes such as generation of new Blacklight applications, which will continue to use solr_wrapper. Also includes: using a lighter base image for the app.

This supersedes #2256.

# Open Question (Now Closed)

I put the documentation here: https://github.com/projectblacklight/blacklight/wiki/Testing-and-Developing-Blacklight

~~It is not clear to me where information like this is currently documented. The README is short and sweet, and for this reason I suspect it is not a good place for such internal development documentation to live. Other options include `CONTRIBUTING.md`, but that is more focused on higher-level contribution process than particular technical steps.~~

~~Note, I also don't think the Quickstart is the right place, so this ticket has nothing to do with #2183, which I think we should either close as a `wontfix` per [my comment](https://github.com/projectblacklight/blacklight/issues/2183#issuecomment-598779218) or discuss what the right way to move forward is particularly w/r/t how much we continue to generate for downstream BL apps: *i.e.*, if we don't give them a working SolrWrapper, do we generate/ship a working docker-compose configuration? Or do we get out of that business entirely? This is beyond the scope of this PR.~~

~~For this reason, perhaps the best place is somewhere on the [project wiki](https://github.com/projectblacklight/blacklight/wiki). Happy to take suggestions on this.~~

# Webpack(er) Broken (But Not Related)

I believe this may be solved by https://github.com/projectblacklight/blacklight/pull/2266. Thanks, @mejackreed!

[...]

PR #2267, opened 2020-03-28
Generate Rails 6 applications with Sprockets 4-style assets

mjgiarlo:
Blacklight is currently generating Rails 6 applications with Sprockets 3-style assets. With this change, the `blacklight:server` task spins up a working server (with Sprockets 4-style assets) instead of screaming about webpacker.

Related to #2266, discovered in #2263.

### DIFF
--- a/lib/generators/blacklight/assets_generator.rb
+++ b/lib/generators/blacklight/assets_generator.rb
@@ -22,6 +22,16 @@ module Blacklight
       end
     end
 
+    ##
+    # Remove the empty generated app/assets/images directory. Without doing this,
+    # the default Sprockets 4 manifest will raise an exception.
+    def appease_sprockets4
+      return if !defined?(Sprockets::VERSION) || Sprockets::VERSION < '4'
+
+      append_to_file 'app/assets/config/manifest.js', "\n//= link application.js"
+      empty_directory 'app/assets/images'
+    end
+
     def assets
       copy_file "blacklight.scss", "app/assets/stylesheets/blacklight.scss"
 
--- a/tasks/blacklight.rake
+++ b/tasks/blacklight.rake
@@ -40,7 +40,7 @@ namespace :blacklight do
   end
 
   namespace :internal do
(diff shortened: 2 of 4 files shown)


tasks/blacklight.rake
```diff
@@ -30,7 +30,10 @@ namespace :blacklight do
   end
 
   namespace :internal do
-    task seed: ['engine_cart:generate'] do
+    desc 'Index seed data in test ap'
+    task :seed do
+      Rake::Task['engine_cart:generate'].invoke unless File.exist?(EngineCart.destination)
```
mjgiarlo: This allows repeated runs of the blacklight server task with the same generated app.

tasks/blacklight.rake
```diff
@@ -30,7 +30,10 @@ namespace :blacklight do
   end
 
   namespace :internal do
-    task seed: ['engine_cart:generate'] do
+    desc 'Index seed data in test ap'
```
jcoyne: typo

tasks/blacklight.rake
jcoyne: Doesn't engine cart already check that the app is the same and do nothing?

lib/generators/blacklight/assets_generator.rb
```diff
@@ -20,6 +20,15 @@ def create_sprockets_javascript
           //= require turbolinks
         CONTENT
       end
+
+      append_to_file 'app/assets/config/manifest.js', "\n//= link application.js"
```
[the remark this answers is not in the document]
mjgiarlo: I'm going to add a version check for Sprockets. Coming in a moment.

tasks/blacklight.rake
mjgiarlo: good eye, thx! fixed

tasks/blacklight.rake
mjgiarlo: It should. I saw it regenerate once, but perhaps things were in a weird state. Removing.

PR #2268, opened 2020-03-30
Extract facet presenters and components

cbeer:
An alternative to #2168. This PR adds view components for facet fields and facet items, and updates views and helpers to call into the components while preserving backwards compatibility for downstream applications which may have overridden the helpers.

This PR also allows downstream applications to use custom components using a couple new field configuration parameters:

```
config.add_facet_field 'whatever', component: MyCustomFacetFieldComponent
``` 

```
config.add_facet_field 'whatever', item_component: MyCustomItemRenderingComponent
``` 

Most of the added code here is to preserve backwards compatibility and can be removed or reshuffled as part of a major release. I've also marked some methods with the `@deprecated` yard tag that I expect to go away as part of another major release (or we can decide to just drop for 8.0), and with `@private` for methods that are only exposed publicly for backwards compatibility.


app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 module Blacklight::FacetsHelperBehavior
```
cbeer, line 2: It's a little hard to tell from the diff, but as a result of this PR, the only non-deprecated helpers are:

- `has_facet_values?` (probably not long for this world)
- `render_facet_partials` (perhaps part of a future component)
- `render_facet_limit` (possibly obsolete once components get traction)
- `facet_field_in_params`  (a little tricky to deprecate at the moment)
- `facet_field_presenter`

PR #2270, opened 2020-04-03
Start migrating constraints towards using view components

cbeer:


app/helpers/blacklight/render_constraints_helper_behavior.rb
```diff
@@ -47,54 +65,53 @@ def render_constraints_query(localized_params = params)
   #
   # @param [ActionController::Parameters] localized_params query parameters
   # @return [String]
-  def remove_constraint_url(localized_params)
-    scope = localized_params.delete(:route_set) || self
```
cbeer, line 51: I /think/ this behavior is unnecessary with the switch to `search_action_path`  (probably mirroring what's going on with the facets anyway).

app/helpers/blacklight/render_constraints_helper_behavior.rb
```diff
@@ -6,39 +6,57 @@
...
   end
 
   ##
   # Render the query constraints
   #
   # @param [ActionController::Parameters] localized_params query parameters
   # @return [String]
-  def render_constraints_query(localized_params = params)
+  def render_constraints_query(params_or_search_state = search_state)
+    search_state = convert_to_search_state(params_or_search_state)
```
dkinzer, line 47: Just curious, why are we doing this now?  Thanks!

cbeer: `SearchState` is nicer for applications to extend and much nicer to consume by creating a consistent interface for Blacklight to work against. It also fixes some problems we currently have where multiple plugins need to override the same methods to inject some custom parameter parsing (adv search + range limit come to mind, but it's generally true of plugins that want to have custom facet behaviors)

PR #2271, opened 2020-04-08
Safe navigate when iterating over the documents ivar

jkeck:
While we guard against no documents, the meta-programmed action definition + rails allowing undefined actions when a view is present makes this view/partial routable and throw an error.

This is likely an edge case, but we've noticed this triggering errors (primarily caused by bots/crawlers) on sites where we've removed the citation tool.


cbeer:
Should we do the same thing in `sms_form` and `email_form`?

PR #2282, opened 2020-05-11
Style bookmark actions as buttons, closes #2272

### LINKED ISSUE #2272: Bookmark actions should be buttons, not nav pills
The bookmarks page has a few actions that users can take on their bookmarked items, like citing and emailing them.  To reproduce:

1. Go to https://demo.projectblacklight.org and search
2. Add a few items to your bookmarks
3. Go to https://demo.projectblacklight.org/bookmarks and look for the Email and Cite links

This can be found in the views here: https://github.com/projectblacklight/blacklight/blob/8b9041544553ce0dd0d979f9f39f78d94b30b46d/app/views/bookmarks/_tools.html.erb#L1

Since these are actions that a user takes, rather than navigation elements, I'd argue that they should be buttons, rather than nav pills.  Additionally, on a stock blacklight instance, the nav pills are just styled as links, making them easy to miss.

If others agree that they should be buttons, rather than nav-pill links, I'd be happy to open a pull request.

sandbergja:
Since these represent actions that a user would take on their bookmarked
items, let's style them as buttons rather than nav-pills.

These actions are still wrapped in a .bookmarksTools class, just in case
anybody was using this to style their app.


jcoyne:
Can you add a before and after screenshot that shows the difference?

sandbergja:
Yeah!  Here's without my commit:

before

And with my commit:

after

app/views/bookmarks/_tools.html.erb
```diff
@@ -1,7 +1,7 @@
-<ul class="<%= controller_name %>Tools nav nav-pills">
+<div class="<%= controller_name %>Tools">
   <%= render_show_doc_actions document_list, document: nil, document_list: @response.documents, url_opts: Blacklight::Parameters.sanitize(params.to_unsafe_h) do |config, inner| %>
-    <li class="nav-item">
+    <button class="btn btn-outline-primary">
       <%= inner %>
```
jcoyne, line 4: isn't inner going to be links?  I'm not sure you'd want a link in a button.

sandbergja: You are totally right.  I force-pushed a change to my branch to avoid this.

sandbergja:
Just force-pushed a commit that uses components, rather than the (now out-of-date) original approach.

jcoyne (approved):

PR #2285, opened 2020-05-20
Moves comment

cdmo:
This looks to me like a description for a method that was removed. Is it? Not sure what this could be referring to that returns an array of hashes with these special symbols because I only see the one method in this module that does not do that. If it is meant to be here, it does kinda end in a cliff hanger :)

### DIFF
--- a/app/models/concerns/blacklight/document/extensions.rb
+++ b/app/models/concerns/blacklight/document/extensions.rb
@@ -32,6 +32,9 @@ module Blacklight::Document::Extensions
   module ClassMethods
     attr_writer :registered_extensions
 
+    # Returns array of hashes of registered extensions. Each hash
+    # has a :module_obj key and a :condition_proc key. Usually this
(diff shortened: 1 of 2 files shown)


jcoyne:
It was originally added here: https://github.com/projectblacklight/blacklight/commit/1167c468adfce01ccb9811c37ec75f8d17d39e01#diff-69b25eaf076d23e3438a3ad4211020acR181 

So it probably should be moved to here: https://github.com/projectblacklight/blacklight/blob/8b9041544553ce0dd0d979f9f39f78d94b30b46d/app/models/concerns/blacklight/document/extensions.rb#L33

cdmo:
Thanks, that seems to fit there for sure

jcoyne:
The rubocop has apprehended you @cdmo 👮 🚨

cdmo:
One of these days rubo's gonna lock me up and take away my editor

PR #2286, opened 2020-05-22
Show views aren't dependent on index_view_type

jcoyne:
This optimizes show template lookup speed as we ignore two patterns that are never used


jcoyne:
I believe this is used by catalog/_document.html.erb, so it is part of the index chain.

PR #2288, opened 2020-05-22
Avoid deprecated method query_has_constraints?

jcoyne:


cbeer:
We only deprecated the helper in 7.8. Until 8.x, I think we're better off silencing the deprecation from the partial in case someone has overridden the helper?

jcoyne:
I'm not sure why overriding the helper would matter. The old method still remains, this just keeps us from using it.

cbeer:
@jcoyne , sure, but if someone has overridden `query_has_constraints?` in their app, the new behavior would end up bypassing it. Until Blacklight 8, I think https://github.com/projectblacklight/blacklight/pull/2294 is the best we can do, while maintaining backwards compatibility?

jcoyne:
@cbeer Ah, right. Thanks.

barmintor:
This is obsolete per https://github.com/projectblacklight/blacklight/pull/2458

PR #2293, opened 2020-06-03
Use deep_transform_values_in_object (and backport it for Rails 5) to get a clean copy of the config

cbeer:


lib/blacklight/configuration.rb
```diff
@@ -402,5 +401,22 @@ def add_action(config_hash, name, opts)
       yield(config) if block_given?
       config_hash[name] = config
     end
+
+    def deep_transform_values_in_object(object, &block)
+      return object.deep_transform_values(&block) if object.respond_to?(:deep_transform_values)
+      _deep_transform_values_in_object(object, &block)
+    end
+
+    def _deep_transform_values_in_object(object, &block)
```
jcoyne, line 410: Can you drop a comment here about being able to remove this when we drop Rails 5 support?

lib/blacklight/configuration.rb
```diff
@@ -282,15 +282,14 @@ def add_field_configuration_to_solr_request!(*fields)
     # Provide a 'deep copy' of Blacklight::Configuration that can be modified without effecting
     # the original Blacklight::Configuration instance.
     #
-    # Rails 4.x provides `#deep_dup`, but it aggressively `#dup`'s class names
-    # too. These model names should not be `#dup`'ed or we might break ActiveModel::Naming.
+    # Note: Rails 4.x provides `#deep_dup`, but it aggressively `#dup`'s class names too, turning them
```
jcoyne, line 285:
```suggestion
    # Note: Rails provides `#deep_dup`, but it aggressively `#dup`'s class names too, turning them
```

PR #2296, opened 2020-06-04
Convert pivot facets to a component and add hierarchy expand/collapse

cbeer:
Screen Shot 2020-06-04 at 10 01 15

### DIFF
--- a/app/components/blacklight/facet_item_pivot_component.rb
+++ b/app/components/blacklight/facet_item_pivot_component.rb
@@ -3,22 +3,41 @@
 module Blacklight
   # Render facet items and any subtree
   class FacetItemPivotComponent < ::ViewComponent::Base
+    # Somewhat arbitrary number; the only important thing is that
+    # it is bigger than the number of leaf nodes in any collapsing
+    # pivot facet on the page.
+    ID_COUNTER_MAX = 2**20 - 1
+
+    # Mint a (sufficiently) unique identifier, so we can associate
+    # the expand/collapse control with labels
+    def self.mint_id
+      @id_counter = ((@id_counter || 0) + 1) % ID_COUNTER_MAX
+
+      # We convert the ID to hex for markup compactness
+      @id_counter.to_s(16)
+    end
+
     with_collection_parameter :facet_item
 
-    def initialize(facet_item:, wrapping_element: 'li', suppress_link: false)
+    def initialize(facet_item:, wrapping_element: 'li', suppress_link: false, collapsing: nil)
       @facet_item = facet_item
       @wrapping_element = wrapping_element
       @suppress_link = suppress_link
+      @collapsing = collapsing.nil? ? facet_item.facet_config.collapsing : collapsing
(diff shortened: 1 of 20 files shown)


app/components/blacklight/facet_item_pivot_component.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def has_items?
+      @facet_item.items.present?
+    end
+
+    def facet_toggle_button
+      content_tag 'button', class: 'btn facet-toggle-handle collapsed',
+                            data: { toggle: 'collapse', target: "##{@id}" },
+                            aria: { expanded: false, controls: @id, describedby: "#{@id}_label" } do
+        concat content_tag 'span', t('blacklight.search.facets.pivot.show_html'), class: 'show'
```
jcoyne, line 50: I'm really not a big fan of putting html in i18n. Is there a reason we can't put just the text of the  .sr-only element in the i18n file?

app/components/blacklight/facet_item_pivot_component.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module Blacklight
+  # Render facet items and any subtree
+  class FacetItemPivotComponent < ::ViewComponent::Base
+    def self.id_counter
```
jcoyne, line 6: Can you explain this counter? Does this just go up without bounds? Why do we need it?

config/locales/blacklight.de.yml
```diff
@@ -194,6 +194,9 @@ de:
         selected:
           remove: '[entfernen]'
         missing: [fehlt]
+        pivot:
+          show:
+          hide_html: ⊟ <span class="sr-only"></span>
```
jcoyne, line 199: Need to fix this one?

app/components/blacklight/facet_item_pivot_component.rb
cbeer: ARIA requires a unique ID to match the control with the thing it controls. I'm open to better ideas (blacklight-hierarchy uses `SecureRandom.uuid`, which is slow and.. unnecessary). 

Even at 1,000,000,000,000 cycles, the base 36 string is just 7 characters long, but I think most production servers will recycle threads well before that.

config/locales/blacklight.de.yml
cbeer, line 199:
```suggestion
          hide:
```

app/components/blacklight/facet_item_pivot_component.rb
jcoyne: That's super helpful context. Can you please add this information to a comment in this class?

cbeer: Sure; and I've gone ahead and capped the counter at ~1 million, and made sure we only add the id  to leaf nodes. If you have 1 million leaf nodes on  a  page, you might have a problem, but it probably isn't markup validity, accessibility or wonky javascript behavior.

PR #2299, opened 2020-06-18
Update blacklight icon accessibility

cbeer:
Part of https://github.com/sul-dlss/earthworks/issues/653

Fixes https://github.com/sul-dlss/earthworks/issues/657


mejackreed:
More context here, we are now moving from Pattern 5 -> Pattern 8 here: https://www.deque.com/blog/creating-accessible-svgs/. So while the underlying implementation is a bit different, the end result for screen readers should be the same or enhanced.

PR #2301, opened 2020-06-18
Add an extra aria-label to icons

### LINKED ISSUE #2300: Add aria-label to Blacklight::Icon
Previously we used `aria-labelledby` but the duplicate ids caused invalid HTML. Removed in https://github.com/projectblacklight/blacklight/pull/2157 This proposal is to add `aria-label` in its place now.

cbeer:
Fixes #2300

### DIFF
--- a/app/models/blacklight/icon.rb
+++ b/app/models/blacklight/icon.rb
@@ -40,7 +40,7 @@ module Blacklight
       {
         class: classes,
         "aria-hidden": (true if aria_hidden),
-        "aria-label": icon_label
+        "aria-label": icon_label if label
       }
     end
 
--- a/app/models/blacklight/icon.rb
+++ b/app/models/blacklight/icon.rb
@@ -40,7 +40,7 @@ module Blacklight
       {
         class: classes,
         "aria-hidden": (true if aria_hidden),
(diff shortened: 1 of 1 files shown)


app/models/blacklight/icon.rb
```diff
@@ -39,7 +39,8 @@ def icon_label
     def options
       {
         class: classes,
-        "aria-hidden": (true if aria_hidden)
+        "aria-hidden": (true if aria_hidden),
+        "aria-label": icon_label
```
mejackreed, line 43: Can we add the `if label` conditional to this please?

cbeer, line 43:
```suggestion
        "aria-label": icon_label if label
```

app/models/blacklight/icon.rb
```diff
@@ -39,7 +39,8 @@ def icon_label
     def options
       {
         class: classes,
-        "aria-hidden": (true if aria_hidden)
+        "aria-hidden": (true if aria_hidden),
+        "aria-label": icon_label if label
```
cbeer, line 43:
```suggestion
        "aria-label": (icon_label if label)
```

PR #2304, opened 2020-06-25
Inform the IndexPresenter about the search offset

### LINKED ISSUE #2303: document_counter_with_offset raises depredcation warning
Called from `app/views/catalog/_thumbnail.html.erb`
```
(byebug) document_counter_with_offset(document_counter)
DEPRECATION WARNING: The @response instance variable is deprecated; use @document.response instead. (called from ___sers_jcoyne____rbenv_versions_______lib_ruby_gems_______gems_blacklight_______app_views_catalog__thumbnail_html_erb__3328884195611251885_70342959936160 at (byebug):1)
```

jcoyne:
This means our document partials don't have to access instance variables, which means that
we are able to call the thumbnail partial in a show context without getting a deprecation warning about accessing the deprecated @response ivar.

Fixes #2303

### DIFF
--- a/app/presenters/blacklight/document_presenter.rb
+++ b/app/presenters/blacklight/document_presenter.rb
@@ -11,10 +11,12 @@ module Blacklight
     # @param [SolrDocument] document
     # @param [ActionView::Base] view_context scope for linking and generating urls
     # @param [Blacklight::Configuration] configuration
-    def initialize(document, view_context, configuration = view_context.blacklight_config)
+    # @param [Integer] counter what offset in the search result is this record (used for tracking)
+    def initialize(document, view_context, configuration = view_context.blacklight_config, counter: nil)
       @document = document
       @view_context = view_context
       @configuration = configuration
+      @counter = counter
     end
 
     # @return [Hash<String,Configuration::Field>]  all the fields for this index view that should be rendered
@@ -102,7 +104,7 @@ module Blacklight
     end
 
     def thumbnail
-      @thumbnail ||= thumbnail_presenter_class.new(document, view_context, view_config)
+      @thumbnail ||= thumbnail_presenter_class.new(document, view_context, view_config, counter: @counter)
     end
 
     ##
--- a/app/views/catalog/_document.html.erb
+++ b/app/views/catalog/_document.html.erb
@@ -1,6 +1,7 @@
 <% # container for a single doc -%>
 <% view_config = local_assigns[:view_config] || blacklight_config.view_config(document_index_view_type) %>
-<%= render view_config.document_component.new(presenter: document_presenter(document), counter: document_counter_with_offset(document_counter)) do |component| %>
+<% counter = document_counter_with_offset(document_counter) %>
+<%= render view_config.document_component.new(presenter: document_presenter(document, counter: counter), counter: counter) do |component| %>
   <% view_config.partials.each do |partial| %>
     <% component.partial do %>
       <%= render_document_partial document, partial, component: component, document_counter: document_counter %>
--- a/app/helpers/blacklight/blacklight_helper_behavior.rb
+++ b/app/helpers/blacklight/blacklight_helper_behavior.rb
@@ -124,8 +124,8 @@ module Blacklight::BlacklightHelperBehavior
 
   ##
   # Returns a document presenter for the given document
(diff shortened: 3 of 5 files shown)


app/views/catalog/_thumbnail.html.erb
```diff
@@ -1,4 +1,7 @@
-<% if presenter(document).thumbnail.exists? && tn = presenter(document).thumbnail.thumbnail_tag({ alt: '' }, 'aria-hidden': true, tabindex: -1, counter: document_counter_with_offset(document_counter)) %>
+<% if presenter.thumbnail.exists? &&
```
cbeer, line 1: For backwards compatibility, maybe we should provide the old behavior if the presenter local isn't available?

app/presenters/blacklight/document_presenter.rb
```diff
@@ -11,10 +11,12 @@ class DocumentPresenter
     # @param [SolrDocument] document
     # @param [ActionView::Base] view_context scope for linking and generating urls
     # @param [Blacklight::Configuration] configuration
-    def initialize(document, view_context, configuration = view_context.blacklight_config)
+    # @param [Integer] counter what offset in the search result is this record (used for tracking)
+    def initialize(document, view_context, configuration = view_context.blacklight_config, counter: nil)
```
cbeer, line 15: I guess this is ok for backwards-compatibility? Maybe it'd be better to check arity in the helpers? Or just call it out in the release notes, because there are at least a couple projects on github that overrode `initialize`..

jcoyne: Because this is an optional argument, the arity hasn't changed.

app/views/catalog/_thumbnail.html.erb
jcoyne: Good catch. I've added https://github.com/projectblacklight/blacklight/pull/2304/files#diff-1228e931859595c68c525cca60385f73R1-R3 to take care of this.

app/presenters/blacklight/document_presenter.rb
jcoyne: Yes, that seems to work fine.

app/views/catalog/_document.html.erb
```diff
@@ -2,5 +2,6 @@
 <article class="document <%= render_document_class document %> document-position-<%= document_counter%> " data-document-counter="<%= document_counter %>" itemscope itemtype="<%= document.itemtype %>">
   <%= render_document_partials document,
     blacklight_config.view_config(document_index_view_type).partials,
-    document_counter: document_counter %>
+    document_counter: document_counter,
+    presenter: presenter(document, document_counter: document_counter) %>
```
barmintor, line 6:
```suggestion
    presenter: presenter(document, counter: document_counter) %>
```

app/presenters/blacklight/document_presenter.rb
barmintor: Now that we're developing for 8.0 on main: Should `configuration` also be a keyword param? Something about the mix of optional positional args and optional keyword args seems fishy to me.

jcoyne: @barmintor I think that is fine, but does that belong in this PR or should that be a separate change?

barmintor: that's a good point

app/views/catalog/_document.html.erb
barmintor: I should revisit this in after a rebase, but the keyword param on the presenter class is `counter`, and we probably also want to use the same value from the DocumentComponent (which takes the search offset into account, if available). `document_counter` is relative to the current list.

barmintor:
Rebased on main

barmintor (approved):

barmintor:
@jcoyne my impression is that this change is more pertinent to the `release-7.x` branch, so I backported it here: #2702 

If you think it's still valuable on main, it's rebased here and I think it's good to go.

jcoyne:
@barmintor this was two years ago. I don't at all remember the context of this.

barmintor:
@jcoyne maybe we should close?

jcoyne:
Sure, I'm fine with that.

PR #2307, opened 2020-06-30
Allow field and facet presenters to be configured

cbeer:


app/helpers/blacklight/facets_helper_behavior.rb
```diff
@@ -169,7 +169,8 @@ def facet_partial_name(display_facet = nil)
   deprecation_deprecate :facet_partial_name
 
   def facet_field_presenter(facet_config, display_facet)
-    Blacklight::FacetFieldPresenter.new(facet_config, display_facet, self)
+    presenter_class = facet_config.presenter || Blacklight::FacetFieldPresenter
```
jcoyne, line 172: could we set this default in the facet_config itself to avoid the conditional here?

cbeer: 🤷‍♂️ , sure.

app/presenters/blacklight/document_presenter.rb
```diff
@@ -93,7 +93,8 @@ def retrieve_values(field_config)
     deprecation_deprecate retrieve_values: 'Use FieldPresenter#values'
 
     def field_presenter(field_config, options = {})
-      FieldPresenter.new(view_context, document, field_config, options)
+      presenter_class = field_config.presenter || Blacklight::FieldPresenter
```
jcoyne, line 96: That looks great. Can the same be done here?

cbeer: Not really. We don't have a class for just index/show fields.

PR #2308, opened 2020-06-30
Add support for a generic preferred citation type

cbeer:


app/views/catalog/_citation.html.erb
```diff
@@ -8,6 +8,11 @@
   <% @documents&.each do |document| %>
     <h1 class="modal-title"><%= document_heading(document) %></h1>
 
+    <% if document.respond_to?(:export_as_preferred_citation_txt) %>
+      <h2><%= t('blacklight.citation.preferred') %></h2>
+      <%= document.send(:export_as_preferred_citation_txt).html_safe %><br/><br/>
```
jcoyne, line 13: why is the `send` needed?

cbeer: No idea, but it's the pattern we've established. I'm happy to kill it, but maybe it's a 8.0 thing?

jcoyne: I think `document.export_as_preferred_citation_txt.html_safe` would be idea.

barmintor:
It seems like at this point - with the logic from `app/views/catalog/_citation.html.erb` moved into `citation_component.html.erb` and now with a somewhat different implementation, this PR would need to be re-implemented, right?

barmintor:
Since this can no longer be merged, I have ticketed it for further discussion of the feature in #2704

PR #2317, opened 2020-07-31
Convert document show/index views to support and use view components

cbeer:
Note that until the componentizing is a little further along (and we can provide the definitive "template you should use in your own app" file), it's going to be difficult to deprecate the partials that this replaces, but at least this provides implementors with the option to adopt these view components now.

If the implementor does nothing, we use the component for the document wrapper but continue using the configured partials everywhere.

However, if the implementor opts in to the view component approach by setting the `document_component` configuration on `index`, `show`, or specific index view type configurations and and avoid the old partials altogether. At least for now, the content of the partials configuration is passed through to the component (but not rendered by default) to allow for gradual adoption, if desired.


app/helpers/blacklight/blacklight_helper_behavior.rb
```diff
@@ -253,7 +261,7 @@ def with_format(format)
   # Should we render a grouped response (because the response
   # contains a grouped response instead of the normal response)
   def render_grouped_response? response = @response
-    response.grouped?
+    response&.grouped?
```
jcoyne, line 264: why do we have to do a nil check?

app/presenters/blacklight/document_presenter.rb
```diff
@@ -93,6 +99,8 @@ def retrieve_values(field_config)
     deprecation_deprecate retrieve_values: 'Use FieldPresenter#values'
 
     def field_presenter(field_config, options = {})
+      return unless field_config
```
jcoyne, line 102: why is this ever nil?

app/presenters/blacklight/document_presenter.rb
```diff
@@ -63,7 +69,7 @@ def display_type(base_name = nil, default: nil)
     # @param [Hash] options
     # @option options [String] :value
     def field_value field_config, options = {}
-      field_presenter(field_config, options).render
+      field_presenter(field_config, options)&.render
```
jcoyne, line 72: why do we need to do a nil check here?

lib/blacklight/configuration.rb
```diff
@@ -65,6 +65,7 @@ def default_values
           index: ViewConfig::Index.new(
             # document presenter class used by helpers and views
             document_presenter_class: nil,
+            document_component: nil,
```
jcoyne, line 68: could we have the default be `Blacklight::DocumentComponent` and avoid a nil check in the view?

app/views/catalog/_show_main_content.html.erb
```diff
@@ -3,17 +3,21 @@
...
+  <%  end %>
 
-<% if @document.respond_to?(:export_as_openurl_ctx_kev) %>
-  <!--
-       // COinS, for Zotero among others.
-       // This document_partial_name(@document) business is not quite right,
-       // but has been there for a while.
-  -->
-  <span class="Z3988" title="<%= @document.export_as_openurl_ctx_kev(document_partial_name(@document)) %>"></span>
+  <% component.with(blacklight_config.show.document_component.blank? && blacklight_config.view_config(:show).partials.any? ? :body : :partials) do %>
```
jcoyne, line 18: This is a bit confusing. Why do we need this block if `blacklight_config.view_config(:show).partials` is empty?

app/helpers/blacklight/blacklight_helper_behavior.rb
cbeer: Mostly for testing convenience. The awful  mechanics for stubbing `@response` in the view context for testing a  component when... we don't care... seemed fishy.

lib/blacklight/configuration.rb
cbeer: I made it `nil` so we'd have some way to know that an implementation has opted in to the view components. Other ideas?

app/views/catalog/_show_main_content.html.erb
cbeer: Someone might have overridden `render_document_partials`?

app/helpers/blacklight/blacklight_helper_behavior.rb
jcoyne: Can we call that out here in a comment?

lib/blacklight/configuration.rb
jcoyne: Please add a comment that explains that anything other than nil indicates the implementation has opted in to components.

app/views/catalog/_show_main_content.html.erb
jcoyne: Please add a comment to explain this.  Then it will be easier to refactor once `render_document_partials` becomes a component itself.

PR #2319, opened 2020-08-10
More components

cbeer:


app/views/catalog/_show_sidebar.html.erb
```diff
@@ -1,12 +1,2 @@
 <%= render 'show_tools' %>
-
-<% unless document.more_like_this.empty? %>
-  <div class="card">
-    <div class="card-header">More Like This</div>
-    <div class="card-body">
-      <%= render collection: document.more_like_this,
-                 partial: 'show_more_like_this',
```
mejackreed, line 8: Do we leave this partial in for backwards compatibility?

cbeer: Exactly. I suspect it's not long for the world, but might make it into 8.x.

lib/blacklight.rb
```diff
@@ -70,6 +70,16 @@ def self.connection_config=(value)
     Blacklight::RuntimeRegistry.connection_config = value
   end
 
+  def defaults_version
```
jcoyne, line 73: What calls this method?

cbeer: Nothing yet. Just getting it out there so we can start opting in to some of the new behavior without needing a major release.

jcoyne (approved):

PR #2320, opened 2020-08-11
Move search_service_class to Searchable

jcoyne:
Anything that uses Searchable should be able to configure a search_service_class, this does not have anything to do with a Controller.  This keeps the concept of the search service encapsulated in the Searchable module.

### DIFF
--- a/app/controllers/concerns/blacklight/controller.rb
+++ b/app/controllers/concerns/blacklight/controller.rb
@@ -30,12 +30,6 @@ module Blacklight::Controller
     # TODO: move to Searchable
     class_attribute :search_state_class
     self.search_state_class = Blacklight::SearchState
-
-    # Which class to use for the search service. You can subclass SearchService if you
-    # want to override any of the methods (e.g. SearchService#fetch)
-    # TODO: move to Searchable
-    class_attribute :search_service_class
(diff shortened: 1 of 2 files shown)


barmintor:
Is there a reason to wait to do the same thing for the `search_state_class` accessor? It would clean Blacklight::Controller up, and it would be nice to have a single documented external dependency for Blacklight:Searchable. Happy to push a commit doing this if you're pressed for time!

barmintor:
Adding that it doesn't seem to me that there's a _real_ backwards compatibility here - it seems unlikely that an app is using Blacklight::Controller in a context without the Catalog mixin while also using the search_service and search_state machinery, right?

barmintor:
> Is there a reason to wait to do the same thing for the search_state_class accessor?

Yes: the `search_state` accessors are also used by SearchHistoryController, which inherits from ApplicationController and only borrows configuration from CatalogController. That needs to be handled separately.

barmintor:
Rebased before CI.

barmintor (approved):

barmintor:
There's a bootstrap5 accessibility failure here re: text contrast that is not related to the change (and not a required check). I am merging.

jcoyne:
That test is flappy and is often resolve by rerunning the suite

PR #2327, opened 2020-08-27
Commit compiled js needed to release v7.10.0

mejackreed:


cdmo:
Thanks Jack. 

Looks like 161 rubocop offenses. And there's this PR that looks like it's intended to fix this problem https://github.com/projectblacklight/blacklight/pull/2298. That PR needs a rebase though.

mejackreed:
Yes that is what I am doing. 

I think generally most of that is stuff that _should_ happen when creating release. And likely the milestone and email are the things that are being skipped. But I'd say we should strive for it so I would be 👍 for leaving it in.

Issue #2234, opened 2020-01-24
Getting `undefined method `highlight' for nil:NilClass` after upgrade


dkinzer:
After we upgraded from v7.3.0 to v7.4.0, we started to see some errors when rendering dummy/empty Solr documents (which we use in place when an id does not return a document from the server).

We believe we can attribute this error to the following change: https://github.com/projectblacklight/blacklight/compare/v7.3.0..v7.4.0#diff-7142b8c9bd0a0da70267f8ebdf09ae49L24-L28

We can duplicate the error in the console

```
doc = SolrDocument.new({})
index_presenter(doc).heading
```

The above throws the error we are reporting (`undefined method 'highlight' for nil:NilClass`)

However the old code returns an empty string as would be expected
```
document = SolrDocument.new({})
index_presenter(doc).label document_show_link_field(doc), {}
# returns ""
```

## Update:
I looked into this issue further this weekend and I realize now that the issue is more related to blacklight_configurations.  The above result is probably OK in terms  an error as if a document does not have either a title or an id it should error out.

The issue is happening because on our site we have both SolrDocument and PrimoCentralDocument.  PrimoCentralDocument implement the Document interface but they have their own set of field configurations including it's one set of unique_key map.  Part of what happens is that on the bookmark view documents can be any combination of SolrDocument or PrimoCentralDocument types.  And in order to render without errorring the blacklight_config must be able to switch to the proper set per document... i.e. when I render a PrimoCentralDocument bookmark I better be using the primo_central_controller.configuration.

Anyway, prior to this change we were not seeing this error because we were swapping the @blacklight_configuration reference between document renders in the view layer.  However, now that everything is moving to the presenter layer, we need to figure out a way to pass the correct configuration per document render at the presenter level.

This issue is actually related to and a consequence of this issue projectblacklight/blacklight/issues/1878 which I closed but now I feel like that was premature.

I think that at some level since blacklight has a Document Interface it doesn't make sense that bookmark controller assumes that all documents it renders will have the same configuration.  This is a bug in how the controller is implemented.

dkinzer:
Note this issue is also happening in version `v7.5.0`, where we debugged it.

dkinzer:
Following thread further determined that this line returns nil and is part of issue:
```
f = fields.lazy.map { |field| field_config(field) }.detect { |field_config| retrieve_values(field_config).any? }
```

PR #2333, opened 2020-09-23
Fix heading error when document has no fields.

dkinzer:
Fixes projectblacklight/blacklight#2234

### DIFF
--- a/app/presenters/blacklight/document_presenter.rb
+++ b/app/presenters/blacklight/document_presenter.rb
@@ -46,7 +46,7 @@ module Blacklight
 
       fields = Array.wrap(view_config.title_field) + [configuration.document_model.unique_key]
       f = fields.lazy.map { |field| field_config(field) }.detect { |field_config| field_presenter(field_config).any? }
-      field_value(f, except_operations: [Rendering::HelperMethod])
+      f ? field_value(f, except_operations: [Rendering::HelperMethod]) : ""
     end
 
     def display_type(base_name = nil, default: nil)
--- a/spec/presenters/blacklight/show_presenter_spec.rb
+++ b/spec/presenters/blacklight/show_presenter_spec.rb
@@ -143,6 +143,14 @@ RSpec.describe Blacklight::ShowPresenter, api: true do
       config.show.title_field = Blacklight::Configuration::Field.new(field: 'x', values: ->(*_) { 'hardcoded' })
       expect(subject.heading).to eq 'hardcoded'
     end
+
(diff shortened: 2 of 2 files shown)


dkinzer:
I'm not getting any rubocop errors on my local and I'm not touching any of those files so I don't know what's going on there with that failure. :/

jcoyne:
This should fix the problem with rubocop: https://github.com/projectblacklight/blacklight/pull/2334

app/presenters/blacklight/document_presenter.rb
```diff
@@ -69,7 +69,7 @@ def display_type(base_name = nil, default: nil)
     # @param [Hash] options
     # @option options [String] :value
     def field_value field_config, options = {}
-      field_presenter(field_config, options).render
+      field_config ? field_presenter(field_config, options).render : ""
```
jcoyne, line 72: what if this nil check was moved up into `heading` (line 44)?  I'm not sure we should allow calling field_value with a nil field_config.

dkinzer: Yeah that fine.  I think I moved it here merely because I could make a one liner :)

dkinzer: Sure thing 👍

app/presenters/blacklight/document_presenter.rb
```diff
@@ -46,7 +46,7 @@ def heading
 
       fields = Array.wrap(view_config.title_field) + [configuration.document_model.unique_key]
       f = fields.lazy.map { |field| field_config(field) }.detect { |field_config| field_presenter(field_config).any? }
-      field_value(f, except_operations: [Rendering::HelperMethod])
+      f ? field_value(f, except_operations: [Rendering::HelperMethod]) : ""
```
jcoyne, line 49: Does this need to allocate a string or would a nil value (indicating no header) be a better?

dkinzer: @jcoyne , I want it to have same behavior as the old code for `presenter.label` which used to return an empty string in this context.  I haven't tested locally what would happen for the nil case.

I think that in terms of backwards compatibility it `presenter.label` it would thus make sense to continue to return and empty string.

dkinzer: Note that I agree that returning `nil` will be a cleaner solution, and originally I did that. But I decided to play it safely and just keep the old behavior.

jcoyne (approved):

PR #2335, opened 2020-09-25
Do not mutate search_state when mutating search_state.to_h

dkinzer:
`Blacklight::SearchState#to_h` needs to return a dupe of it's internal
state @params not a reference to it.  Otherwise whatever mutation you do
to the reference gets reflected in the `Blacklight::SearchState` instance.

It's unexpected behavior for the `to_h` method to expose the
`Blacklight::SearchState` instance's copy of `@params` and it can lead
to very confusing bugs.

### DIFF
--- a/lib/blacklight/search_state.rb
+++ b/lib/blacklight/search_state.rb
@@ -47,7 +47,7 @@ module Blacklight
     end
 
     def to_hash
-      @params
+      @params.dup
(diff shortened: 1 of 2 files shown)


jcoyne (approved):

jcoyne:
Thanks!

PR #2338, opened 2020-09-29
Return of the solr wrapper

cbeer:
Have the rake task:

- complain if `docker-compose` doesn't work
- fall back on solr_wrapper if docker isn't available

### DIFF
--- a/.solr_wrapper.yml
+++ b/.solr_wrapper.yml
@@ -1,0 +1,5 @@
+# Place any default configuration for solr_wrapper here
+# port: 8983
+collection:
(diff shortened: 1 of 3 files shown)


andrewbattista:
Happy to see the return of solr wrapper =)

dkinzer:
This makes a lot of sense to me.  Basically if docker-compose is available on the system us it (makes docker camp happy), otherwise use solr_wrapper (makes sor_wrapper camp happy).

jkeck (approved):

jkeck:
Going to ship this since there doesn't seem to be any descent.

PR #2339, opened 2020-09-29
Allow blacklight:index:seed to pull the fixture file from an env vari…

cbeer:
…able or a well-known location in the app


jcoyne (changes requested):
Looks good. I think it would be helpful if you add some documentation about this change.  Add an example invocation perhaps?

jcoyne (approved):

Issue #2391, opened 2020-12-16
Email bookmarks past first page not working


dkinzer:
Emailing of saved bookmarks past the first page is not working.

This seems to be happening because the 'page' param is not being propagated by the email form.

PR #2392, opened 2020-12-16
Propagate the page param into email action.

dkinzer:
Fixes projectblacklight/blacklight#2391

### DIFF
--- a/app/views/catalog/_email_form.html.erb
+++ b/app/views/catalog/_email_form.html.erb
@@ -33,6 +33,9 @@
     <%- if params[:per_page] -%>
       <%= hidden_field_tag "per_page", params[:per_page] %>
     <%- end -%>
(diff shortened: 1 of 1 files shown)


cbeer:
Hm. For regular search results, we're using the `id[]` parameter to figure out what documents to export. It looks like we're passing `id[]` here too, but the bookmarks controller has overridden the method:

https://github.com/projectblacklight/blacklight/blob/master/app/controllers/concerns/blacklight/bookmarks.rb#L25-L29

The custom `action_documents` method sure looks like it intended to include all bookmarked documents in the action (which.. I think makes sense? it'd be annoying as a user to have to go to each page to get all the things you selected, right?). Maybe it's possible `SearchService#fetch` or the equivalent previously ignored `per_page` and now it does?

PR #2394, opened 2020-12-28
Support ruby 3

jcoyne:


spec/lib/blacklight/open_struct_with_hash_access_spec.rb
```diff
@@ -62,15 +62,6 @@
     end
   end
 
-  describe "#replace" do
```
cbeer, line 65: I'm confused; are we just dropping these methods without deprecation?

app/presenters/blacklight/facet_item_presenter.rb
```diff
@@ -46,7 +46,7 @@ def label
         facet_config.query[value][:label]
       elsif facet_config.date
         localization_options = facet_config.date == true ? {} : facet_config.date
-        I18n.l(Time.zone.parse(value), localization_options)
+        I18n.l(Time.zone.parse(value), **localization_options)
```
cbeer, line 49: Do we need to require i18n >= 1.7.0 in our gemspec (when named params were added to the i18n gem), or otherwise handle both?

spec/lib/blacklight/open_struct_with_hash_access_spec.rb
jcoyne: Ostruct in Ruby 3 has quite a different implementation than in Ruby 2.  As such, I'm unable to devise a method such that these continue to work in the way they did in ruby 2.  Any ideas?

app/presenters/blacklight/facet_item_presenter.rb
jcoyne: Good idea.

lib/blacklight/configuration.rb
```diff
@@ -166,12 +168,12 @@ def initialize(hash = {})
     end
 
     def document_model
-      super || ::SolrDocument
+      self[:document_model] || ::SolrDocument
```
cbeer, line 171: It looks like ostruct 0.3.2 fixes the `super` bug. I don't know if we want to pin the gem to >= 0.3.2 to get the fix?

jcoyne:
superseded by #2406

PR #2402, opened 2021-01-05
Blow away the document key for document_actions when used to render t…

### LINKED ISSUE #2400: Bookmark page is causing an error when viewing using add_show_tools_partial
Introduced in Blacklight v7.12.0 via https://github.com/projectblacklight/blacklight/commit/7651c768c8605fd1fdf400c3d10e50f559078fc5

`render_show_doc_actions_method_from_blacklight?` is always evaluating `true` for `add_show_tools_partial`.

This is one reason for the GeoBlacklight build to be failing: https://github.com/geoblacklight/geoblacklight/blob/main/lib/generators/geoblacklight/templates/catalog_controller.rb#L246-L250 (and for a Bookmarks page in GeoBlacklight to cause an error to render).

I'm not 💯 sure of the fix here, but I'm thinking `render_show_doc_actions_method_from_blacklight?` needs to be more clever?

cbeer:
…he batch action for bookmarks; fixes #2400


mejackreed (approved):
Thanks for looking into this. Working great for me.

PR #2411, opened 2021-01-08
Prevent incorrect deprecation warning about thumnail presenter

### LINKED ISSUE #2410: Erroneous deprecation warning about thumbnail presenter
```
DEPRECATION WARNING: Detected as custom thumbnail presenter; make sure it has a #render method that returns just the thumbnail image tag. (called from call at /Users/jcoyne85/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/blacklight-23265db8aafa/app/components/blacklight/document/thumbnail_component.html.erb:6
```

This is because `Blacklight::ThumbnailPresenter#thumbnail` returns an instance not a class.

jcoyne:
Fixes #2410

### DIFF
--- a/app/components/blacklight/document/thumbnail_component.html.erb
+++ b/app/components/blacklight/document/thumbnail_component.html.erb
@@ -6,7 +6,7 @@
       <% warn_about_deprecated_behavior %>
       <%= value %>
     <% else %>
-      <%= @view_context.link_to_document(@presenter.document, value, 'aria-hidden': true, tabindex: -1, counter: @counter) %>
+      <%= helpers.link_to_document(@presenter.document, value, 'aria-hidden': true, tabindex: -1, counter: @counter) %>
     <% end %>
   </div>
 <% end %>
--- a/app/components/blacklight/document/thumbnail_component.rb
+++ b/app/components/blacklight/document/thumbnail_component.rb
@@ -10,7 +10,7 @@ module Blacklight
       def initialize(presenter:, counter:)
         @presenter = presenter
         @counter = counter
-        @use_thumbnail_tag = @presenter.thumbnail != Blacklight::ThumbnailPresenter
+        @use_thumbnail_tag = !@presenter.thumbnail.instance_of?(Blacklight::ThumbnailPresenter)
       end
 
       def render?
--- a/blacklight.gemspec
+++ b/blacklight.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |s|
   s.add_dependency "deprecation"
   s.add_dependency "i18n", '>= 1.7.0' # added named parameters
   s.add_dependency "ostruct", '>= 0.3.2'
-  s.add_dependency "view_component"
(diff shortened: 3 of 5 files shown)


app/components/blacklight/document/thumbnail_component.html.erb
```diff
@@ -6,7 +6,7 @@
       <% warn_about_deprecated_behavior %>
       <%= value %>
     <% else %>
-      <%= @view_context.link_to_document(@presenter.document, value, 'aria-hidden': true, tabindex: -1, counter: @counter) %>
+      <%= helpers.link_to_document(@presenter.document, value, 'aria-hidden': true, tabindex: -1, counter: @counter) %>
```
jcoyne, line 9: `@view_context` is the parent, in this case it's an instance of `Blacklight::DocumentComponent` which causes:

```
     NameError:
       undefined local variable or method `search_state' for #<Blacklight::DocumentComponent:0x00007ff256fe40e8>
       Did you mean?  search_field_tag
     # ./app/helpers/blacklight/url_helper_behavior.rb:9:in `url_for_document'
     # ./app/helpers/blacklight/url_helper_behavior.rb:42:in `block in link_to_document'
     # ./app/helpers/blacklight/url_helper_behavior.rb:41:in `link_to_document'
     # ./app/components/blacklight/document/thumbnail_component.html.erb:11:in `call'
     # (eval):3:in `render_template_for'
     # ./app/components/blacklight/document_component.rb:90:in `thumbnail'
     # ./app/components/blacklight/document_component.html.erb:27:in `block in call'
     # ./app/components/blacklight/document_component.html.erb:1:in `call'
     # (eval):3:in `render_template_for'
     # ./spec/components/blacklight/document_component_spec.rb:11:in `block (2 levels) in <top (required)>'
```

`helpers` goes back to the root where the rails helpers are present.

cbeer: I didn't have good luck with `helpers` and overrides earlier, but maybe they've fixed something upstream. Instead, I had to explicitly render things in the view_context context, e.g.

https://github.com/projectblacklight/blacklight/blob/master/app/components/blacklight/document_component.rb#L97

jcoyne: I'm not sure I understand where you've been seeing trouble. It might be fixed. I haven't noticed any problems with `helpers` though.

cbeer: Yeah, it looks like `helpers` has changed around a bunch depending on the view_components version. It wasn't there at first, then was an alias for `view_context`, then `view_context.controller.view_context`, then back to `view_context` before being reverted. So, maybe it works right now?

We might want to pin to `>= 2.23.0` just to avoid any weird inconsistencies?

jcoyne: Okay, let's do that.

PR #2416, opened 2021-01-14
Prefer safe navigation over `try`

cbeer:


jcoyne (approved):

jcoyne:
It looks like `try` first does a `respond_to?` before calling the method.

cbeer:
Yeah -- I think in all these case we should be calling it on an OpenStruct instance (except for the presenter change, which... i guess was a bug that `try` actually masked) which will respond to stuff.

PR #2421, opened 2021-01-22
Backport some modal a11y changes

jkeck:
Another potential change that could be tacked onto this (but I'm less convinced to do) is to replace the `aria-hidden` with an `aria-modal` attribute when showing the modal and swapping them back when hiding.  This happens for us in BS4 but not BS3.


app/views/shared/_ajax_modal.html.erb
```diff
@@ -1,5 +1,5 @@
-<div id="ajax-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="modal menu" aria-hidden="true">
```
jkeck, line 1: Not sure if there is any concern that somebody has "patched" this a11y issue downstream by adding elements w/ those IDs in their application.  Latest blacklight 7 does not use `aria-labelledby` on the modal element FWIW.

I think it may be more likely that a downstream application had an unrelated element w/ the ID `menu` that may have caused all kinds of other issues w/ this attribute.

PR #2431, opened 2021-03-05
Update ViewConfig#display_label for consistency with similar methods

cbeer:


lib/blacklight/configuration/view_config.rb
```diff
@@ -17,15 +17,18 @@ def search_bar_presenter_class
       super || Blacklight::SearchBarPresenter
     end
 
-    def display_label(key)
+    def display_label(deprecated_key = nil, **options)
+      Deprecation.warn('Passing the key argument to ViewConfig#display_label is deprecated')
```
jcoyne, line 21: Does this need a conditional?

jcoyne (changes requested):

lib/blacklight/configuration/view_config.rb
cbeer, line 21:
```suggestion
      Deprecation.warn('Passing the key argument to ViewConfig#display_label is deprecated') if deprecated_key.present?
```

jcoyne (approved):

PR #2435, opened 2021-03-30
Try harder to map response parameters provided using the JSON Request API

cbeer:
I feel like I must be missing something; surely it can't be this hard to know what parameters Solr's actually decided to apply??


epugh:
Does this help? https://solr.apache.org/guide/8_7/common-query-parameters.html#echoparams-parameter

mejackreed (approved):

PR #2437, opened 2021-04-05
Add support for JSON facet responses

cbeer:
This PR adds support for Solr's "new"(ish) JSON facet responses. This initial implementation maps the json facet response data into compatible structures for term-based facets, facet queries, and pivot facets.

Additional JSON facet response data is available through the `data` accessor on the fields and facet items, which provides access to functions, stats,  and other bucket aggregations.

This PR also adds support for sending solr requests that include JSON facet api fields:

```
config.add_facet_field 'my_field', json: { missing: true, numBuckets: true }
#...
config.add_facet_fields_to_solr_request!
```

(See https://solr.apache.org/guide/8_8/json-facet-api.html#terms-facet for other parameters)

### DIFF
--- a/lib/blacklight/solr/facet_paginator.rb
+++ b/lib/blacklight/solr/facet_paginator.rb
@@ -20,6 +20,8 @@ module Blacklight::Solr
     def initialize(all_facet_values, arguments = {})
       super
 
+      @sort = arguments[:sort].keys.first.to_s if arguments[:sort].is_a? Hash
+
       # count is solr's default
       @sort ||= if @limit.to_i > 0
                   'count'
--- a/lib/blacklight/solr/response/facets.rb
+++ b/lib/blacklight/solr/response/facets.rb
@@ -29,6 +29,7 @@ module Blacklight::Solr::Response::Facets
   # represents a facet; which is a field and its values
   class FacetField
     attr_reader :name, :items
+    attr_accessor :missing
 
     def initialize name, items, options = {}
       @name = name
@@ -50,6 +51,14 @@ module Blacklight::Solr::Response::Facets
 
     def prefix
       @options[:prefix] || solr_default_prefix
+    end
+
+    def type
+      @options[:type] || 'terms'
+    end
+
+    def data
+      @options[:data] || {}
     end
 
     def index?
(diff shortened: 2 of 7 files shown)


magibney:
For comparison, another approach:
Adding to search request:
https://github.com/upenn-libraries/discovery-app/blob/781c2ba/app/models/search_builder.rb#L61-L66
Parsing out of response:
https://github.com/upenn-libraries/discovery-app/blob/781c2ba/lib/blacklight/solr/response/facets.rb#L250-L337
Example in config:
https://github.com/upenn-libraries/discovery-app/blob/781c2ba/app/controllers/catalog_controller.rb#L298-L369
Also relevant:
https://github.com/upenn-libraries/discovery-app/blob/781c2ba/lib/penn_lib/json_facet.rb

I regret that I haven't had time to make this into a proper PR -- mainly really because it's based on v6.8.0, and I haven't had time/mandate to port forward to the main branch of BL, especially since "JSON Facet" hadn't previously seemed to be on anyone's roadmap and I didn't want to force the issue.

The changes here may at first glance appear intermingled with some of our local modifications, but they're designed to be cleanly separated into a general abstraction. Off-the-cuff random thoughts here, wrt why this is more complex than it might otherwise be: "JSON Facet" can quite powerful, but if the BL abstractions hide that power, then it's not clear to me what JSON facet actual gets you, practically speaking, over the current/legacy/flat Solr facet API.

I'm curious to know your thoughts. If nothing else, I hope the Penn example linked to here provides a concrete, non-trivial example use case that actually requires JSON Facet (and also a framework for integrating different facet sorts and paging in the existing Facet UI modal interface).

cbeer:
Thanks @magibney ; it was really useful to see some real-world uses. I think this implementation is not-dissimilar to what you're doing. I think the big difference is I'm punting on how to display any of the JSON facet data that's outside the regular faceting happy paths (so... regular term facets, groups of query facets, and pivot-style facets should work).  

I think that's probably fine, because any nuanced use of json facets can just provide their own view component to do something useful with the raw data. Likewise, sending json facet requests is kinda left up to the implementor, we should just pass it through.

jcoyne (approved):

PR #2439, opened 2021-04-06
Include the BlacklightHelper after initializing

### LINKED ISSUE #2438: Warning about autoloading in Rails 6
```
DEPRECATION WARNING: Initialization autoloaded the constants Blacklight::UrlHelperBehavior, Blacklight::HashAsHiddenFieldsHelperBehavior, Blacklight::LayoutHelperBehavior, Blacklight::IconHelperBehavior, Blacklight::BlacklightHelperBehavior, and BlacklightHelper.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload Blacklight::UrlHelperBehavior, for example,
the expected changes won't be reflected in that stale Module object.

These autoloaded constants have been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /Users/jcoyne85/workspace/sul-dlss/argo/config/environment.rb:7)

```

This seems to originate from https://github.com/projectblacklight/blacklight/blob/c310fc8cb07635cac8e2474b1afa9fafdb6c89de/lib/blacklight/engine.rb#L10-L12

jcoyne:
Because autoloading is not supported during the initialization step in rails any more.
Fixes #2438

### DIFF
--- a/lib/blacklight/engine.rb
+++ b/lib/blacklight/engine.rb
@@ -8,7 +8,9 @@ module Blacklight
     # BlacklightHelper is needed by all helpers, so we inject it
     # into action view base here.
     initializer 'blacklight.helpers' do
-      ActionView::Base.include BlacklightHelper
+      config.after_initialize do
+        ActionView::Base.include BlacklightHelper
+      end
     end
 
     # This makes our rake tasks visible.


cdmo (approved):
Tried this out locally in our catalog that uses version 7.10.0 and it worked! I still have some other problems reported, see:

```
DEPRECATION WARNING: Initialization autoloaded the constants Blacklight::Document::SchemaOrg, Blacklight::Document::SemanticFields, Blacklight::Document::CacheKey, Blacklight::Document::Export, Blacklight::Document, and Blacklight::Document::ActiveModelShim.
```

but those are not related to this issue specifically

Thanks!

PR #2440, opened 2021-04-07
Remove deprecated and unused content areas from SearchBarComponent

jcoyne:


cbeer:
I'm not sure why you're removing `append` and `prepend`. I thought they would be useful for applications?

app/components/blacklight/metadata_field_layout_component.rb
```diff
@@ -3,7 +3,8 @@
 module Blacklight
   class MetadataFieldLayoutComponent < ::ViewComponent::Base
     with_collection_parameter :field
-    with_content_areas :label, :value
+    renders_one :label
+    renders_one :value
```
cbeer, line 7: I'm idly wondering if `renders_many :value` would be helpful, but i'm still trying to understand the new slots approach..

regardless, do we need to bump the view_component minimum version?

jcoyne: Are they used by anything? I didn't see them used in Blacklight, and they don't have any tests or documentation that I could find, so I'm having trouble visualizing how they would be used.

jcoyne:
Are pretend/append used by anything? I didn't see them used in Blacklight, and they don't have any tests or documentation that I could find, so I'm having trouble visualizing how they would be used.

cbeer:
They can be used by downstream components. I don't think anything exists to do that, but it seems like an obvious thing that people are doing today by overriding the whole partial, so I don't know why we'd take that slot away.

I'm not sure what documentation you're expecting other than something that says... these content areas exist.. and to me, it feels like any test for them would just be exercising view_components for the sake of it, no?

jcoyne:
superseded by #2443

PR #2459, opened 2021-05-10
Support bootstrap 4 + 5

cbeer:
This addresses the major differences/incompatibilities between bootstrap 4 + 5 by supporting both markups simultaneously.

I briefly considered adding something to allow us to render one or the other, but so many changes need to be made in the javascript (and in the i18n files, apparently), it seemed like more of a challenge than was warranted by these mostly innocuous changes 🤷


jcoyne:
Do you really want to support both at the same time?  It seems to me that this will add significant complexity and I'd rather avoid it if possible and just go to Bootstrap 5.

cbeer:
I don't think we can just make the switch without a major version bump, and it's a huge pain to address deprecation warnings and blacklight-changes while also dealing with visual style changes at the same time. Bootstrap 4->5 isn't quite as drastic as the other changes, but that probably makes it easier to build in support for both 🤷

PR #2464, opened 2021-05-27
Fix pivot facet fields not recognized.

### LINKED ISSUE #2463: Pivot facet field fields are not added to search_state.filters
Configured pivot facet fields are failing to be added to the`search_state.filters`.  This issue is related to #2401 but in this case the fields are actually being added via a facet configuration but  are still not being processed properly.

## Reason to fix this bug:
This bug is unlikely to effect users that do not have a local custom feature that depends on the accurate representation of the search state as it pertains to pivot facet fields.  As is happens Temple U. has a custom pivot field feature that uses the search state to make decisions about rendering. The custom feature reasonably expects the search state filters to include the pivot fields state but without a workaround they do not appear.

dkinzer:
Fixes #2463

Pivot facet fields are not being added to the `search_state.filters`.
This commit fixes the issue by correctly processing the pivot field when
`field.pivot` is present.

### DIFF
--- a/lib/blacklight/search_state.rb
+++ b/lib/blacklight/search_state.rb
@@ -132,11 +132,14 @@ module Blacklight
     end
 
     def filters
-      @filters ||= blacklight_config.facet_fields.each_value.map do |value|
-        f = filter(value)
-
-        f if f.any?
-      end.compact
+      @filters ||= blacklight_config.facet_fields.each_value.reduce([]) do |acc, value|
+        if value.pivot
+          acc + value.pivot.map { |v| value.class.new(key: v) }
+                     .map { |f| filter(f) }.select(&:any?)
+        else
+          acc + Array(filter(value)).select(&:any?)
+        end
+      end
     end
 
     def filter(field_key_or_field)
--- a/spec/lib/blacklight/search_state_spec.rb
+++ b/spec/lib/blacklight/search_state_spec.rb
@@ -544,4 +544,33 @@ RSpec.describe Blacklight::SearchState do
       expect(search_state.facet_prefix).to eq 'A'
     end
   end
+
+  describe '#filters' do
+    context 'pivot facet config but no facet params' do
+      it 'maps no filters' do
+        blacklight_config.add_facet_field 'some_key', pivot: %w[pivot_field_1 pivot_field_2]
+        expect(search_state.filters).to eq([])
+      end
+    end
+
+    context 'pivot facet config and some pivot facet params' do
+      let(:params) { { 'f' => { 'pivot_field_1' => 'foo' } } }
+
+      it 'maps the pivot matching param pivot_field' do
+        blacklight_config.add_facet_field 'some_key', pivot: %w[pivot_field_1 pivot_field_2]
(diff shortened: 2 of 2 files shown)


dkinzer:
Not sure what the failure is about, since `bundle exec rake` is passing locally for me.

dkinzer:
I'm not able to reproduce this error locally using ruby 3.0.1 ... So I think the error either not real or it's dependent on more than the ruby version.

dkinzer:
I decided to just force push a `git commit --amend --no-edit` and that fixed it.

cbeer:
I was looking at why our integration test for this behavior works fine:

https://github.com/projectblacklight/blacklight/blob/master/spec/features/facets_spec.rb#L66-L79

This seems like it's only for pivot facet fields that don't have the subfields configured (and in the integration test, all of the  pivot field, format, language_ssim are defined)?

lib/blacklight/search_state.rb
```diff
@@ -132,11 +132,14 @@ def remove_query_params
     end
 
     def filters
-      @filters ||= blacklight_config.facet_fields.each_value.map do |value|
-        f = filter(value)
-
-        f if f.any?
-      end.compact
+      @filters ||= blacklight_config.facet_fields.each_value.reduce([]) do |acc, value|
+        if value.pivot
```
cbeer, line 136: I wonder if it would it make sense to push the pivot facet handling down into `FilterField` (or maybe a subclass, `PivotFilterField`?)?

dkinzer:
> This seems like it's only for pivot facet fields that don't have the subfields configured (and in the integration test, all of the pivot field, format, language_ssim are defined)?

Yeah I believe that must be the case.

lib/blacklight/search_state.rb
dkinzer: @cbeer do you mean so that `f = filter(value)` itself handles pivots?  I think I couldn't figure out how to do that in a straight forward way.

barmintor: A complicating factor (vis-a-vis filter fields) is that the pivot values represented in the URL as plain old facets, and assembled into pivots in the search builder. We could put a PivotFilterField together, but it would either need to parameterize its values differently (maybe it actually uses the key, and has a hash of values) or just be bossy about the implicated pivot facets (they're all single values that it manipulates without regard to other facets).

barmintor: There are also some changes to make to the various sanitization stuff we've been doing if a pivot facet's component fields don't need to be present as configured filters.

dkinzer: @barmintor shouldn't the fact that the field is defined via configuration to be a of type pivot be all we need?  And that's already happening.

barmintor: For what it's worth, this is what it would look like: https://github.com/projectblacklight/blacklight/pull/2696 (I don't think it's a bad idea, but it's definitely an 8.x change).

dkinzer:
I've documented in a workaround in the issue. Maybe just adding descriptive tests to further document suffices.

PR #2468, opened 2021-06-10
Update the search bar component to use the form builder

cbeer:
By using `form_with` with the `namespace` argument, Rails will automatically prefix the generated form element ids for us.

The additional `form_options` allows downstream applications to fine-tune the form behavior.


barmintor:
It took a bit to rebase this, and I needed to add a `local` option to the `form_with` call to be able to verify the changes in a feature spec. That's up for review in #2604 if you want to confirm, @cbeer.

Issue #2029, opened 2018-11-08
When spellcheck.collate is turned on, Blacklight should display collated "Did you mean" links


sandbergja:
The default "Did you mean" behavior in Blacklight is not helpful for users when their query contains multiple words (see #2028).  If a user types in a 5-word query, for example, Blacklight will go ahead and suggest a correctly spelled one-word query instead.  Expected behavior would be to offer the user a 5-word query with correct spelling.

Solr is capable of doing this, with spellcheck collation: https://lucene.apache.org/solr/guide/6_6/spell-checking.html#SpellChecking-Thespellcheck.collateParameter

But it looks like Blacklight doesn't pay any attention to the collations array in solr responses.  Here's an excerpt of an example response from a solr query against the sample marc records:

    "spellcheck":{
      "suggestions":[
        "strongh",{
          "numFound":3,
          "startOffset":2,
          "endOffset":9,
          "origFreq":0,
          "suggestion":[{
              "word":"strong",
              "freq":1},
            {
              "word":"sroṅ",
              "freq":1},
            {
              "word":"stran",
              "freq":1}]},
        "medicne",{
          "numFound":1,
          "startOffset":10,
          "endOffset":17,
          "origFreq":0,
          "suggestion":[{
              "word":"medicine",
              "freq":1}]}],
      "correctlySpelled":false,
      "collations":[
        "collation","*:strong medicine"]}}

It would be so nice if that "strong medicine" collation could be presented to the user, rather than the jumble of single-word suggestions.

sandbergja:
Closed by #2605

PR #2472, opened 2021-06-16
logic to check if we have a collation available in the spelling response versus just words.

epugh:
This adds a check to see if we have a collation, and then uses that instead of the individual words.  I didn't try to handle having multiple collations, we just grab a singular one.  

See https://github.com/projectblacklight/blacklight/issues/2029.

I also didn't change the default blacklight solrconfiguration, as I wasn't sure about how to test it.

This is what my spellcheck set up looks like, including a custom `mm=100%` parameter to make sure to filter to only collations that we have documents for:
```
      <!-- spellchecking -->
      <str name="spellcheck">true</str>
      <str name="spellcheck.dictionary">title</str>
      <str name="spellcheck.onlyMorePopular">true</str>
      <str name="spellcheck.extendedResults">true</str>
      <str name="spellcheck.collate">true</str>
      <str name="spellcheck.maxCollations">100</str>
      <str name="spellcheck.maxCollationTries">5</str>
      <str name="spellcheck.count">5</str>
      <!-- filter out any collations that don't match documents -->
      <str name="spellcheck.collateParam.mm">100%</str>
```

### DIFF
--- a/app/components/blacklight/response/spellcheck_component.rb
+++ b/app/components/blacklight/response/spellcheck_component.rb
@@ -8,7 +8,12 @@ module Blacklight
       # @param [Array<String>] options explicit spellcheck options to render
       def initialize(response:, options: nil)
         @response = response
-        @options = options || @response&.spelling&.words
+        @options = options
+        if @response&.spelling&.collation
+          @options = [@response.spelling.collation]
+        elsif @response&.spelling&.words
+          @options = response.spelling.words
+        end
       end
 
       def link_to_query(query)


app/components/blacklight/response/spellcheck_component.rb
```diff
@@ -8,7 +8,12 @@ class SpellcheckComponent < ViewComponent::Base
       # @param [Array<String>] options explicit spellcheck options to render
       def initialize(response:, options: nil)
         @response = response
-        @options = options || @response&.spelling&.words
+        @options = options
```
barmintor, line 11: There's a change in behavior here that I don't think is intentional: Prior to the PR, a passed options param will always provide the value of `@options`; here, either form of spelling suggestion will override it. Was this intentional?

barmintor:
> I'd love to see a test included with this PR that shows how this would be used.

I think this is important, too. The solrconfig used in CI sets the default value of `spellcheck.collate` to false, so an integration test would need to add an explicit param to test the behavior. A [unit test](https://github.com/projectblacklight/blacklight/blob/main/spec/components/blacklight/response/spellcheck_component_spec.rb) would just need to mock the response and verify the behavior. The [mocked responses](https://github.com/projectblacklight/blacklight/blob/main/spec/models/blacklight/solr/response_spec.rb) for  Blacklight::Solr::Response suggest that collations come back alongside word suggestions - in that scenario, do we want to include both? In this PR, the collations will obscure the word suggestions (maybe intentionally?).

barmintor:
@epugh this change is much easier to spec and verify when rebased, so I took a crack at it in #2605 

There's a minor bug here in that explicitly passed `options` arguments are supposed to "win" - that's also spec'd in the other PR. I'm opening it as a draft until you have a chance to weigh in there.

epugh:
Closing in favour of https://github.com/projectblacklight/blacklight/pull/2605

PR #2476, opened 2021-06-22
Use webdrivers instead of appartition for capybara testing

cbeer:


jcoyne:
Can you explain more about why this change is necessary?

cbeer:
Apparition breaks with aXe and #2478. I'm not sure what apparition gets us that the mainline rails approach doesn't provide 🤷‍♂️ .

jcoyne:
> what apparition gets us that the mainline rails approach doesn't provide 

It was selected because it was one less dependency to have and it seemed to be the approach favored by the capybara developer: https://github.com/teamcapybara/capybara#apparition

I had forgotten that these were the Rails defaults: https://github.com/rails/rails/blob/0b3a63fe07507fca5b2cab6409e85789a46ed624/railties/lib/rails/generators/rails/app/templates/Gemfile.tt#L71-L73

PR #2478, opened 2021-06-22
Test a11y with aXe

cbeer:
Includes #2476 , #2477


spec/features/axe_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Autocomplete suggestions', api: false, js: true do
```
cjcolvar, line 3: It looks like this describe title was copied here and should be changed.  How about "Accessibility testing"?  What about renaming the whole file to be `accessibility_spec.rb` so it is independent of the underlying tech used to do the a11y testing?

cbeer:
```suggestion
RSpec.describe 'Accessibility testing', api: false, js: true do
```

cbeer: I thought about that too; my concern was these tests are really about the technology and how and what it chooses to report on (which isn't and can't be a complete picture of the accessibility anyway) 🤷‍♂️ .

cjcolvar (approved):
Looks good.  Awesome to have some a11y testing built into the automated tests!

PR #2479, opened 2021-06-24
Handles graciously error generating back to catalog link

hectorcorrea:
This handles the case where users' session has invalid information (either a corrupt session or the URLs in the system have changed) that results in an invalid URL. In that case `link_to` throws an exception and prevents the *page* from rendering. 

This fix handles the exception by producing a link that goes back to the root rather than crashing the rendering process.


jcoyne (changes requested):
Rather than rescuing an error around `link_to` is it possible to figure out what `link_url` is set to when this occurs?  Can a test be added to detect that?

PR #2515, opened 2021-11-09
Drop IE11 as a supported browser

### LINKED ISSUE #2332: Drop IE11 support
Microsoft itself is dropping IE11 support for Teams in November 2020 and in the rest of Microsoft 365 in Aug 2021. https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666

Can blacklight do the same?  

This would allow us to use modern javascript without requiring Babel (e.g. https://github.com/projectblacklight/blacklight/pull/2161#discussion_r331052190)

jcoyne:
Fixes #2332

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ rails generate blacklight:install
 * [Issue Tracker](https://github.com/projectblacklight/blacklight/issues)
 * [Support](https://github.com/projectblacklight/blacklight/wiki/Support)
 
+## Browser Compatibility
+
+Blacklight assumes a modern browser with support for ES6. This means we explicitly do not support Internet Explorer 11.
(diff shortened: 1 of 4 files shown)


tpendragon (changes requested):
Can you add something about browser compatibility to the README? Maybe something like:

## Browser Compatibility

Blacklight assumes a modern browser with support for ES6. This means we explicitly do not support Internet Explorer 11.

jcoyne:
@tpendragon Thanks for the suggestion. I've added that.

tpendragon (approved):

PR #2522, opened 2021-11-16
Move the class configuration defaults into Blacklight::Configuration.default_values

cbeer:
And, now that we've dropped Rails 5.x, we can use `#deep_transform_values` directly.


jcoyne:
@cbeer there is a conflict here that needs to be resolved.

dkinzer:
Is there a way to keep track of these backwards incompatible changes to make the upgrade path easier?

dkinzer (approved):

cbeer:
These aren't backwards incompatible changes, just moving where/how the values are set initially.

dkinzer:
@cbeer Yeah, sorry.  I was focusing on the signature of the deep_transform_values method changing without taking into account that's a private mehod.

PR #2527, opened 2021-11-18
Explicity pass the documents to the render_document_index method

jcoyne:
This keeps controller instance variables from appearing in the helpers


dkinzer:
I know that this PR is towards the 8.x release, but I'm wondering if if makes sense to start to keep  track of these types of backward incompatible changes in an upgrade path file somewhere so that it's easier to document how someone can upgrade from 7.x to 8.x.  I think both of these templates are very popular to override.

bess:
@dkinzer I'm seeing that @jcoyne also has a 7.x deprecation warning about this ([here](https://github.com/projectblacklight/blacklight/pull/2530)) and I've added it to the [upgrade notes](https://github.com/projectblacklight/blacklight/wiki/Upgrading-to-Blacklight-8). Does that address your concern?

bess (approved):

PR #2546, opened 2021-11-19
Deprecate render_document_partials. 

### LINKED ISSUE #2537: Consider getting rid of render_document_partials
It seems to only be used by the atom builder: https://github.com/projectblacklight/blacklight/blob/2430033de0f8c54ef7407e28228b3702f1fd3e0a/app/views/catalog/_document.atom.builder#L23

jcoyne:
This was only used by the atom partial renderer. We can replace this with the document component

Fixes #2537

### DIFF
--- a/app/helpers/blacklight/render_partials_helper_behavior.rb
+++ b/app/helpers/blacklight/render_partials_helper_behavior.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 module Blacklight::RenderPartialsHelperBehavior
+  extend Deprecation
+  self.deprecation_horizon = 'blacklight 9.0'
+
   ##
   # Render the document index view
   #
@@ -22,6 +25,7 @@ module Blacklight::RenderPartialsHelperBehavior
       render_document_partial(doc, action_name, locals)
     end, "\n")
   end
+  deprecation_deprecate render_document_partials: 'Render the document_component instead'
 
   ##
   # Return the list of xml for a given solr document. Doesn't safely escape for HTML.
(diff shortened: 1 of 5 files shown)


cbeer:
It's also used quite a bit in overrides:

https://github.com/search?q=render_document_partials&type=code

As a kindness to upgraders, maybe it's worth keeping around in 8.x with a deprecation warning?

jcoyne:
@cbeer sounds good, I've switched it to a deprecation.

app/helpers/blacklight/render_partials_helper_behavior.rb
```diff
@@ -22,6 +25,7 @@ def render_document_partials(doc, partials = [], locals = {})
       render_document_partial(doc, action_name, locals)
     end, "\n")
   end
+  deprecation_deprecate render_document_partials: 'Render the document_component instead'
```
tpendragon, line 28: What does this mean? If I get this error I'm not sure what to do. Can we expand this deprecation error? Or if I just put in `render_document_component` instead of `render_document_partials` it'll be good?

tpendragon (changes requested):

app/helpers/blacklight/render_partials_helper_behavior.rb
jcoyne: that would be a rather long message to stick here:

```
Replace this call with: "
document_component = blacklight_config.view_config(:atom).summary_component
render document_component.new(presenter: document_presenter(document), component: :div, show: true)" instead.
```

This assumes that the current context has a `document_presenter` helper and that `:atom` is the current config context.

tpendragon: I might be in the minority here, but I'm :+1: to that long message. Changing it to "Replace this call with a document component render, e.g: allthat" might be nice to be more clear?

tpendragon (approved):
I'm removing my block, I'm not going to be able to act on this for a while and I don't want to hold you back.

bess (approved):

PR #2547, opened 2021-11-19
Upgrade to latest rubocop

jcoyne:


dkinzer:
Wondering if it's OK to enable dependabot on this project.

dkinzer (approved):

dkinzer:
Actually if we do dependabot, I think we can have it make PRs against both main and the 7.x branch.

PR #2562, opened 2021-11-26
Fix docker compose up broken

### LINKED ISSUE #2557: Docker-compose does not work solr can not be run as root
The docker-compose scripts don't work and can't be used to get a working installation due to solr_wrapper refusing to start as root.

dkinzer:
Backport of https://github.com/projectblacklight/blacklight/pull/2559, Fixes #2557


Dantali0n:
Very happy to report your patch works. I'm a bit weirded out by the solr_wrapper else statement if its not expected to be used / work anyway why not remove it outright?

At least I can now thoroughly compare blacklight against Omeka S, very much appreciated have a good day!

tpendragon (approved):

PR #2566, opened 2021-12-02
Remove dead code for submitting forms in modal

jcoyne:
The selector for this was removed long ago in https://github.com/projectblacklight/blacklight/commit/68affcfee3d28f9824c83be6e2e33bee38b5da2b and it hasn't worked since. Blacklight itself does not use this functionality


jcoyne:
This is used here too: https://github.com/projectblacklight/blacklight/blob/6701ee120dbd0fff95e15946f6b52ef845af38d1/app/views/catalog/_sms_form.html.erb#L1

PR #2568, opened 2021-12-02
Use dependency injection for handling the bootstrap modal

jcoyne:


cbeer:
I'd prefer we did this in a way that we can backport to Blacklight 7.x (preserving IE11 support, e.g. maybe with just a dumb `if else` ).. at least until we get 8.0 out the door.

jcoyne:
@cbeer Bootstrap 5 does not support IE11. https://getbootstrap.com/docs/5.0/getting-started/browsers-devices/#supported-browsers

cbeer:
Yes, but Blacklight 7 does.

jcoyne:
@cbeer You can't use Bootstrap 5 and support ie11 at the same time. The 8.x javascript and the 7.x javascript have already diverged substantially, I don't see an easy backport path here.  Perhaps you want to take a shot as this?

cbeer:
I was thinking something like https://github.com/projectblacklight/blacklight/pull/2571. Other than some whitespace conflicts, it seems to backport easily enough 🤷‍♂️

jcoyne:
After  #2571 went in and was backported to release-7.x, I reworked this PR into a refactoring that extracts the bootstrap modal handling from the blacklight specific modal behavior.

app/javascript/blacklight/index.js
```diff
@@ -2,14 +2,32 @@ import BookmarkToggle from './bookmark_toggle'
...
+// Create lazily if someone else created first.
+if (Core.modal === undefined) {
+  Core.modal = {
+    modalSelector: '#blacklight-modal' // a Bootstrap modal div that should be already on the page hidden
+  }
+}
+
+Core.onLoad(function() {
+  let bootstrapModal;
+  if (typeof(jQuery) !== 'undefined' && jQuery.fn.tooltip.Constructor.VERSION) {
```
barmintor, line 19:
```suggestion
  if (typeof(jQuery) !== 'undefined' && jQuery.fn.tooltip.Constructor.VERSION.match(/^4\./)) {
```

barmintor: Recognizing that apps might still pull in jquery (cf https://getbootstrap.com/docs/5.0/getting-started/javascript/#still-want-to-use-jquery-its-possible), maybe better to explicitly test the version number?

barmintor (commented):
Aside from the conflict, I think this seems like a good change. I made a suggestion here only because, as I understand it, you might still have jquery loaded for BS5 - and I'm not sure whether that scenario would complicate the modal behavior. If BS5 with jquery could happily use the BS4 modal implementation, feel free to ignore!

app/javascript/blacklight/index.js
jcoyne: Yes, I have just this situation:
Screen Shot 2022-02-09 at 3 02 24 PM

jcoyne:
@barmintor Thanks for the suggestion, I added that and rebased

barmintor:
Is this change still necessary after https://github.com/projectblacklight/blacklight/pull/2671 ?

jcoyne:
No, I don't think so.

PR #2573, opened 2021-12-02
Allow templates to govern where the output goes 

jcoyne:
Currently we have targets for modal form updates hardcoded in the javascript.  This gives the downstream apps/plugins a way to customize what on the page is updated without having to do hardcoding. This follows the format that turbo streams is using: https://turbo.hotwired.dev/handbook/streams


app/views/catalog/email_success.html.erb
```diff
@@ -1,6 +1,5 @@
-<%= render Blacklight::System::ModalComponent.new do |component| %>
```
cbeer, line 1: Is this effectively requiring browsers support javascript?

jcoyne: The `_success` template is only rendered when the request is initiated by Javascript (being an XHR request), right? https://github.com/projectblacklight/blacklight/pull/2573/files#diff-20dfe95ebbfa7c11f6eff4bead05d3a0b0977b382a0164757c80f29cf5a02061R41

app/javascript/blacklight/modalForm.js
```diff
@@ -0,0 +1,60 @@
...
+
+    bind() {
+      document.addEventListener('submit', (e) => {
+         if (e.target.matches(this.triggerFormSelector))
+          this.onSubmit(e)
+      })
+    }
+
+    // This is like a light-weight version of turbo that only supports append presently.
+    updateTurboStream(data) {
```
tpendragon, line 27: Question: This isn't actually turbo streams, but something that uses similar naming conventions, yeah? Is this just prepping Blacklight to head in the direction of turbo streams?

tpendragon (approved):
There's a little rebasing to do, but this looks good to me. I do have one question, but I think it's probably okay - just formalizing that we're headed in this direction or maybe adding a comment about how we might replace it with actual turbostreams would be good - unless I'm misunderstanding.

barmintor:
Rebased, awaiting CI

PR #2577, opened 2021-12-03
Avoid deprecation with view_component loading

jcoyne:
Ref https://github.com/github/view_component/pull/1117 and https://github.com/github/view_component/pull/1162


cbeer:
Do we need to bump the `view_component` dependency?

jcoyne:
@cbeer good idea. I've bumped the gemspec.

PR #2591, opened 2021-12-09
Implement modal.onFailure without using jquery

jcoyne:


app/assets/javascripts/blacklight/blacklight.js
```diff
@@ -327,19 +327,24 @@
...
+          <span aria-hidden="true">&times;</span>
+        </button>
+        </div>
+        <div class="modal-body">
+          <p>Expected a successful response from the server, but got an error</p>
+          <pre>${this.type} ${this.url}\n${jqXHR.status}: ${errorThrown}</pre>
+        </div>`;
+        // const domparser = new DOMParser();
+        // const dom = domparser.parseFromString(contents, "text/html")
+        // const nodes = dom.querySelectorAll('body > *')
```
dkinzer, line 344: @jcoyne is this left here for documentation purposes?

jcoyne: Oops, this is cruft. I'll pull it.

jcoyne: The files in app/assets/javascript are just compiled versions (using `npm run js-compile-bundle`) of what is in `app/javascript`.  In this instance I didn't recompile after changing src.

jcoyne: I recompiled and pushed the results up. That had the effect of removing these old comments.

dkinzer (approved):

PR #2595, opened 2021-12-11
Use the helper context methods or constructing action routes

cbeer:


jcoyne:
Can you explain why we need this change? Doesn't view_components already include the routes? https://github.com/github/view_component/blob/8d284a08a963fe3c5e7587da73f5d9906c7e7937/lib/view_component/base.rb#L370

cbeer:
blacklight-marc does its own thing: https://github.com/projectblacklight/blacklight-marc/blob/master/app/helpers/blacklight_marc_helper.rb#L8-L17 🤷‍♂️

jcoyne:
Can you drop that in a comment so we don't accidentally factor it out later?

jcoyne (approved):

app/components/blacklight/document/action_component.rb
```diff
@@ -24,14 +24,21 @@ def label
...
+      #     default Rails routing behavior)
+      # - calling out to an implicit helper method with a conventional name (unlikely)
       def url
         url_opts = @url_opts.merge(({ id: @document } if @document) || {})
         if @action.path
-          public_send(@action.path, url_opts)
+          helpers.public_send(@action.path, url_opts)
         elsif url_opts[:id].class.respond_to?(:model_name)
-          url_for([@action.key.to_sym, url_opts[:id]])
+          helpers.url_for([@action.key.to_sym, url_opts[:id]])
```
jcoyne, line 39: Does this need to delegate to `helpers`?

app/components/blacklight/document/action_component.rb
```diff
@@ -24,14 +24,21 @@ def label
...
         else
-          public_send("#{@action.key}_#{controller_name}_path", url_opts)
+          helpers.public_send("#{@action.key}_#{helpers.controller_name}_path", url_opts)
```
jcoyne, line 41: Do we need `helpers.controller_name`?

app/components/blacklight/document/action_component.rb
cbeer: Yes? At least for now, until we sort out a better way for plugins like blacklight-marc (above) to do its thing.

app/components/blacklight/document/action_component.rb
cbeer: Definitely. There's no such method in scope here.

PR #2596, opened 2021-12-15
We don't need to define a collection parameter on DocumentComponent

jcoyne:


cbeer:
I agree we don't need to use collection parameters, but `with_collection` is pretty handy so I'm not sure why we want to just throw it away here.

jcoyne:
having that `with_collection` means that if you want to extend this component, then you have to explicitly accept the `document: nil` keyword arg.  I think this will make backward compatible changes more difficult in the future.

See: https://github.com/sul-dlss/argo/blob/main/app/components/document_component.rb#L3-L7

If you omit it you get the error:
```
The initializer for DocumentComponent doesn't accept the parameter `document`, which is required in order to render it as a collection.

To fix this issue, update the initializer to accept `document`.

See https://viewcomponent.org/guide/collections.html for more information on rendering collections.
```

Can you say more about where you would use the document parameter?

barmintor:
Adding for context, for people (like me!) who weren't familiar with this mechanism:
https://viewcomponent.org/guide/collections.html#with_collection_parameter
https://viewcomponent.org/api.html#with_collection_parameterparameter

My read of the ViewComponent documentation is that removing that `with_collection_parameter` won't prevent the use of `with_collection`, but it will infer the required keyword param from the component name (or presumably the subclass) if `with_collection` is used. Is that right?

If so there's some creature comfort in not requiring the keyword param in subclasses, but it would come at the expense of a bit of a trap - if you were passing a subclass around as a configured implementation, the calling code might not be able to construct keyword params appropriately? Or do I misunderstand?

jcoyne:
I just made this because the `with_collection` seemed unnecessary and imposed a real challenge with configuration. I do not understand why we want to have something that isn't used anywhere.

spec/components/blacklight/document_component_spec.rb
```diff
@@ -156,4 +156,40 @@ def call
...
+    end
+
+    let(:config) do
+      Blacklight::Configuration.new do |config|
+        config.add_index_field 'title_tsim', label: 'Title:'
+      end
+    end
+
+    let(:render) do
+      described_class.with_collection(docs).render_in(view_context)
```
jcoyne, line 183: We never use this class in this way in the code. Why do we want a test for this invocation?

barmintor: Verifying comes at little cost (this is just demonstrating that), and speaking for myself: If I understand that Blacklight is moving to ViewComponents, I am going to expect that I can learn how to use ViewComponents and apply that knowledge to my Blacklight app. Whether the current, say, index view uses `with_collection` or not seems like less of an issue than whether it might.

barmintor:
I just want to reiterate that this PR has never removed `with_collection`, but only the collection parameter override via `with_collection_parameter` - and that the latter was what was complicating subclasses. I've added a commit here to demonstrate that it's still available to client code to address @cbeer's reservation, but it's not strictly necessary to keep it.

barmintor:
I think we should close this in favor of #2698 , which includes the removal of the unnecessary collection parameter specification here.

PR #2602, opened 2022-01-12
Add test for title prefix

jcoyne:


barmintor:
Does the ViewComponent initializer take a block? I know `render`/`render_in` work this way, and an after-initialization call to title would work:
```ruby
  context 'with before_title' do
    before do
      component.title do |c|
        c.before_title { 'Prefix!' } # could also override the let for :render
      end
    end

    it 'shows the prefix' do
      expect(rendered).to have_content "Prefix!"
    end
  end
end
```
If the base ViewComponent initializer doesn't yield to a block and we want it to, we could also add a block behavior to `Blacklight::Component#initialize`?

barmintor:
Re-based and updated the test to configure the slot with a black to `render_in`, since that seems to be the most like our actual use cases in Blacklight.

barmintor (approved):

PR #2318, opened 2020-08-05
Add aria label to search

K8Sewell:
# Summary
Add aria label to the search form 

# Related Issue
#2147 

# WIP
Have added the aria label to the search form but struggling with creating an appropriate test to accompany this change.

Issue #2147, opened 2019-09-20
Consider adding an aria-label to the "search" landmark


jvine:
`role="search"` defines the search landmark in Blacklight. 👍 

However some SRs don't list landmarks that don't have an aria-label. 
- VoiceOver does not list the search landmark.
- NVDA does list it. 

Consider adding `aria-label="search"` to `<form role="search" ...>` for consistency of support.

cjcolvar:
@barmintor Can this be closed by #2603?

PR #2603, opened 2022-01-14
Add aria label to search

barmintor:
@K8Sewell I think your [PR](https://github.com/projectblacklight/blacklight/pull/2318) was a great idea, but it could no longer be merged as-is. I've ported your change over to the relevant component and spec in this branch for review.


barmintor:
Related issue #2147

jcoyne (approved):

PR #2605, opened 2022-01-17
Add collations to did you mean

barmintor:
@epugh this is a rebase of your branch for collations, with some rudimentary tests added to ensure the existing API was satisfied and as well as your changes. Since @bess reviewed the original PR #2472 , I think there's consensus regarding the use case, but @jcoyne raised good questions about specs - there was a departure from intended behavior in the PR (and from the coverage goals we have for the project).

It would have been a bit more daunting to write the specs on the original PR, but since it was opened a spec for `Blacklight::Response::SpellcheckComponent` was merged to main, so now it's only a matter of adding to a couple of existing spec suites. Please let me know if you have any questions about the additional commits, but if you're satisfied that your use case is covered I'll remove the draft tag from this PR.

### DIFF
--- a/app/components/blacklight/response/spellcheck_component.rb
+++ b/app/components/blacklight/response/spellcheck_component.rb
@@ -8,7 +8,12 @@ module Blacklight
       # @param [Array<String>] options explicit spellcheck options to render
       def initialize(response:, options: nil)
         @response = response
-        @options = options || @response&.spelling&.words
+        @options = options
+        if @response&.spelling&.collation
+          @options = [@response.spelling.collation]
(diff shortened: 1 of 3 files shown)


epugh:
@barmintor  this looks great, thank you for taking this on.  Honestly, what I did was kind of a "late night make this work for a demo tomorrow" level code, so this looks great.   I'll close my PR in favour of this one!

barmintor:
Awesome, thanks for getting back to me! I'll remove the draft tag.

tpendragon (approved):

epugh:
thanks for the merge @tpendragon

PR #2610, opened 2022-01-19
Deprecate engine config keys

barmintor:
- Deprecation warnings for top-level config key access
- moved an erroneously excluded key behind `Blacklight::Engine.config.blacklight`
- moved the default value for a key intended to be in `Blacklight::Configuration` out of `Blacklight::Engine.config`


barmintor:
Aargh.

PR #2618, opened 2022-02-11
Work around for rspec-mocks#1460 issue with expectations on kwargs

cbeer:


jcoyne:
Here's the issue: https://github.com/rspec/rspec-mocks/issues/1460

PR #2626, opened 2022-02-17
Convert search history display to use components

cbeer:
This will require backporting to 7.x with a nice deprecation warning for the `Blacklight::SearchHistoryConstraintsHelperBehavior` methods.

This PR does change the way the search history is displayed. Currently, facets are grouped by their field:

Screen Shot 2022-02-17 at 15 49 17

This PR displays them more like the way constraints are display elsewhere in the application:
Screen Shot 2022-02-17 at 15 49 47

---

Part of getting bounding box info (from `params[:bbox]`) displaying on the search history page without overriding helpers: https://github.com/geoblacklight/geoblacklight/pull/1163/


app/helpers/blacklight/url_helper_behavior.rb
```diff
@@ -102,8 +102,9 @@ def link_back_to_catalog(opts = { label: nil })
     link_to label, link_url, opts
   end
 
-  # Search History and Saved Searches display
+  # Search History s display
```
jcoyne, line 105: Typo here

jcoyne (changes requested):

tpendragon (approved):

PR #2628, opened 2022-02-24
Make some more search-related components configurable

cbeer:


lib/blacklight/configuration.rb
```diff
@@ -84,7 +84,11 @@ def default_values
             # what field, if any, to use to render grouped results
             group: false,
             # additional response formats for search results
-            respond_to: OpenStructWithHashAccess.new
+            respond_to: OpenStructWithHashAccess.new,
+            # component class used to render the facet grouping
+            facet_group_component: nil,
```
jcoyne, line 89: Does it make sense to put the defaults here rather than in the ERB? Then we don't have to have a conditional and it's easy to find the expected interface.

cbeer: I wish.. unfortunately class reloading makes that pretty screwy. We can get away with it for stuff in `lib`, but anything in `app` doesn't play nice.

jcoyne (approved):

PR #2632, opened 2022-02-25
Use the class_methods module provided by ActiveSupport::Concern

jcoyne:
This is the documented way of using AS::Concern, using the ClassMethods module seems to rely on an undocumented implementation detail.


cbeer:
Seems like `class_methods` came in with Rails 4.2. It's probably time we start using it too 😬

Issue #2647, opened 2022-02-26
Blacklight should have a generated ViewComponent base class that permits overriding a template in the installing app 


barmintor:
- The `ViewComponent::Base.view_component_path` attribute should be usable to decide what relative path to search
- A mixin should be available to override the unstable `_sidecar_files` method to check against Rails root before checking against the source location
- The mixin should be optional since it will have a performance penalty across the app

barmintor:
The mixin approach would probably necessitate something like a generated `ApplicationComponent` that could extend a module with the relevant overrides?

barmintor:
release-7.x fix: [766b3e305c175d54a7f3563e39a2963bb8091446](https://github.com/projectblacklight/blacklight/commit/766b3e305c175d54a7f3563e39a2963bb8091446)
main fix: [60cb79bab87dd18dad0aed8dda2f787f4abc59e2](https://github.com/projectblacklight/blacklight/commit/60cb79bab87dd18dad0aed8dda2f787f4abc59e2)

PR #2664, opened 2022-03-12
implementation ideas for #2647

barmintor:
A couple of ideas for how to implement a base view component that would allow a more familiar template override pattern to Blacklight installations not using view components, and a less labor intensive way to have only a change in template.

See also: https://github.com/github/view_component/issues/411

### DIFF
--- a/app/components/blacklight/compiler_demonstration_component.html.erb
+++ b/app/components/blacklight/compiler_demonstration_component.html.erb
@@ -1,0 +1,1 @@
+<h1>Engine Template</h1>
--- a/app/components/blacklight/compiler_demonstration_component.rb
+++ b/app/components/blacklight/compiler_demonstration_component.rb
@@ -1,0 +1,30 @@
+# this approach has less API vulnerability but won't allow addition of sidecar files
+module Blacklight
+  class CompilerDemonstrationComponent < ::ViewComponent::Base
+    class << self
+      def compiler
+        @__vc_compiler ||= EngineCompiler.new(self)
+      end
+    end
+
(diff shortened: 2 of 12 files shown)


barmintor:
The strategy for these would be a little complicated - you need to run the tests in the installing app to really verify the behavior, and the generated app isn't really set up for that.

barmintor:
One strategy is to have a custom Compiler implementation that looks for a component template in the parallel path under `Rails.root` and prefers it to the template in the engine path. This makes use of a ViewComponent API that documented as private, but not experimental, on the ViewComponent::Base class method `compiler`. This is the more stable API of my two suggestions, but in the event that a component uses translation sidecar files, they cannot be overridden (or added to).

A second strategy is to override the `_sidecar_files` method class method on ViewComponent::Base. This has the significant drawback of being a documented unstable API, but permits overriding both partials and translation sidecars. It also "feels" like a closer approximation of partial search paths.

Proposing the inclusion of one of these strategies would also require a decision about how to accomplish it - one method would be defining a base component class (`Blacklight::BaseComponent < ViewComponent::Base` or similar) and inheriting from it in all the Blacklight components. This would be  more transparent to installations, but would impose a modest performance penalty regardless of whether the partials were being overridden or not. Another option would be to introduce a generated base component, e.g. `ApplicationViewComponent`, and to provide the override strategies as optional concerns/mixins that an installation could elect to include in the generated base class. That would introduce a step to upgrading, but would permit keeping the faster template identification in ViewComponent::Base if overrides were not necessary.

barmintor:
Closing since we have pulled one of the approaches into `main` and the 7.x branch.

PR #2666, opened 2022-03-14
BookmarksController#create should not return success on invalid input

### LINKED ISSUE #2665: Bookmark create method does not check for invalid input
I am trying to add a custom validator to our Bookmark model.  With the validator in place on the model, when I send invalid input to the BookmarkController#create, nothing persists to the database (as expected), but the controller still considers it a success, so I can't notify the user about the validation issue.  The issue is on this line: https://github.com/projectblacklight/blacklight/blob/3c204022a278bab6e4fad4d17422627b63f28d06/app/controllers/concerns/blacklight/bookmarks.rb#L83-L85

`current_or_guest_user.bookmarks.create(bookmark)` will return bookmark [whether or not validations pass](https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-create) -- so, a truthy value even if nothing got persisted to the database.

I think checking `current_or_guest_user.bookmarks.create(bookmark).valid?` instead should do the trick.  PR coming up.

sandbergja:
Closes #2665

### DIFF
--- a/app/controllers/concerns/blacklight/bookmarks.rb
+++ b/app/controllers/concerns/blacklight/bookmarks.rb
@@ -80,12 +80,15 @@ module Blacklight::Bookmarks
 
     current_or_guest_user.save! unless current_or_guest_user.persisted?
 
-    success = @bookmarks.all? do |bookmark|
-      current_or_guest_user.bookmarks.where(bookmark).exists? || current_or_guest_user.bookmarks.create(bookmark).valid?
+    bookmarks_to_add = @bookmarks.reject { |bookmark| current_or_guest_user.bookmarks.where(bookmark).exists? }
+    success = ActiveRecord::Base.transaction do
+      current_or_guest_user.bookmarks.create!(bookmarks_to_add)
+    rescue ActiveRecord::RecordInvalid
+      false
     end
 
     if request.xhr?
-      success ? render(json: { bookmarks: { count: current_or_guest_user.bookmarks.count } }) : render(plain: "", status: "500")
+      success ? render(json: { bookmarks: { count: current_or_guest_user.bookmarks.count } }) : render(json: current_or_guest_user.errors.full_messages, status: "500")
     else
       if @bookmarks.any? && success
         flash[:notice] = I18n.t('blacklight.bookmarks.add.success', count: @bookmarks.length)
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/blacklight/bookmarks.rb
```diff
@@ -81,7 +81,7 @@ def create
     current_or_guest_user.save! unless current_or_guest_user.persisted?
 
     success = @bookmarks.all? do |bookmark|
-      current_or_guest_user.bookmarks.where(bookmark).exists? || current_or_guest_user.bookmarks.create(bookmark)
+      current_or_guest_user.bookmarks.where(bookmark).exists? || current_or_guest_user.bookmarks.create(bookmark).valid?
```
jcoyne, line 84: Won't create already return nil when the record is invalid?

app/controllers/concerns/blacklight/bookmarks.rb
```diff
@@ -81,7 +81,7 @@ def create
     current_or_guest_user.save! unless current_or_guest_user.persisted?
 
     success = @bookmarks.all? do |bookmark|
```
cbeer, line 83: Will the `all?` short-circuit as soon as one bookmark fails to validate? Is the intent to persist what we can and drop any failures on the floor?

I'm not sure where we're creating multiple bookmarks at once either.. but it goes way, way back.

sandbergja: Good questions -- I think the block would stop as soon as it tries to persist something invalid.  Would it be better not to add any at all if one or more are invalid?

sandbergja:
Just added another commit.  It wraps the bookmark creation call in a transaction, so if one of the posted bookmarks is invalid, the whole batch will be rolled back and none will persist to the database.

I also let any validation errors bubble up into the JSON, so that if blacklight apps wish to display the error to the user, they can implement that.

If anybody wants to try this with an actual validation, I've been testing with adding this to app/models/bookmark.rb:
```
  validates_associated :user
```

And this to app/models/user.rb
```
  validates :bookmarks, length: {maximum: 5, message: 'Too many bookmarks!'}
```
Which should start erroring as soon as you try to add your sixth bookmark.

sandbergja:
tried rebasing against main to see if that makes CI complete.

barmintor (approved):
I think with the changes this is merge-able.

barmintor:
Waiting on CI to merge.

barmintor:
rebased to grab CI config changes

PR #2670, opened 2022-03-17
Use inline icons as components

jcoyne:
This makes the icons simple to customize and allows us to deprecate behavior that deals with the internals of the asset pipeline. This approach is an alternative to https://github.com/projectblacklight/blacklight/pull/2668 where we add awareness of other asset pipelines (e.g. propshaft).


cbeer:
We've had a lot of luck recently just putting the SVG inline in the CSS like bootstrap's now doing:

https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L881

jcoyne:
@cbeer the problem with that approach is you can't style the svg using css, can you?

tpendragon (changes requested):
We talked about this on the committer's call and like this pattern. We'd just like to see documentation in the class or method level to describe how to customize the icon if that's what you want to do

jcoyne:
@tpendragon I added class documentation in the two icon classes.

carolyncole:
Another option I had to research is to still read the svg out of a file...
https://www.seancdavis.com/posts/render-inline-svg-rails-middleman/
I'm not suggesting this PR needs to change, just putting this documentation out there if we regret putting the SVG in rails code.

tpendragon (approved):

tpendragon:
CI's failing, but besides that 👍

barmintor:
I'm thinking about @hackartisan's questions yesterday - if we made the new components simple extensions of `Blacklight::BaseComponent`, could the svg source be a sidecar/template file? That would keep the svg out of the ruby itself. Not sure if svg is one of the ActionView template handlers, but that would be super easy to test.

PR #2673, opened 2022-03-18
Stop using SearchBarPresenter

jcoyne:
Now that we have view components the SearchBarPresenter is unnecessary, so it is now deprecated.


barmintor:
I think this is a good idea, but should we do this on 7.x and just remove it on main?

jcoyne:
@barmintor I think just removing on main is fine. I'm not sure how you deprecate this.

barmintor (approved):

barmintor:
Waiting on CI; backported to `release-7.x` with deprecations in #2700

PR #2674, opened 2022-03-18
Enable new rubocop rules and autocorrect

jcoyne:


app/presenters/blacklight/thumbnail_presenter.rb
```diff
@@ -72,7 +72,7 @@ def default_thumbnail_value(image_options)
     end
 
     def thumbnail_value_from_document
-      Array(thumbnail_field).lazy.map { |field| retrieve_values(field_config(field)).first }.reject(&:blank?).first
+      Array(thumbnail_field).lazy.map { |field| retrieve_values(field_config(field)).first }.compact_blank.first
```
carolyncole, line 75: @jcoyne looks like this line causes an issue with rails 6.  Can you reasonably bump the rails 6 version to 6.1.3.1 to grab it? (https://apidock.com/rails/Enumerable/compact_blank) or disable the rubocop check and go back to the original code?

app/models/record_mailer.rb
```diff
@@ -31,6 +31,6 @@ def sms_record(documents, details, url_gen_params)
     @config         = details[:config]
     @url_gen_params = url_gen_params
 
-    mail(to: details[:to], subject: "")
+    mail(to: details[:to], subject: "") # rubocop:disable Rails/I18nLocaleTexts
```
carolyncole, line 34: 👍 I never knew you could do one line disables this way

carolyncole (changes requested):
@jcoyne looks like worthwhile updates.  Just a small issue with compact_blank causing some of the test to fail with Rails 6.

app/presenters/blacklight/thumbnail_presenter.rb
jcoyne: Yeah, according to https://github.com/rspec/rspec-rails#rspec-rails-- we should be using version 5.x if we need to support Rails 6.0

barmintor:
Re-requesting from @carolyncole since @jcoyne made the requested changes. If we're bumping the linter's target Rails version, are we also bumping the gemspec Rails spec?

barmintor:
Rebased on `main`

barmintor (approved):

carolyncole (approved):

carolyncole:
@barmintor @jcoyne sorry, I checked back a couple times and then forgot to look again...

PR #2676, opened 2022-03-18
Stop testing on ruby 2.6

jcoyne:
Ruby 2.6 is no longer supported. It no longer receives security updates.  https://www.ruby-lang.org/en/news/2021/11/24/ruby-2-6-9-released/

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [2.6]
+        ruby: [2.7]
     steps:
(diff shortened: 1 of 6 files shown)


cbeer:
Should we also bump https://github.com/projectblacklight/blacklight/blob/main/blacklight.gemspec#L26?

cbeer:
Should we also bump https://github.com/projectblacklight/blacklight/blob/main/blacklight.gemspec#L26?

jcoyne:
I think that makes sense. Then we can use new language features.

barmintor (approved):
gemspec: check
testing matrix: check
AllCops/TargetRubyVersion: check
👍

PR #2679, opened 2022-03-24
Extract common Blacklight::Component base class for blacklight components 

cbeer:


jcoyne (approved):

jcoyne:
It looks like there's an upstream fix for this: https://github.com/github/view_component/pull/1317  Should we wait it out?

jcoyne:
The fix is released now: https://github.com/github/view_component/blob/main/docs/CHANGELOG.md#2520

lib/blacklight/component.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Blacklight
+  class Component < ViewComponent::Base
+    def render_in(view_context, &block)
```
jcoyne, line 5: Can you remove this?

jcoyne (changes requested):

lib/blacklight/component.rb
cbeer: Done

jcoyne (approved):

PR #2686, opened 2022-04-14
align early return condition and render logic in Blacklight::RenderConstraintsHelperBehavior

barmintor:
use `search_state.filters` for both

### DIFF
--- a/app/helpers/blacklight/render_constraints_helper_behavior.rb
+++ b/app/helpers/blacklight/render_constraints_helper_behavior.rb
@@ -84,14 +84,12 @@ module Blacklight::RenderConstraintsHelperBehavior
     Deprecation.warn(Blacklight::RenderConstraintsHelperBehavior, 'render_constraints_filters is deprecated')
     search_state = convert_to_search_state(params_or_search_state)
 
-    Deprecation.silence(Blacklight::SearchState) do
-      return "".html_safe if search_state.filter_params.blank?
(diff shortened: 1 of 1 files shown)


app/helpers/blacklight/render_constraints_helper_behavior.rb
```diff
@@ -84,14 +84,12 @@ def render_constraints_filters(params_or_search_state = search_state)
     Deprecation.warn(Blacklight::RenderConstraintsHelperBehavior, 'render_constraints_filters is deprecated')
     search_state = convert_to_search_state(params_or_search_state)
 
-    Deprecation.silence(Blacklight::SearchState) do
-      return "".html_safe if search_state.filter_params.blank?
+    return "".html_safe unless search_state.filters.any?
```
jcoyne, line 87: Is it possible that this is a breaking change?  Could people have overridden `filter_params` expecting it will be called?

barmintor: I was worried about that, too. I'm trying to imagine, tho, what the scenario might be, since the actual output already relies on `search_state.filters` and not `filter_params` - that is, there is drift between the exit condition and the output right now. If this entire method is overridden, that's another thing.

barmintor: I guess an implementation might have overridden `filter_params` to have a mutating side-effect but that's a thorny path to go down.

PR #2687, opened 2022-04-15
7.x delegate normalization to filters

barmintor:
This is what I was thinking re: using filter field definitions to sanitize the params.


barmintor:
@cbeer I needed to make an addition to the `SearchState::FilterField` interface to really get this to work - but I think it's one that would make writing plugins more reliable?

jcoyne (approved):

jcoyne:
Does this need a forward port to main?

barmintor:
Would recommend - am working on the backport of #2679 and resolving the question of #2647, then will turn to forward port of this change and #2647?

PR #2690, opened 2022-04-25
Extract a base component for blacklight

barmintor:
Backports #2679 to 7.x

### DIFF
--- a/app/components/blacklight/constraints_component.rb
+++ b/app/components/blacklight/constraints_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Blacklight
-  class ConstraintsComponent < ::ViewComponent::Base
+  class ConstraintsComponent < Blacklight::Component
     include Blacklight::ContentAreasShim
 
     renders_many :query_constraints_area
--- a/lib/blacklight/component.rb
+++ b/lib/blacklight/component.rb
@@ -2,5 +2,39 @@
 
 module Blacklight
   class Component < ViewComponent::Base
+    class << self
+      # rubocop:disable Naming/MemoizedInstanceVariableName
+      def compiler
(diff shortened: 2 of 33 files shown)


barmintor:
Since we've made a decision on `main` regarding base components, I've decided to take the more conservative approach to overrides and substitute in a `ViewComponent::Compiler` subclass for descendants of `Blacklight::Component` in this PR.

This allows simple overrides of component templates that exist in the engine by placing them in the same place in the installing application. If a file in the installing application doesn't have a same-named analog in the engine, it won't be recognized - this has an impact on translation sidecars if the installing app is more expansive than the engine in language scope. But the Compiler interface is more stable, and if `ViewComponent` commits to an interface for sidecar files we can revisit without disrupting existing apps.

app/components/blacklight/constraints_component.rb
```diff
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 
 module Blacklight
-  class ConstraintsComponent < ::ViewComponent::Base
-    include Blacklight::ContentAreasShim
```
jcoyne, line 5: Was this exclusion intentional?

barmintor: Added a commit, will squash after review

lib/blacklight/component.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Blacklight
+  class Component < ViewComponent::Base
+    class << self
```
jcoyne, line 5: Please add documentation around what this is doing and why.

jcoyne (changes requested):

lib/blacklight/component.rb
barmintor: Documentary comments added.

PR #2694, opened 2022-04-26
Port Blacklight::SearchState only manages permitted params to main

### LINKED ISSUE #2691: Port FilterField-base param sanitization forward to main
Port #2687 forward to main

barmintor:
- FilterFields are responsible for normalization of filter field data from params
- Base permitted search params are configurable in Blacklight:Configuration
- class methods for param normalization are deprecated

Co-authored-by: Justin Coyne <jcoyne@justincoyne.com>

Fixes #2691 
This is a forward-port of #2687 to main

### DIFF
--- a/lib/blacklight/search_state.rb
+++ b/lib/blacklight/search_state.rb
@@ -14,30 +14,67 @@ module Blacklight
 
     delegate :facet_configuration_for_field, to: :blacklight_config
 
+    def self.modifiable_params(params)
+      if params.respond_to?(:to_unsafe_h)
+        # This is the typical (not-ActionView::TestCase) code path.
+        params = params.to_unsafe_h
+        # In Rails 5 to_unsafe_h returns a HashWithIndifferentAccess, in Rails 4 it returns Hash
+        params = params.with_indifferent_access if params.instance_of? Hash
+      elsif params.is_a? Hash
+        # This is an ActionView::TestCase workaround for Rails 4.2.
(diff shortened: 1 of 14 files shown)


lib/blacklight/search_state.rb
```diff
@@ -14,30 +14,67 @@ class SearchState
 
     delegate :facet_configuration_for_field, to: :blacklight_config
 
+    def self.modifiable_params(params)
+      if params.respond_to?(:to_unsafe_h)
+        # This is the typical (not-ActionView::TestCase) code path.
+        params = params.to_unsafe_h
+        # In Rails 5 to_unsafe_h returns a HashWithIndifferentAccess, in Rails 4 it returns Hash
+        params = params.with_indifferent_access if params.instance_of? Hash
+      elsif params.is_a? Hash
```
jcoyne, line 23: I think this clause was already removed from the main branch, so we don't want to reintroduce it.

jcoyne (changes requested):
One small change requested, then it's good to go.

lib/blacklight/search_state.rb
barmintor: Thanks for catching that! Added a commit.

barmintor:
Rebased on main

barmintor:
And fixed a linting problem.

jcoyne (approved):

PR #2696, opened 2022-04-28
Pivot filter field

barmintor:
This is essentially implementing pivot fields as if they were a filter add-on, like range limits.

See also the conversation in #2464 , this permits use of pivot fields that are not individually configured as facets. I've included @dkinzer's tests from the other PR to make sure we have consensus about desired outcomes. 

Fixes: #2463

### DIFF
--- a/app/presenters/blacklight/facet_item_pivot_presenter.rb
+++ b/app/presenters/blacklight/facet_item_pivot_presenter.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module Blacklight
+  class FacetItemPivotPresenter < FacetItemPresenter
+    ##
+    # Check if the query parameters have the given facet field with the
+    # given value.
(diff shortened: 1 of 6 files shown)


barmintor:
This PR does *not* remove the pivot behaviors from`Blacklight::SearchState::FilterField`, even though they are not used in this branch.

dkinzer:
@barmintor I'm going to see if I can test this solution on our current app.  But, the app is 7.x is this a solution that would work in the 7.x release?

tpendragon:
As per Committer's Call, we're marking this ready for review. @dkinzer is taking on rebasing this so CI will pass, and has affirmed that this would solve his problem.

dkinzer:
After rebase, there are some tests that are failing. So I'm putting this back to draft for now.

dkinzer (approved):

PR #2698, opened 2022-05-02
Use collection parameters in line with idiomatic ViewComponent behaviors

### LINKED ISSUE #2697: Component counter variables should conform to ViewComponent iteration idioms
This is an issue that cuts across some open PRs (#2596 #2304) - ViewComponent provides a `#{x}_counter` mechanism for rendering collections of components, and our existing `document_counter` logic doesn't conform to it (when ViewComponent provides the parameter it will be 1-offset; the existing DocumentComponent assignment assumes zero-offset).

barmintor:
Remove redundant `with_collection_parameter :document` on `Blacklight::DocumentComponent` (from #2596)
Use `#{x}_counter` keyword params in the idiomatic manner for ViewComponent
- remove _document.html.erb
- fix #2697 
Use collection rendering of document_component in _document_list.html.erb

### DIFF
--- a/app/views/catalog/_document.atom.builder
+++ b/app/views/catalog/_document.atom.builder
@@ -21,7 +21,7 @@ xml.entry do
   with_format(:html) do
     xml.summary "type" => "html" do
       document_component = blacklight_config.view_config(:atom).summary_component
-      xml.text! render document_component.new(presenter: document_presenter(document), component: :div, show: true)
+      xml.text! render document_component.new(document_component.collection_parameter => document_presenter(document), component: :div, show: true)
     end
   end
 
--- a/app/views/catalog/_document_list.html.erb
+++ b/app/views/catalog/_document_list.html.erb
@@ -1,5 +1,6 @@
 <% # container for all documents in index list view -%>
 <% view_config = local_assigns[:view_config] || blacklight_config&.view_config(document_index_view_type) %>
 <div id="documents" class="documents-<%= view_config&.key || document_index_view_type %>">
-  <%= render documents, as: :document, view_config: view_config %>
+  <% document_presenters = documents.map { |doc| document_presenter(doc) } -%>
+  <%= render view_config.document_component.with_collection(document_presenters, partials: view_config.partials, counter_offset: @response&.start || 0) %>
 </div>
--- a/app/components/blacklight/document_component.rb
+++ b/app/components/blacklight/document_component.rb
@@ -56,8 +56,6 @@ module Blacklight
     # Backwards compatibility
     renders_one :actions
 
-    with_collection_parameter :document
-
     # rubocop:disable Metrics/ParameterLists
     # @param document [Blacklight::Document]
     # @param presenter [Blacklight::DocumentPresenter]
--- a/app/components/blacklight/document_component.rb
+++ b/app/components/blacklight/document_component.rb
@@ -57,35 +57,32 @@ module Blacklight
     renders_one :actions
 
     # rubocop:disable Metrics/ParameterLists
-    # @param document [Blacklight::Document]
-    # @param presenter [Blacklight::DocumentPresenter]
+    # @param document [Blacklight::DocumentPresenter]
+    # @param partials [Array, nil] view partial names that should be used to provide content for the `partials` slot
     # @param id [String] HTML id for the root element
     # @param classes [Array, String] additional HTML classes for the root element
     # @param component [Symbol, String] HTML tag type to use for the root element
     # @param title_component [Symbol, String] HTML tag type to use for the title element
     # @param counter [Number, nil] a pre-computed counter for the position of this document in a search result set
-    # @param document_counter [Number, nil] alternatively, the document's position in a collection and,
-    # @param counter_offset [Number] with `document_counter`, the offset of the start of that collection counter to the overall result set
+    # @param document_counter [Number, nil] a 1-indexed value provided by ViewComponent collection iteration
+    # @param counter_offset [Number] the offset of the start of the collection counter parameter for the component to the overall result set
     # @param show [Boolean] are we showing only a single document (vs a list of search results); used for backwards-compatibility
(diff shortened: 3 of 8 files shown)


lib/blacklight/component.rb
```diff
@@ -8,6 +8,10 @@ def compiler
         @__vc_compiler ||= EngineCompiler.new(self)
       end
       # rubocop:enable Naming/MemoizedInstanceVariableName
+
+      def with_individual(object, **kwargs)
```
jcoyne, line 12: I'm not sure I understand why we want or need this method.  This makes the logic more challenging to follow.  Why not just use the collection parameter outrightright?

barmintor: I'm basically trying to accommodate a configured document component called e.g. `DocComponent` with idiomatic parameters (`doc:, ...`) being usable here without overriding the calling partial.

barmintor: If we require `presenter:` and never take `document:` we will be working against ViewComponent's iteration (at least without specifiying the collection parameter, and then we're back to square one on subclasses). I think we want to strive to be as "normal" as possible about the ViewComponent stuff, and be able to rely on their documentation. But I'm also open to ideas about how to do this!

app/views/catalog/_document_list.html.erb
```diff
@@ -1,5 +1,11 @@
 <% # container for all documents in index list view -%>
 <% view_config = local_assigns[:view_config] || blacklight_config&.view_config(document_index_view_type) %>
 <div id="documents" class="documents-<%= view_config&.key || document_index_view_type %>">
-  <%= render documents, as: :document, view_config: view_config %>
+  <% view_config.document_component.with_collection(documents, counter_offset: (@response&.start || 0), presenter: Proc.new { |doc| document_presenter(doc) }).each do |component| %>
```
jcoyne, line 4: Why is presenter a proc and not just a reference?

barmintor: (it's a reference in non-collection contexts)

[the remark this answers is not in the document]
barmintor: We could definitely do something like that generally - say that we really expect `document:` to be a DocumentPresenter. We might then want to *require* that `presenter:` be a proc, and *only* be passed if `document:` is not a presenter.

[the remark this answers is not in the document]
barmintor: Added a commit to defer the presenter resolution in the component, which cleans this partial up a bit.

app/views/catalog/_document.atom.builder
```diff
@@ -21,7 +21,7 @@ xml.entry do
   with_format(:html) do
     xml.summary "type" => "html" do
       document_component = blacklight_config.view_config(:atom).summary_component
-      xml.text! render document_component.new(presenter: document_presenter(document), component: :div, show: true)
+      xml.text! render document_component.with_individual(document, presenter: document_presenter(document), component: :div, show: true)
```
jcoyne, line 24: Do we need to pass the document now?

app/views/catalog/_document_list.html.erb
```diff
@@ -1,5 +1,11 @@
 <% # container for all documents in index list view -%>
 <% view_config = local_assigns[:view_config] || blacklight_config&.view_config(document_index_view_type) %>
 <div id="documents" class="documents-<%= view_config&.key || document_index_view_type %>">
-  <%= render documents, as: :document, view_config: view_config %>
+  <% view_config.document_component.with_collection(documents, counter_offset: (@response&.start || 0)).each do |component| %>
```
jcoyne, line 4: why is this not passing a presenter?  It used to when it was here: https://github.com/projectblacklight/blacklight/pull/2698/files#diff-8225b3193b5c90c9d0e11c0ab59b930bcd01e527430e0a1fc795093271946cc5L3

barmintor: the presenter it passed before reproduced the default logic for the presenter in the component; if we are committing to having document component wrap the presenter, we would change this call but also probably remove the  presenter-default logic in the component; this seemed like the smaller change.

app/views/catalog/_document.atom.builder
barmintor: Pushed a commit removing `with_individual` and passing only the presenter here.

app/views/catalog/_document_list.html.erb
jcoyne: I wonder if we can hash this out on the next call.  I feel like making the component wrap the presenter, and have no knowledge of the document (fewer dependencies) is the better design.

barmintor: I don't disagree with that - I might suggest something similar to that `with_individual` method I removed in that case, maybe just to build a splattable options hash-- Just because it feel weird to say that DocumentPresenter's `document:` param should be a presenter. I just didn't want to be presumptuous here outside the scope of using unsurprising collection behaviors for the components.

barmintor: Like, I would be totally happy to open a ticket saying `DocumentComponent should expect a document presenter, not a SolrDocument` and do that next! It would let us delete a good amount of code in the DocumentComponent.

tpendragon (changes requested):

tpendragon:
@barmintor Would you like another committer to pick up this refactor?

barmintor:
@tpendragon yes, if possible

barmintor:
@tpendragon @jcoyne ok, that refactor is done

barmintor:
After the rebase and these changes, `_document.html.erb` doesn't really do anything unless an installing application was overriding it. Should it be deleted?

cbeer:
I think `catalog/document` is still a useful seam (for doing e.g. `render document`, if nothing else) 🤷‍♂️

barmintor:
thanks for everyone's patience while I worked that ViewComponent 3 issue out!

tpendragon (approved):

PR #2699, opened 2022-05-02
Leverage ActiveSupport::Concern for Blacklight::User

jcoyne:
This is the Rails idiom for "concerns" See: https://api.rubyonrails.org/v7.0.2.3/classes/ActiveSupport/Concern.html


barmintor (approved):
Idiomatic rails 👍

PR #2708, opened 2022-05-06
Remove the _previous_next_doc.html.erb partial

jcoyne:
Observations have been made about how challenging it is to follow program execution through various partials, helpers and components.  I identified that the _previous_next_doc.html.erb partial does nothing but render a component. We can remove this partial and make it easier for people reading the code to follow the control flow.

This partial is not used by spotlight, blacklight-range_limit, or blacklight-advanced_search.


barmintor:
Seems like a good change to me, so I will ask my question: Should we remove the partial entirely on `main` and backport the deprecation to `release-7.x`?

jcoyne:
@barmintor  I have no preference on that matter.

barmintor (approved):
Opened #2711 to deprecate in 7.x

PR #2709, opened 2022-05-06
Remove render_document_main_content_partial helper

jcoyne:
All this method does is call a partial. To make the program flow easier to understand, we can just render the partial and avoid the helper call.

I don't see evidence of any open source repo on github doing anything differently with this helper: 
https://github.com/search?q=%22def+render_document_main_content_partial%22&type=Code


barmintor (approved):
Backported deprecation to 7.x in #2712

PR #2712, opened 2022-05-07
7.x Deprecate render_document_main_content_partial helper

barmintor:
All this method does is call a partial

To make the program flow easier to understand, we can just render the partial and avoid the helper call.

The method should be removed in 8.0; see #2709


barmintor:
Folding this into #2711

PR #2715, opened 2022-05-09
render_search_bar should delegate to SearchBarComponent until it is removed in 8.0

barmintor:
The deprecation of SearchBarPresenter introduced and breaking change to apps that call `render_search_bar` from a partial. This method is deprecated and slated for removal, but this change will allow 7.x apps that call it to otherwise work as before.

### DIFF
--- a/app/helpers/blacklight/blacklight_helper_behavior.rb
+++ b/app/helpers/blacklight/blacklight_helper_behavior.rb
@@ -79,7 +79,14 @@ module Blacklight::BlacklightHelperBehavior
   # Render the search navbar
   # @return [String]
   def render_search_bar
-    search_bar_presenter.render
+    component_class = blacklight_config&.view_config(document_index_view_type)&.search_bar_component || Blacklight::SearchBarComponent
+    component_class.new(
+      url: search_action_url,
+      advanced_search_url: search_action_url(action: 'advanced_search'),
+      params: search_state.params_for_search.except(:qt),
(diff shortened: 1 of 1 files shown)


barmintor:
This ensures that the app dpwnstream does not have to sort out a replacement for `search_fields` yet if it is using this helper.

app/helpers/blacklight/blacklight_helper_behavior.rb
```diff
@@ -79,7 +79,14 @@ def extra_body_classes
   # Render the search navbar
   # @return [String]
   def render_search_bar
-    search_bar_presenter.render
```
cbeer, line 82: I'm worried about the backwards-compatible implications of skipping the presenter here. I think downstream applications could have overridden that or the `catalog/search_form` partial.

cbeer, line 82: I'm worried about the backwards-compatible implications of skipping the presenter here. I think downstream applications could have overridden that or the `catalog/search_form` partial.

barmintor: How would you feel about checking the class of the presenter, and dispatching to the subclass if it is not the default?

barmintor: Added a commit that falls back to the deprecated behavior if the presenter class is not the default, or if the partial does not appear to be the default partial.

jcoyne (approved):

PR #2720, opened 2022-05-12
add a proc mechanism to Blacklight::Configuration#property to allow deferred resolution of some default values

### LINKED ISSUE #2705: Blacklight::DocumentFactory class is not loaded when Blacklight::Configuration property definitions are set

barmintor:
- if property default values are not defined under lib, they may not be available until application is loaded, causing issues for downstream apps running generators
- fixes #2705


barmintor:
I'm not enamored of this mechanism, but... it works. Without something like this, I was left using `nil` for the default and overriding accessors, which appears to be the 7.x approach we want to move away from.

barmintor:
Closed as was #2705

PR #2757, opened 2022-06-30
Facet rendering should fall back on a null facet field response

cbeer:


jcoyne (changes requested):
Would you please add a test and a bit of documentation explaining why this is useful?

jcoyne (approved):

.rubocop_todo.yml
```diff
@@ -133,7 +133,7 @@ Metrics/ParameterLists:
 # Offense count: 36
 # Configuration parameters: IgnoredMethods.
 Metrics/PerceivedComplexity:
-  Max: 14
+  Max: 15
```
jcoyne, line 136: is this necessary since you added the disable in the source code?

PR #2771, opened 2022-07-13
yaml config exposed on ActiveRecord::Base

barmintor:


cbeer:
Looks like Rails exposes this on different classes in Rails 6 vs Rails 7:

https://github.com/rails/rails/blob/v6.1.6.1/activerecord/lib/active_record/coders/yaml_column.rb#L49
https://github.com/rails/rails/blob/v7.0.3.1/activerecord/lib/active_record/coders/yaml_column.rb#L49

PR #2775, opened 2022-07-14
Try to be friendly to application with action_on_unpermitted_paramete…

cbeer:
…rs = :raise set


tampakis:
@jrochkind take a look at this when you get a chance and comment on here if this proposed change helps you out.

PR #2777, opened 2022-07-17
Fix Solr 9 modules load path

cbeer:


jcoyne (approved):

epugh:
i think lines 21 and 22 could be deleted as well.....

cbeer:
Only if we wanted to drop support for Solr 8, though, right?

epugh:
oh right...  that is an interesting point that you *can* do both via what you did...  cool...

PR #2780, opened 2022-07-20
Use `$border-color` when we're outside of a table context

cbeer:
This allows the css to compile with bootstrap 5.2.

Part of #2781


marlo-longley (approved):
Thanks Chris!

PR #2788, opened 2022-07-24
Add a component parameter to the FacetComponent, allowing callers to …

cbeer:
…override the configured value.


app/components/blacklight/facet_component.rb
```diff
@@ -30,9 +30,10 @@ def initialize(display_facet_or_field_config: nil, display_facet: nil, field_con
         raise ArgumentError, 'You must provide one of display_facet or field_config' unless @field_config
       end
 
+      @component = component
```
jcoyne, line 33: Can this be `@component = component || @filed_config.component`? Then we remove the conditional on line 46?

jcoyne (approved):

PR #2792, opened 2022-07-25
Silence deprecation on search_fields within the blacklight code

hackartisan:
This silences deprecation warnings downstream.
refs https://github.com/projectblacklight/blacklight/commit/009a4266771eab2c5713bc54a4c7996ef803ed5a

### DIFF
--- a/app/views/catalog/index.json.jbuilder
+++ b/app/views/catalog/index.json.jbuilder
@@ -67,14 +67,16 @@ json.included do
     end
   end
 
-  json.array! search_fields do |(label, key)|
-    json.type 'search_field'
-    json.id key
-    json.attributes do
-      json.label label
-    end
-    json.links do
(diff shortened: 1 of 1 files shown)


app/views/catalog/index.json.jbuilder
```diff
@@ -67,6 +67,10 @@ json.included do
     end
   end
 
+  search_fields = blacklight_config.search_fields.values.
```
cbeer, line 70: I think it's possible downstream applications have overridden the `search_fields` helper, so the best we can do is wrap it in `Deprecation.silence(Blacklight::ConfigurationHelperBehavior)`?

hackartisan: Good point! Wrap this bit of code then? I will try that.

hackartisan:
@cbeer how's this? silences deprecations when I try it with my app.

PR #2804, opened 2022-08-16
Expand on document attributes to make them more expressive and capture some additional use cases

cbeer:
This also better aligns the `attribute` class method with upstream conventions in Rails (i.e. `default`, type-as-symbol, etc)

### DIFF
--- a/app/models/concerns/blacklight/document.rb
+++ b/app/models/concerns/blacklight/document.rb
@@ -21,6 +21,7 @@ module Blacklight::Document
   include Blacklight::Document::SemanticFields
   include Blacklight::Document::CacheKey
   include Blacklight::Document::Export
+  include Blacklight::Document::Attributes
 
   included do
     extend ActiveModel::Naming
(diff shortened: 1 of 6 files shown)


app/models/concerns/blacklight/document.rb
```diff
@@ -109,21 +117,5 @@ def more_like_this
     def unique_key
       @unique_key ||= 'id'
     end
-
```
barmintor, line 112: This method was extracted into a concern

sandbergja:
Thanks for this PR, @cbeer !  It looks good, and we (the people on the 19 October Committers Call) like that it aligns closer to the rails API.  The PR name mentions "additional use cases".  Could you please let us know which additional use cases that this PR enables?  Any documentation that you could point us toward?

Also, it seems that the usage examples in the attributes.rb don't seem to match the usage in the spec -- could you please update?

jcoyne (approved):

app/models/concerns/blacklight/document/attributes.rb
```diff
@@ -0,0 +1,47 @@
...
+    class_methods do
+      # Define an attribute reader on a document model
+      # @param [Symbol] name the name of the attribute to define
+      # @param [Symbol, Class] type the type of the attribute to define
+      # @param [String] field the name of the document's field to use for this attribute
+      # @param [any, Proc] default the default value for the attribute
+      # @example
+      #   class SolrDocument
+      #     include Blacklight::Solr::Document
+      #     attribute :title, Blacklight::Types::String, 'title_tesim'
```
jcoyne, line 23:
```suggestion
      #     attribute :title, :string, 'title_tesim'
      #
      #     Deprecated syntax:
      #     attribute :title, Blacklight::Types::String, 'title_tesim'
```

PR #2817, opened 2022-09-28
Allow missing view icons

cbeer:
Instead of blowing up with an error, we can just fall back on the view name:

Screen Shot 2022-09-28 at 14 43 21


jcoyne:
I believe this requires a rebase.

jcoyne (approved):

PR #2821, opened 2022-10-03
Don't let turbo interfere with our modal behavior

cbeer:


jcoyne:
I don't understand where this is a problem. Can you give an example?

jcoyne:
Oh, is this because we have 2 javascript functions binding to "onclick" here?

cbeer:
I guess? None of the content that opens in modals work at all if turbo is installed in the app.

jcoyne (approved):

PR #2828, opened 2022-10-07
Remove dialog-polyfill

jcoyne:
It's no longer necessary to polyfill as it's supported by all browsers: https://caniuse.com/dialog


sandbergja (approved):
Thanks, @jcoyne !

PR #2833, opened 2022-10-11
Document component extension points

jcoyne:


jrochkind (approved):

jcoyne:
these are just the extension points that we are using in arclight. There may be some others that should be discovered and added in the same way.

cjcolvar (approved):
This looks like a good starting point which can be expanded upon.

PR #2837, opened 2022-10-11
update rubocop and rubocop-rspec

jcoyne:


thatbudakguy:
fyi — if you split out the changes to `.rubocop.yml`, `.rubocop_todo.yml`, and the `Gemfile` from the actual running of rubocop, then you can add the hash of the "run rubocop" commit to [a `.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) and commit it along with this PR so that all the rubocop changes don't interfere with `git blame`.

jcoyne:
@thatbudakguy I would prefer if we can use git blame to see that these changes came from running rubocop when doing git blame.

thatbudakguy:
ok — I'm ambivalent, I guess, but it would be good to get somebody else's input before merging this

sandbergja (approved):
Thanks, @jcoyne !  The `.git-blame-ignore-revs` feature seems interesting, but I don't think it's necessary.

PR #2843, opened 2022-10-13
Restore the wrapping blacklight-icons span to componentized icons

cbeer:
In #2670 , we added support for component-supplied icons to avoid a hard dependency on sprockets (which... turns out not to be going away,  🤷‍♂️ ). I'm not sure why the wrapping `.blackight-icons` span was dropped, but it's heavily used in e.g. arclight, so it seems like it'd be nice to bring back (and allow us to backport icon components to 7.x)

### DIFF
--- a/app/components/blacklight/icons/icon_component.rb
+++ b/app/components/blacklight/icons/icon_component.rb
@@ -6,15 +6,27 @@ module Blacklight
     # You can override the default svg by setting:
     #   Blacklight::Icons::ListComponent.svg = '<svg>your SVG here</svg>'
     class IconComponent < ::ViewComponent::Base
-      def initialize(svg: nil)
+      # rubocop:disable Metrics/ParameterLists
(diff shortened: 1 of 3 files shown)


bess (approved):

app/components/blacklight/icons/icon_component.rb
```diff
@@ -6,15 +6,27 @@ module Icons
     # You can override the default svg by setting:
     #   Blacklight::Icons::ListComponent.svg = '<svg>your SVG here</svg>'
     class IconComponent < ::ViewComponent::Base
-      def initialize(svg: nil)
+      # rubocop:disable Metrics/ParameterLists
+      def initialize(svg: nil, component: :span, name: nil, label: nil, aria_hidden: nil, classes: '', **options)
```
jcoyne, line 10: I think `component` here is confusing as it's not a view component.  How about calling it `tag:` or `wrapper:` instead?

jcoyne (changes requested):

jcoyne (approved):

PR #2856, opened 2022-10-28
Style the autocomplete component

jcoyne:


corylown (changes requested):
This change is causing the li option values to overflow the ul listbox instead of wrapping:
Screen Shot 2022-11-04 at 3 33 08 PM

jcoyne:
@corylown I think this is still an improvement over the current styles. I'm not really sure how to fix it aside from `overflow: clip`. Do you have any suggestions?

jcoyne:
@corylown ah, I pulled off the `width: 100%` off the `ul` and I think that maybe resolves this.

corylown (approved):

PR #2861, opened 2022-10-31
Visually hide search button text on xs devices

thatbudakguy:
* Create visually-hidden-* helpers using bootstrap 5 API
* Add bootstrap 4 support using @sr-only
* Apply visually-hidden-xs to search button text
* Adjust padding in button slightly

This helps with the lack of space on very small screens. It seems like there were already styles to do this, but they aren't working at the moment (maybe markup has changed since they were written?)

### DIFF
--- a/app/assets/stylesheets/blacklight/_header.scss
+++ b/app/assets/stylesheets/blacklight/_header.scss
@@ -40,18 +40,6 @@
     @extend .col-lg-8;
     padding-left: 0;
   }
-
-  .submit-search-text {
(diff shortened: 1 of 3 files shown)


thatbudakguy:
ping @jrochkind since I got the sense you'd been down this path before, and I actually ran into your issue https://github.com/twbs/bootstrap-sass/issues/498 from more than 8 years ago about this very problem!

jcoyne (approved):

PR #2864, opened 2022-11-01
Add a slot for rendering extra inputs in the search bar

thatbudakguy:
Refs #2862 and #2863


jcoyne:
@thatbudakguy what do you think about this approach instead? https://github.com/projectblacklight/blacklight/pull/2866  We could then have our own search navbar component rather than adding an extension point to this one.  I think that we want to avoid adding arbitrary extension points to every component and encourage people to override components instead.

thatbudakguy:
I think #2866 is sensible (and might help us for the reasons you mention), but I don't think it solves the problem in arclight that this PR is intended to solve — or at least not fully, anyway.

Arclight currently renders:
```html
<div class="navbar-search navbar">
  <div class="container">
    <form class="search-query-form">
      <div class="input-group"> <!-- input group for the search input -->
        <div class="input-group"></div> <!-- input group for the extra dropdown arclight adds -->
      </div>
    </form>
  </div>
</div>
```

To make it more correct vis-a-vis bootstrap, and to solve issues like https://github.com/projectblacklight/arclight/issues/1239 that need the appropriate HTML structure, we instead want:
```html
<div class="navbar-search navbar">
  <div class="container">
    <form class="search-query-form">
      <div class="input-group"></div> <!-- input group for the extra dropdown arclight adds -->
      <div class="input-group"></div> <!-- input group for the search input -->
    </form>
  </div>
</div>
```

The issue is that blacklight's `SearchBarComponent` actually includes the entire `<form>`, so if you want to add more inputs to it, you need to either override the whole component, or use a slot. #2866 definitely helps solve the problem of adding more content above/below the search bar, but it doesn't address this narrower use case of making sure that more inputs end up (as siblings) inside the search form.

That said, I very much agree with:
> I think that we want to avoid adding arbitrary extension points to every component and encourage people to override components instead.

So maybe overriding the entire `SearchBarComponent` is the way to go regardless.

app/components/blacklight/search_bar_component.rb
```diff
@@ -5,6 +5,7 @@ class SearchBarComponent < Blacklight::Component
     renders_one :append
     renders_one :prepend
     renders_one :search_button
+    renders_many :inputs
```
cbeer, line 8: I don't love this slot name; `before_input_groups` is kinda clunky, but might express the intent better?

thatbudakguy: Yeah, I kind of punted on the naming. Being clear is a good idea 👍🏻

PR #2865, opened 2022-11-01
Allow the application link in the header to be configured

jcoyne:


lib/blacklight/configuration.rb
```diff
@@ -50,6 +50,7 @@ def initialized_default_configuration?
 
     # rubocop:disable Metrics/BlockLength
     default_configuration do
+      property :logo_link
```
cjcolvar, line 53: It looks like all of the other properties have defaults explicitly specified.  How about doing that here?
```suggestion
      property :logo_link, default: nil
```

thatbudakguy (approved):

PR #2866, opened 2022-11-02
Extract components for the headers

jcoyne:
This allows us to override the header component, while reusing existing components. For example, at Stanford we would like to insert a hidden third navbar that has a feedback form between the top bar and the search navbar.  This will allow us to override the header component to do that.


app/components/blacklight/search_navbar_component.html.erb
```diff
@@ -0,0 +1,6 @@
+
```
cbeer, line 1:
```suggestion
```

app/components/blacklight/header_component.rb
```diff
@@ -0,0 +1,23 @@
...
+module Blacklight
+  class HeaderComponent < Blacklight::Component
+    def initialize(blacklight_config:)
+      @blacklight_config = blacklight_config
+    end
+
+    attr_reader :blacklight_config
+
+    def call
+      render(top_bar) + render(search_bar)
```
cbeer, line 12: Maybe `top_bar` and `search_bar` could be slots with default values, like we've done in other components? It's a little easier to change them out that way.

app/views/layouts/blacklight/base.html.erb
```diff
@@ -31,17 +31,17 @@
       <%= link_to t('blacklight.skip_links.main_content'), '#main-container', class: 'element-invisible element-focusable rounded-bottom py-2 px-3', data: { turbolinks: 'false' } %>
       <%= content_for(:skip_links) %>
     </nav>
-    <%= render partial: 'shared/header_navbar' %>
+    <%= render blacklight_config.header_component.new(blacklight_config: blacklight_config) %>
```
cbeer, line 34: This is a pretty big extension point for a lot of applications; I wonder if it'd be nicer to downstream apps if we keep rendering `shared/header_navbar` and render the header component inside it?

PR #2868, opened 2022-11-03
Copy the config to each new component as it is created

### LINKED ISSUE #2867: ViewComponent 2.75.0 breaks blacklight
Screen Shot 2022-11-02 at 3 04 28 PM

jcoyne:
Fixes #2867

### DIFF
--- a/lib/blacklight/component.rb
+++ b/lib/blacklight/component.rb
@@ -3,6 +3,12 @@
 module Blacklight
   class Component < ViewComponent::Base
     class << self
+      def inherited(subclass)
+        # Workaround for https://github.com/ViewComponent/view_component/issues/1565
+        subclass.config = ViewComponent::Base.config
+        super
+      end
+
       # rubocop:disable Naming/MemoizedInstanceVariableName
       def compiler
         @__vc_compiler ||= EngineCompiler.new(self)
--- a/lib/blacklight/component.rb
+++ b/lib/blacklight/component.rb
@@ -3,10 +3,9 @@
 module Blacklight
   class Component < ViewComponent::Base
     class << self
-      def inherited(subclass)
-        # Workaround for https://github.com/ViewComponent/view_component/issues/1565
-        subclass.config = ViewComponent::Base.config
-        super
+      # Workaround for https://github.com/ViewComponent/view_component/issues/1565
+      def config
+        @config ||= ViewComponent::Config.defaults.merge(ViewComponent::Base.config)
       end
 
       # rubocop:disable Naming/MemoizedInstanceVariableName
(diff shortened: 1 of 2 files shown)


barmintor:
This is a workaround for a documented issue in ViewComponent.

barmintor:
There have been two breaking changes w/ ViewComponent minor releases: One was the promotion of an experimental API (which we were using) to actual API (hurrah!); this one is a bug related to upstream changes: https://github.com/ViewComponent/view_component/issues/1565

cbeer:
I wonder if we should use (or default to?) `ViewComponent::Base.config.view_component_path` instead of `component_class.view_component_path`? I'm not sure whether changing the config per-class is a feature?

lib/blacklight/component.rb
```diff
@@ -3,6 +3,12 @@
 module Blacklight
   class Component < ViewComponent::Base
     class << self
+      def inherited(subclass)
+        # Workaround for https://github.com/ViewComponent/view_component/issues/1565
+        subclass.config = ViewComponent::Base.config
```
barmintor, line 8: This obviously works, but it also means that you can't actually configure component classes in any way that varies from `Base.config`. This seems like it might not be desirable. The ViewComponent hooks initiailize to the `config.defaults`... also seems wrong in cases outside the Base class.

jcoyne: should we dup it?

lib/blacklight/component.rb
```diff
@@ -3,6 +3,12 @@
...
+        super
+      end
```
barmintor, line 10:
```suggestion
      # Workaround for https://github.com/ViewComponent/view_component/issues/1565
      def config
        @config ||= ViewComponent::Config.defaults.merge(ViewComponent::Base.config)
      end
```

barmintor: I don't love this, either - but it's a safe copy that would capture the base config. It doesn't address the question for me of whether this config's entry point should be in `config.blacklight.view_component` or what.

barmintor: I think this also means we have to pin ViewComponent to `~> 2.65`

jcoyne:
Changed to @barmintor's implementation

barmintor:
@jcoyne can you add a pin to `  s.add_dependency "view_component", '~> 2.66'` in the gemspec?

barmintor:
@cbeer that's a good question, too - though in this case, `Base.config` is empty, so we could approach as a separate issue?

barmintor (approved):
This seems like a reasonable workaround to to the upstream bug.

PR #2870, opened 2022-11-04
Move code from the template to the class

jcoyne:
This makes it easier to override bits of it without having to copy the whole template

### DIFF
--- a/app/components/blacklight/document/action_component.rb
+++ b/app/components/blacklight/document/action_component.rb
@@ -16,8 +16,26 @@ module Blacklight
         @link_classes = link_classes
       end
 
+      def render_control
+        return link_to_modal_control if using_default_document_action?
+
+        render_partial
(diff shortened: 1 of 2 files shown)


app/components/blacklight/document/action_component.rb
```diff
@@ -16,10 +16,28 @@ def initialize(document:, action:, options: {}, url_opts: {}, id: nil, link_clas
         @link_classes = link_classes
       end
 
+      def call
```
cbeer, line 19: By overriding `#call`, I think we're taking away the ability for action subclasses to use templates (because they'll inherit this `#call` method). Can we keep using the partial?

jcoyne: Sounds good. I've updated to keep the template.

corylown (approved):

PR #2872, opened 2022-11-07
The default inspect can be so long that it's hard to read

jcoyne:
This makes it short by default and allows the users to configure the displayed fields for SolrDocument

### DIFF
--- a/app/models/concerns/blacklight/document.rb
+++ b/app/models/concerns/blacklight/document.rb
@@ -26,6 +26,8 @@ module Blacklight::Document
     extend ActiveModel::Naming
     include Blacklight::Document::Extensions
     include GlobalID::Identification
+
+    class_attribute :inspector_fields, default: [:_source]
   end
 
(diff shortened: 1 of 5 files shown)


app/models/concerns/blacklight/document.rb
```diff
@@ -82,6 +84,11 @@ def first key
     Array(self[key]).first
   end
 
+  def inspect
+    fields = inspector_fields.map { |field| "#{field}: #{public_send(field)}" }.join(", ")
```
cbeer, line 88: I wonder if just `:id` is too limiting. I often use `inspect` to see what data we got back from solr (and... I'm not sure I've ever inspected a `SolrDocument` except to do that?). Would it be possible to include all of `_source` by default (and omit `response`, which I 💯  agree is unwieldy and easy to get at when it's useful)?

jcoyne: Sure, I changed the default to `_source`

PR #2881, opened 2022-11-09
Add X-CSRF-Token header to checkbox AJAX

### LINKED ISSUE #2871: Unclicking bookmark causes error
And the text next to the bookmark checkbox gets stuck at "Saving..."

Screen Shot 2022-11-04 at 17 24 16

Via @jcoyne , the error is caused by a CSRF error on the second submission because the CSRF token is out of date.

marlo-longley:
Fixes #2871 

This fixes the issue, but as I'm not super familiar with Rails CSRF tokens, not sure if this is a solid soution.


jcoyne (changes requested):
You've changed the compiled assets.  You want to change the source assets and then compile with `npm run js-compile-bundle`

jcoyne (changes requested):
Should we add ` authenticity_token: false` here? https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/document/bookmark_component.html.erb#L4  so that it's not confusing as to why we're no longer using the per-form token?

PR #2885, opened 2022-11-10
Extract a component for the show tools

jcoyne:


app/components/blacklight/document/show_tools_component.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="card show-tools">
+  <div class="card-header">
+    <h2 class="mb-0 h6"><%= t('blacklight.tools.title') %></h2>
+  </div>
+  <%= render Blacklight::Document::ActionsComponent.new(document: document,
```
cbeer, line 5: I think we might as well pull in the logic from ActionsComponent too. I guess I thought `ActionsComponent` was the interesting component seam, but if we're pulling the whole thing over...

jcoyne: Blacklight::Document::ActionsComponent is used in many other places, so I'm not sure how we would pull that in here.

cbeer: ActionsComponent isn't very interesting; it just wraps a set of actions with some markup... and that's what it seems like we're trying to do here too. Having multiple layers of wrapping feels like unnecessary indirection; if this component wants actions as a `ul` with `li > .nav-link` children, I think it would be easier to deal with if this template did something like:

```
<div class="card show-tools">
  <div class="card-header">
    <h2 class="mb-0 h6"><%= t('blacklight.tools.title') %></h2>
  </div>

  <ul classes="list-group list-group-flush">
    <% actions.each do |action| %>
      <%= tag.li(class: %w[list-group-item list-group-item-action] + [action.key]) do %>
        <%= action %>
      <% end %>
    <% end %>
  </ul>
</div>
```

I don't know if there's an opportunity to inherit from `ActionsComponent`, if we'd need to duplicate a couple a couple lines, or what, but I think that'd be a clearer seam for downstream 🤷‍♂️

barmintor:
I'm adding a commit by way of review that tries to respond to @cbeer 's questions.
`ShowToolsComponent` is in that commit a thin subclass of `ActionsComponent`, and `ActionComponent` is able to handle its own tag wrapping.

If that seems satisfactory, I think we should consider dropping the `show_tools` partial in 8.0, though we'd need to do some ugly deprecation on the 7.x branch.

barmintor:
Oh no, this needs a rebase!

jcoyne:
@barmintor rebased.

tpendragon (approved):

PR #2889, opened 2022-11-10
Extract a logo_link method on the TopNavbarComponent

jcoyne:
This allows you specify the logo text on the component if you don't want it to be the same as the `application_name`. This also gives you the flexibility to override the link method in a downstream component without having to copy the whole template.


marlo-longley:
Question: is supporting this in BL's topnav something we want to support? Does this solve any issues for other BL implementers?

barmintor (approved):

carolyncole (approved):

PR #2890, opened 2022-11-10
Improve the pivot facet display

cbeer:
Screen Shot 2022-11-10 at 14 57 57

### DIFF
--- a/app/assets/stylesheets/blacklight/_facets.scss
+++ b/app/assets/stylesheets/blacklight/_facets.scss
@@ -173,6 +173,11 @@
   }
 }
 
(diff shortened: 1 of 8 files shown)


ggeisler:
This is a minor question and not really related to the work in this PR, but I wonder would it be difficult to replace the icons? I know implementers can override them, but it seems like starting with more readable icons might be helpful. The current icons don't read as "expand/collapse" very well (I think for DLME I watched some recorded user sessions where it seemed unclear to the users that they could use them to expand). The ones I suggested in https://github.com/sul-dlss/vt-arclight/issues/184 are Bootstrap icons and would work here, maybe?

jcoyne (approved):

PR #2893, opened 2022-11-11
Expose the original, uncompiled assets in importmaps

cbeer:


jcoyne (approved):

lib/blacklight/engine.rb
```diff
@@ -39,7 +39,11 @@ class Engine < Rails::Engine
     end
 
     initializer "blacklight.importmap", before: "importmap" do |app|
-      app.config.importmap.paths << Engine.root.join("config/importmap.rb") if app.config.respond_to?(:importmap)
+      if app.config.respond_to?(:importmap)
```
jcoyne, line 42: Is this necessary? Doesn't the `before: "importmap"` imply this?

jcoyne (approved):
Approved pending consideration of comment.

lib/blacklight/engine.rb
cbeer: Does it? I'm not sure what happens if you set a `before` on something that doesn't exist.

PR #2895, opened 2022-11-14
Pass through additional engine cart options when building the test app

cbeer:
It looks like the parameter was renamed at some point during development..

### DIFF
--- a/lib/generators/blacklight/assets_generator.rb
+++ b/lib/generators/blacklight/assets_generator.rb
@@ -31,7 +31,7 @@ module Blacklight
       return if has_blacklight_assets?
 
       if Rails.version > '7'
-        gem "sassc-rails", "~> 2.1"
+        gem "sassc-rails", "~> 2.1" if defined?(Sprockets)
         append_to_file 'config/importmap.rb' do
           <<~CONTENT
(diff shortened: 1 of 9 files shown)


jcoyne (approved):

lib/generators/blacklight/assets_generator.rb
```diff
@@ -1,96 +1,21 @@
...
-
-    def has_blacklight_assets?
-      application_js.include?('blacklight/blacklight')
-    end
+    def run_asset_pipeline_specific_generator
+      generated_options = "--bootstrap-version='#{options[:'bootstrap-version']}'" if options[:'bootstrap-version']
 
-    def application_js
-      path = File.expand_path("app/assets/javascripts/application.js", destination_root)
+      generator = if defined?(Propshaft)
```
jcoyne, line 10: Propshaft  is something you can use with Importmap. Only Sprockets and Propshaft are mutually exclusive.  Do we need to deal with that case?

cbeer: Let's see if someone asks for it? I don't see a way to generate a new rails app that uses both, so it feels like an advanced/edge case configuration to me 🤷‍♂️

PR #2904, opened 2022-11-16
Prevent checkboxes from double rendering

### LINKED ISSUE #2860: Using forward and back buttons causes more bookmark controls
Screenshot 2022-10-31 at 4 53 32 PM

Only the last one added actually works to create the bookmark.

jcoyne:
Previously the use of turbo causes the transformed html to be cached. This causes duplicate checkboxes to appear when returning to a cached page.  Here we switch to a server rendered component to prevent the transformation step.

fixes #2860


jcoyne:
This blocks https://github.com/projectblacklight/arclight/issues/1267

PR #2905, opened 2022-11-16
Update documentation for Bookmarks#create

jcoyne:


app/controllers/concerns/blacklight/bookmarks.rb
```diff
@@ -63,11 +63,9 @@ def update
     create
   end
 
-  # For adding a single bookmark, suggest use PUT/#update to
-  # /bookmarks/$docuemnt_id instead.
-  # But this method, accessed via POST to /bookmarks, can be used for
-  # creating multiple bookmarks at once, by posting with keys
-  # such as bookmarks[n][document_id], bookmarks[n][title].
+  # For adding a single bookmark, suggest use PUT to /bookmarks/$document_id instead (triggering the #update method).
```
cbeer, line 66:
```suggestion
  # For adding a single bookmark, suggest use PUT to /bookmarks/:document_id instead (triggering the #update method).
```

PR #2907, opened 2022-11-17
Prevent checkbox from double rendering

jcoyne:
Previously the use of turbo causes the transformed html to be cached. This causes duplicate checkboxes to appear when returning to a cached page. Here we switch to a server rendered component to prevent the transformation step.

fixes https://github.com/projectblacklight/blacklight/issues/2860


cbeer:
Doesn't this prevent non-javascript users from being able to use bookmarks? I thought the current approach was intended to  support progressive enhancement.

jcoyne:
Yes, this would prevent non-js users from using bookmarks, however there is a major trade-off in terms of complexity.  We don't know if there are any users who want to use bookmarks without JS. This allows us to remove 150 lines of code and fixes a bug for the existing majority of users.

cbeer:
I guess I don't understand why we can't just make the checkbox conversion idempotent:

https://github.com/projectblacklight/blacklight/pull/2908

jcoyne:
@cbeer that fixes one problem, but the other problem is that dynamic actions that add bookmarks to the page are not handled.

PR #2908, opened 2022-11-17
Make the checkbox form transformations idempotent

cbeer:


jcoyne:
I get this error when using this branch:
Screenshot 2022-11-17 at 12 55 38 PM

jcoyne:
The script no longer works when turbo restores the page from the cache. 

To reproduce:
1. Visit http://localhost:3000/?f%5Bformat%5D%5B%5D=Book
2. Click the first result
3. Use the browsers back button
4. Use the browsers forward button
5. Click the bookmark checkbox.

jcoyne:
@cbeer this still does not initialize when there is an action that dynamically adds a bookmark form to the page (e.g. turbo-frame).

cbeer:
> (e.g. turbo-frame).

Isn't the solution there to run Blacklight's javascrip on `turbo:frame-load`?  e.g. https://github.com/projectblacklight/blacklight/pull/2915

It seems like we'd want to support any of Blacklight's behaviors 🤷‍♂️

cbeer:
Alternatively, https://github.com/projectblacklight/blacklight/pull/2916 pushes the rendering into ERB without requiring stimulus etc, and maintains a semblance of progressive enhancement

jcoyne (changes requested):
Sure, I'm willing to give it a shot. Can you run the javascript compile step and commit the result to this PR.

jcoyne:
@cbeer can you resolve the merge conflict on this PR?

cbeer:
I /think/ this one can just be closed, because the alternative PR (#2916) was merged. Is there something from this worth keeping?

marlo-longley:
We want to rebase this, check out what the diff is, and see if there is still value; otherwise close.

PR #2913, opened 2022-11-17
Revert "Remove unused method"

cbeer:
This reverts commit 6ba04a395adec98606355fa2a5776ed66a55b3ba.

Going to Blacklight 8, we got rid of the calling code.. but I think it's still a good pattern to keep around if/when we need to change configuration defaults.


dkinzer (approved):

jcoyne:
@cbeer can you add some documentation to this then? I'm totally unclear what it's for.

PR #2932, opened 2022-12-13
Test on ruby 3.2.0

jcoyne:


jcoyne:
Fails due to https://github.com/rails/rails/issues/46716

PR #2937, opened 2022-12-13
Don't include helpers on the views

jcoyne:
This causes blacklight to pollute views that do not need these methods and it makes it confusing if a user is trying to override the helper methods, but they do not also override the views

### DIFF
--- a/app/components/blacklight/top_navbar_component.rb
+++ b/app/components/blacklight/top_navbar_component.rb
@@ -8,6 +8,6 @@ module Blacklight
 
     attr_reader :blacklight_config
 
-    delegate :container_classes, to: :helpers
+    delegate :application_name, :container_classes, to: :helpers
   end
 end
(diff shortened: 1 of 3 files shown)


marlo-longley (approved):

app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -12,6 +12,7 @@ module Blacklight::Catalog
   # The following code is executed when someone includes blacklight::catalog in their
   # own controller.
   included do
+    helper BlacklightHelper
```
cbeer, line 15: Doesn't the blacklight-provided layout require these? Should this go on `Blacklight::Controller`?

jcoyne: Yes, good catch. I can make that change.

thatbudakguy (approved):

PR #2954, opened 2023-01-03
Permit configurable replacement of search tracking with client session storage

### LINKED ISSUE #2558: Consider replacing track_url/session mechanism for previous/next docs with window.sessionStorage
At the cost of constraining browsing a given search to the tab the search was executed in, the previous/next behavior could be rewritten to use window.sessionStorage and a "window" of offsets into the result set. This would simplify the Blacklight back-end and routing, and eliminate POST/redirect cycles from viewing a result document.

barmintor:
fixes #2558 

Implement separate client-side and server-side components for tracking search session, displaying appliedParams block in show view, and displaying result pagination from show view

In this configuration, clicking on a search result stores the current search (query string) in `sessionStorage`, and submit a form that GETs the show view with a `counter` parameter (without POSTing to the track URL). The page links information (ie, previous/next and result set size/offset display) is added dynamically with data fetched from the current search controller.


app/components/blacklight/search_context/client_applied_params_component.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Blacklight
+  module SearchContext
+    class ClientAppliedParamsComponent < Blacklight::Component
+      delegate :search_action_path, to: :helpers
+      def render?
+        controller.params.permit(:counter)[:counter]
```
jcoyne, line 8: Is there a reason you need `.permit(:counter)` here?

app/components/blacklight/search_context/client_applied_params_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<div id="appliedParams" class="clearfix constraints-container">
+  <%= render 'start_over' %>
+  <%= link_to t('blacklight.back_to_search'), search_action_path(only_path: true), class: 'btn btn-outline-secondary back-to-catalog d-none', aria: { live: "off" } %>
```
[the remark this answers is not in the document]
jcoyne: Similarly, the `only_path: true` is the default for search_action_path, so can that be ommitted?

app/javascript/blacklight/search_context.js
```diff
@@ -11,34 +11,42 @@ const SearchContext = (() => {
 
   Blacklight.csrfToken = () => document.querySelector('meta[name=csrf-token]')?.content
   Blacklight.csrfParam = () => document.querySelector('meta[name=csrf-param]')?.content
+  Blacklight.searchStorage = () => document.querySelector('meta[name=blacklight-search-storage]')?.content
 
   // this is the Rails.handleMethod with a couple adjustments, described inline:
   // first, we're attaching this directly to the event handler, so we can check for meta-keys
   Blacklight.handleSearchContextMethod = function(event) {
```
jcoyne, line 18: Is it possible to break this out into two separate functions? This is now 44 lines long and is challenging to make sense of.

app/javascript/blacklight/search_context.js
```diff
@@ -54,6 +62,68 @@ const SearchContext = (() => {
   };
 
   Blacklight.doSearchContextBehavior();
+
+  const clientAppliedParams = function(storedSearch) {
```
jcoyne, line 66: Please add documentation about what this function does.

app/javascript/blacklight/search_context.js
```diff
@@ -54,6 +62,68 @@ const SearchContext = (() => {
...
+        const backToCatalogEle = appliedParams.querySelector('.back-to-catalog')
+        if (backToCatalogEle) {
+          const backToCatalogUrl = new URL(backToCatalogEle.href)
+          backToCatalogEle.href = `${backToCatalogUrl.pathname}?${backToCatalogParams.toString()}`
+          backToCatalogEle.classList.remove('d-none')
+        }
+      }
+    }
+  }
+  const clientPrevNext = function(prevNextPath, storedSearch) {
```
jcoyne, line 83: Please add documentation about what this function does.

app/javascript/blacklight/search_context.js
```diff
@@ -54,6 +62,68 @@ const SearchContext = (() => {
...
+        setContent(pageLinks.querySelector('.pagination-counter-delimited'), responseData.counterDelimited)
+        setContent(pageLinks.querySelector('.pagination-total-raw'), responseData.totalRaw)
+        setContent(pageLinks.querySelector('.pagination-total-delimited'), responseData.totalDelimited)
+        setHrefOrDelete(pageLinks.querySelector("a[rel='prev']"), responseData.prev)
+        setHrefOrDelete(pageLinks.querySelector("a[rel='next']"), responseData.next)
+        pageLinks.classList.remove('d-none')
+      })
+    })
+  }
+  Blacklight.onLoad(function(){
```
jcoyne, line 119: Is it necessary to use onLoad here or can this function be called by `doSearchContextBehavior`?  I think it's confusing to depend on both approaches.

jcoyne (changes requested):

app/components/blacklight/search_context/client_applied_params_component.rb
barmintor: I misunderstood the source of the unpermitted parameter warnings I was getting, but I see they are from search state on show actions. I'll remove the two places this pattern is used in this PR.

app/components/blacklight/search_context/client_applied_params_component.html.erb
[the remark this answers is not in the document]
barmintor: I was under the impression that the `aria-live` attribute indicated to screen readers that the content was dynamic, but the `off` value signaled not to notify. This is far from my strong suit, so if I'm wrong please don't frame your response as a question!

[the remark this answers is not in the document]
barmintor: The `only_path` was copied from the existing behavior - I can remove it in both places.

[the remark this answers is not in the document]
barmintor: @jcoyne hmm - `only_path` is not quite the default. It's the default if there's a hash argument to `search_action_path` that doesn't have the key `:only_path`. If there's no arguments at all, you get the full URL.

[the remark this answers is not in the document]
jcoyne: ugh, that's confusing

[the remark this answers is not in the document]
barmintor: Yeah, I definitely agree. Seems like a separate change though (it breaks a test if I change it in the other location).

app/javascript/blacklight/search_context.js
barmintor: `doSearchContextBehavior` adds an event listener for clicks, which doesn't inspect the DOM (yet). The "page links" behavior in the show view depends on DOM content. I'm trying to refactor.

barmintor: ok, pushed a refactor up. I think that I have to either piggyback on `Blacklight.onLoad` 's event detection, or I would have to reimplement it. I at least relocated the event handler attachment and the onload addition to `doSearchContextBehavior`.

app/javascript/blacklight/search_context.js
```diff
@@ -2,106 +2,157 @@ import Blacklight from './core'
...
+    // intercept clicks on search results to use search context behavior
     document.addEventListener('click', (e) => {
       if (e.target.matches('[data-context-href]')) {
         Blacklight.handleSearchContextMethod.call(e.target, e)
       }
     })
+    // page-links dynamic content depends on a loaded document
+    Blacklight.onLoad(() => {
+      // if page-links container has a data URL, do client-side search context behaviors
+      let clientPageLinks = document.querySelectorAll('.page-links[data-page-links-url]')
```
jcoyne, line 14: This can be a `const`, right?

app/javascript/blacklight/search_context.js
```diff
@@ -2,106 +2,157 @@ import Blacklight from './core'
...
 
-    let formContent = ''
-    for (const [paramName, paramValue] of contextUrl.searchParams.entries()) {
-      formContent += `<input name="${paramName}" value="${paramValue}" type="hidden" />`
-    }
+    // check for meta keys.. if set, we should open in a new tab
+    const target = (event.metaKey || event.ctrlKey) ? '_blank' : formTarget
+    if (target) form.target = target
+
+    form.style.display = 'none'
```
jcoyne, line 59: Is it possible to do `form.hidden = true` here instead? (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden)

app/javascript/blacklight/search_context.js
```diff
@@ -2,106 +2,157 @@ import Blacklight from './core'
...
+  /**
+   * fetch the JSON data at pageLinksPath for the given search
+   * use the fetched data to build the page-links labels and prev/next links
+   * @param {string} pageLinksPath
+   * @param {?URLSearchParams} storedSearch
+   */
+  Blacklight.clientPageLinks = function(pageLinksPath, storedSearch) {
+    const pageLinksUrl = new URL(pageLinksPath, window.location)
+    // pageLinksUrl should already have a counter param, but needs search params
+    let prevNextParams = new URLSearchParams(pageLinksUrl.search);
```
jcoyne, line 149: Can this be a `const`?

jcoyne (changes requested):

app/components/blacklight/search_context/client_applied_params_component.html.erb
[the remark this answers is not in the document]
jcoyne: Rather than `d-none`, can we use the hidden attribute on the html?  That would prevent coupling the behavior with the css toolkit.

app/javascript/blacklight/search_context.js
barmintor: yep, done

app/javascript/blacklight/search_context.js
barmintor: yep, done

jcoyne (approved):

hackartisan (approved):
It's important to document the implications of this PR. Where?

cbeer:
I think it'd be good to split the search history tracking from search context pagination. Spotlight, for example, has a pretty significant dependency on search history, but I think it'd be great to get this better client-side context in play too.

barmintor:
I updated the PR description to be more explicit about the changes here.

barmintor:
@jcoyne assuming we are pursuing the breaking-API PR for 8.0, I'll need to fix this PR up afterwards. I'm converting to draft while I work on #2970

PR #2959, opened 2023-01-24
On modals, flash messages display once

maxkadel:
Co-authored-by: Jane Sandberg <sandbergja@users.noreply.github.com>


maxkadel:
Flash messages are implemented differently on main, so this only needs to be done against the release-7.x branch.

PR #2961, opened 2023-01-24
On modals, flash messages display once

### LINKED ISSUE #2960: 7.x - Success flash messages from modals display twice
For actions like emailing bookmarks or sending a record via SMS that open in a modal, when the action is successful the modal is closed and the flash message is displayed on the main page. If you dismiss this message and then reload the page (or go to a different page, or open a modal), the flash message is displayed again. 

As a user, I would expect not to see the message again after dismissing it the first time. 

Since the Blacklight 8 uses a different implementation for flash messages (turbo-streams), this fix is not needed on the main branch

### On initial submit
Bookmark page displays success flash message

### After dismissing first modal
Bookmark page with no flash message

### After dismissing and doing a new search
Search results page with success flash message

maxkadel:
This PR only needs to be merged into the release-7.x branch, since the main branch uses a different implementation for flash messages.

### On initial submit
Bookmark page displays success flash message

### After dismissing first modal
Bookmark page with no flash message

### After dismissing and doing a new search
Search results page with no flash message

Closes #2960

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -535,7 +535,10 @@ RSpec.describe CatalogController, api: true do
         post :email, xhr: true, params: { id: doc_id, to: 'test_email@projectblacklight.org' }
         expect(request).to render_template 'email_success'
         expect(request.flash[:success]).to eq "Email Sent"
-        expect(flash.instance_variable_get("@discard").first).to eq("success")
+        allow(search_service).to receive(:search_results)
+        # When we go to another page, the flash message should no longer display
+        get :index
+        expect(request.flash[:success]).to be_nil
       end
     end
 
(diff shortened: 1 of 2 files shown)


bess (approved):

spec/controllers/catalog_controller_spec.rb
```diff
@@ -535,6 +535,7 @@ def export_as_mock
         post :email, xhr: true, params: { id: doc_id, to: 'test_email@projectblacklight.org' }
         expect(request).to render_template 'email_success'
         expect(request.flash[:success]).to eq "Email Sent"
+        expect(flash.instance_variable_get("@discard").first).to eq("success")
```
jcoyne, line 538: This seems like a fragile test because it is tightly coupled to the implementation.  A better way might do another `get` request and show that the `request.flash[:success]` is empty.

maxkadel: Thanks @jcoyne, good point, I've changed the tests.

jcoyne (approved):

sandbergja (approved):

PR #2962, opened 2023-01-25
Issue 2957

barmintor:
Use ViewComponent::SlotableV2 API for compatibility with ViewComponent 3
Configure a CI matrix run to use ViewComponent 3 RC


jcoyne (approved):

.env
```diff
@@ -1,5 +1,6 @@
 ALPINE_RUBY_VERSION=3.0.3
 RAILS_VERSION=6.1.4.6
+VIEW_COMPONENT_VERSION=2.6.6
```
jcoyne, line 3: Is this the right version?

barmintor: It's the minimum version we advertise support for - which seemed analogous to the `RAILS_VERSION` value (unless I misread). I'm not sure what the guiding principle for these values should be.

jcoyne: I think this is for the docker build and in that case I don't think you  need this line at all.

PR #2964, opened 2023-01-30
Restore whole document modal ajax behavior

### LINKED ISSUE #2963: Modal receiveAjax regression
The documentation describes a behavior where the receiveAjax handles either a page with a node that has an attribute like `<div data-blacklight-modal="container">` or the server will only return the data to be displayed.  

https://github.com/projectblacklight/blacklight/blob/main/app/javascript/blacklight/modal.js#L20-L24

I believe that Blacklight 7 handled both of these cases, but blacklight 8 doesn't.   There appears to be a regression https://github.com/projectblacklight/blacklight/commit/7a84b0d3a577bc619c2b2cc1df45e9702e7ee7b0

jcoyne:
Fixes #2963


app/javascript/blacklight/modal.js
```diff
@@ -105,10 +105,15 @@ const Modal = (() => {
   }
 
   // Add the passed in contents to the modal and display it.
-modal.receiveAjax = function (contents) {
+  modal.receiveAjax = function (contents) {
       const domparser = new DOMParser();
       const dom = domparser.parseFromString(contents, "text/html")
-      const elements = dom.querySelectorAll(`${modal.containerSelector} > *`)
+      // If there is a containerSelector on the document, use it's children.
```
sandbergja, line 111:
```suggestion
      // If there is a containerSelector on the document, use its children.
```

sandbergja (changes requested):
Looks good to me, thanks @jcoyne !  Just one nitpicky grammar change.

barmintor (approved):
This behavior was inadvertently removed in the rewrite without jquery: https://github.com/projectblacklight/blacklight/commit/7a84b0d3a577bc619c2b2cc1df45e9702e7ee7b0

That change was not backported, so this is a regression fix. The previous behavior in https://github.com/projectblacklight/blacklight/commit/4976e2d34539a5eb90a7de40689357fc9711e66e used a jquery method `container.html();` `dom.querySelectorAll('*')` should restore the behavior without the jquery dependency.

sandbergja (approved):
Thank you, @jcoyne !

PR #2968, opened 2023-01-30
Select the childNodes of the ajax response

jcoyne:
Not all the decendant nodes. Otherwise the returned content is flattened as it is returned.


carolyncole (approved):
I assume this is a follow up to #2963.  I may not be the correct person to review this as I have no context, but it seems ok to me.

PR #2969, opened 2023-01-31
Update docs to show modal button behavior

jcoyne:
You need to use data-bl-dismiss not data-bs-dismiss


sandbergja (approved):
Thanks, @jcoyne !

PR #2972, opened 2023-02-01
Remove unused code

cbeer:


jcoyne:
Can we add https://github.com/projectblacklight/blacklight/blob/main/lib/blacklight/configuration/view_config.rb#L29 to this too?

jcoyne (approved):

PR #3012, opened 2023-02-13
New style for collapsed facets on narrow browsers

jcoyne:
When the browser is narrow, it may be confusing that there are two hamburger menus.  In order to differentiate between the two, this changes the label for the button that shows and hides facets. There are no visible changes at wider viewport widths.

Before:
Screenshot 2023-02-13 at 11 21 47 AM

After:

Screenshot 2023-02-13 at 1 50 38 PM
Screenshot 2023-02-13 at 1 50 49 PM


thatbudakguy (approved):

seanaery:
Thanks @jcoyne -- I like having this change be part of core Blacklight and the way you've done it makes the labeling easily customizable. In the past couple years we had started doing this as a local customization on some of our apps because the double 🍔  was unclear.

bess (approved):

bess:
@jcoyne Looks like a great UX improvement to me. Thank you!

PR #3028, opened 2023-03-01
There is no need to have an initializer load the defaults

jcoyne:
The defaults are loaded when the configuration is initialized.  The existing initializer was overwriting values that Spotlight was trying to set in it's initializer

### DIFF
--- a/lib/blacklight/engine.rb
+++ b/lib/blacklight/engine.rb
@@ -6,10 +6,6 @@ module Blacklight
   class Engine < Rails::Engine
     engine_name "blacklight"
 
-    config.after_initialize do
-      Blacklight::Configuration.initialize_default_configuration
-    end
-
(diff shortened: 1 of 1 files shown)


cbeer:
If you remove this, won't there be times that `Blacklight::Configuration.default_values` isn't populated at all?

I'd think #2950 would address this just fine:
https://github.com/projectblacklight/spotlight/pull/2930

jcoyne:
@cbeer thanks for your help. This is an area of the code that I don't have a lot of experience in.

It looks to me that this code ensures that the default configuration is populated as soon as the `Blacklight:Configuration` class is defined: https://github.com/projectblacklight/blacklight/blob/9e517006cf8373af1d623c2664276038a2258b2e/lib/blacklight/configuration.rb#L60

PR #3039, opened 2023-03-19
deprecate _show_tools partial

### LINKED ISSUE #3038: The _show_tools.html.erb partial should be deprecated in favor of a configurable show tools component
This partial only renders a component; the component should be configurable and the partial slated for removal in the next major version of Blacklight.

barmintor:
Deprecate the _show_tools partial in favor of a configurable component
fixes #3038

This PR also restores the helper method for inspecting whether a partial was overridden to minimize log noise when a Blacklight app is using default behaviors

### DIFF
--- a/lib/blacklight/configuration.rb
+++ b/lib/blacklight/configuration.rb
@@ -181,6 +181,9 @@ module Blacklight
         # document presenter class used by helpers and views
         document_presenter_class: nil,
         document_component: Blacklight::DocumentComponent,
+        # in Blacklight 9, the default show_tools_component configuration will
+        # be Blacklight::Document::ShowToolsComponent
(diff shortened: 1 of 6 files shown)


lib/blacklight/configuration.rb
```diff
@@ -181,6 +181,9 @@ def initialized_default_configuration?
         # document presenter class used by helpers and views
         document_presenter_class: nil,
         document_component: Blacklight::DocumentComponent,
+        # in Blacklight 9, the default show_tools_component configuration will
+        # be Blacklight::Document::ShowToolsComponent
+        show_tools_component: nil,
```
jcoyne, line 186: Can we set this as the default right now?

PR #3073, opened 2023-08-17
Backport #3066

sandbergja:
Adds #3066 to the 7.x release series

### DIFF
--- a/lib/blacklight/configuration.rb
+++ b/lib/blacklight/configuration.rb
@@ -78,6 +78,10 @@ module Blacklight
     # @since v5.2.0
     # @return [String] The url path (relative to the solr base url) to use when requesting only a single document
     property :document_solr_path, default: 'get'
+    # @!attribute json_solr_path
+    # @since v7.34.0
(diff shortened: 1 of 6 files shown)


lib/blacklight/configuration.rb
```diff
@@ -78,6 +78,10 @@ def header_component
     # @since v5.2.0
     # @return [String] The url path (relative to the solr base url) to use when requesting only a single document
     property :document_solr_path, default: 'get'
+    # @!attribute json_solr_path
+    # @since v7.34.0
+    # @return [String] The url path (relative to the solr base url) to use when using Solr's JSON Query DSL (as with the advanced search)
+    property :json_solr_path, default: 'advanced'
```
cbeer, line 84:
```suggestion
    property :json_solr_path, default: nil
```

Should we set this to `nil` by default for backwards compatibility (esp. for people who worked around that upstream solr bug some other way)?

tpendragon (approved):

PR #3075, opened 2023-09-08
Update to use options for feature specs

maxkadel:
This should get CI for the release-7.x branch passing.


sandbergja (approved):
Thanks, @maxkadel !

PR #3103, opened 2023-11-08
[3100-highlight] When no highlighted field is available fall back into the field value

### LINKED ISSUE #3100: [highlight] Activating highlight: true for specific fields will hide the field if the search doesn't find a match in the specific field
Steps to replicate:
In solrconf.xml:
```
<str name="hl">true</str>
<str name="hl.method">unified</str>
<str name="hl.fl">author_tsim title_tsim</str>
 <str name="hl.snippets">2</str>
```
In catalog_controller.rb
```
config.add_index_field 'title_tsim', label: 'Title', highlight: true
config.add_index_field 'author_tsim', label: 'Author', highlight: true
```
`docker compose down `
`docker compose up -d solr `
`rake blacklight:index:seed`

visit: [Blacklight url](http://localhost:3000/)
Search for `nectar` will return a result with the emphasized title. The title displays. The author does not display.

christinach:
closes #3100

### DIFF
--- a/app/services/blacklight/field_retriever.rb
+++ b/app/services/blacklight/field_retriever.rb
@@ -22,17 +22,19 @@ module Blacklight
 
     # @return [Array]
     def fetch
-      Array.wrap(
-        if field_config.highlight
-          retrieve_highlight
-        elsif field_config.accessor
-          retieve_using_accessor
-        elsif field_config.values
-          retrieve_values
-        else
-          retrieve_simple
-        end
-      )
+      if field_config.highlight
+        value = retrieve_highlight
+      end
+      if value.blank?
+        value = if field_config.accessor
+                  retieve_using_accessor
+                elsif field_config.values
+                  retrieve_values
+                else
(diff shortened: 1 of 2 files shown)


app/services/blacklight/field_retriever.rb
```diff
@@ -22,17 +22,19 @@ def initialize(document, field_config, view_context = nil)
...
-          retrieve_highlight
-        elsif field_config.accessor
-          retieve_using_accessor
-        elsif field_config.values
-          retrieve_values
-        else
-          retrieve_simple
-        end
-      )
+      if field_config.highlight
```
tpendragon, line 25: It's funny, I almost wish that the `retrieve_` methods just reutnred nil if `accessor` or `values` or whatever wasn't set, so we could do `retrieve_highlight || retrieve_using_accessor || retrieve_values || retrieve_simple`, but that's probably too big for here.

tpendragon (approved):

PR #3145, opened 2024-02-09
Implement context tracking with the beacon API

### LINKED ISSUE #3144: Replace custom tracking with sendBeacon
This has good support in all modern browsers https://caniuse.com/beacon

jcoyne:
This means less code to deal with ourselves.
This API has good support in all modern browsers https://caniuse.com/beacon

Fixes #3144


jcoyne:
While the beacon works, the next page load occurs at the same time, so the tracking URL is not in the same session cookie as the item page draw request.

PR #3153, opened 2024-05-10
Generate <button> for flash alert dismissal

marlo-longley:
Currently we are generating `<button-tag>` which is not valid.
This is causing styling difficulties downstream.

### DIFF
--- a/app/components/blacklight/system/flash_message_component.html.erb
+++ b/app/components/blacklight/system/flash_message_component.html.erb
@@ -1,6 +1,6 @@
 <div class="alert <%= @classes %>">
   <%= message %>
-  <%= tag.button_tag button_contents, type: "button", class: "btn-close",
+  <%= tag.button button_contents, type: "button", class: "btn-close",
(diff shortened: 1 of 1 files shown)


app/components/blacklight/system/flash_message_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="alert <%= @classes %>">
   <%= message %>
-  <%= tag.button_tag button_contents, type: "button", class: "btn-close",
+  <%= button_tag button_contents, type: "button", class: "btn-close",
```
cbeer, line 3:
```suggestion
  <%= tag.button button_contents, type: "button", class: "btn-close",
```

Was maybe the intent?

marlo-longley: Thanks @cbeer -- not sure the difference, other than `button_tag` seems to generate an additional `name` attribute. Happy to switch to this though.

Screenshot 2024-05-10 at 17 19 49

jcoyne (approved):

PR #3159, opened 2024-05-23
Fix the facet group toggle button accessible name

taylor-steve:
At smaller sizes when the facet group toggle button is displayed, we're getting a [Label in Name (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name) violation because the button has been given a fixed aria-label, but the visible text varies based on state (e.g., "Toggle Facets" versus "Show Facets"/"Hide Facets").

If we stop explicitly setting the toggle aria-label I believe it will use the visible text for accessibility purposes (it makes Siteimprove happy).

### DIFF
--- a/app/components/blacklight/response/facet_group_component.html.erb
+++ b/app/components/blacklight/response/facet_group_component.html.erb
@@ -15,7 +15,6 @@
       aria: {
         controls: @panel_id,
         expanded: 'false',
-        label: t('blacklight.search.facets.group.toggle'),
       } do %>
       <span data-show-label><%= t('blacklight.search.facets.group.open') %></span>
       <span data-hide-label><%= t('blacklight.search.facets.group.close') %></span>
--- a/config/locales/blacklight.en.yml
+++ b/config/locales/blacklight.en.yml
@@ -191,7 +191,6 @@ en:
           show: Show
           hide: Hide
         group:
-          toggle: Toggle facets
           open: Show facets
           close: Hide facets
       group:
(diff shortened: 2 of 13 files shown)


app/components/blacklight/response/facet_group_component.html.erb
```diff
@@ -15,7 +15,6 @@
       aria: {
         controls: @panel_id,
         expanded: 'false',
-        label: t('blacklight.search.facets.group.toggle'),
```
jcoyne, line 18: Should this key be removed from our i18n files?

jcoyne (changes requested):

app/components/blacklight/response/facet_group_component.html.erb
taylor-steve: It is the only place this key is used but I intentionally left it in case downstream applications utilize it in an override. Would you typically remove it here? Happy to.

jcoyne: I think we need to remove it or we'll just accumulate junk forever.  At the very least ticket it for removal at some other time so it isn't forgotten.

taylor-steve: I have removed it. I accumulate enough junk in other areas of my life!

config/locales/blacklight.en.yml
```diff
@@ -191,7 +191,6 @@ en:
           show: Show
           hide: Hide
         group:
-          toggle: Toggle facets
```
jcoyne, line 194: We have to do all the languages, not just `en`

marlo-longley (approved):

jcoyne (approved):

PR #3181, opened 2024-06-07
Remove extraneous solr 9 config

jcoyne:
All other tests are running solr 9, so we don't need a separate test


cjcolvar (approved):
Looks good to me.

corylown (approved):

PR #3207, opened 2024-06-20
Remove unnecessary clearfix

jcoyne:
There are no floating elements here to clear

Ref https://github.com/projectblacklight/blacklight/issues/3206


tampakis (approved):
Looks good!

PR #3217, opened 2024-07-15
add optional bookmark icon for bookmarks

dnoneill:
When the user updates the config to point to a icon component it will render the icon as a bookmark instead of the checkbox.

Screenshot 2024-07-17 at 1 28 24 PM
Screenshot 2024-07-17 at 1 28 20 PM
Screenshot 2024-07-17 at 1 28 12 PM

### DIFF
--- a/app/assets/stylesheets/blacklight/_bookmark.scss
+++ b/app/assets/stylesheets/blacklight/_bookmark.scss
@@ -18,3 +18,23 @@ label.toggle-bookmark {
 		display: none;
 	}
 }
+.toggle-bookmark .blacklight-icons svg {
+  height: 1.25rem;
+  width: 1.25rem;
+  overflow: visible;
+  fill: var(--bs-primary);
+
+  &.bookmark-checked {
+    display: none;
+  }
+}
+
+.toggle-bookmark[type="checkbox"]:checked+span svg {
+  &.bookmark-checked {
+    display: inherit;
+  }
+
+  &.bookmark-unchecked {
+    display: none;
+  }
+}
--- a/app/components/blacklight/document/bookmark_component.rb
+++ b/app/components/blacklight/document/bookmark_component.rb
@@ -21,6 +21,12 @@ module Blacklight
         helpers.bookmarked? @document
       end
 
+      def bookmark_icon
+        return unless helpers.blacklight_config.bookmark_icon_component
+
+        render helpers.blacklight_config.bookmark_icon_component.new(name: 'bookmark')
+      end
+
       def bookmark_path
         @bookmark_path || helpers.bookmark_path(@document)
       end
--- a/lib/blacklight/configuration.rb
+++ b/lib/blacklight/configuration.rb
@@ -150,6 +150,12 @@ module Blacklight
       # @return [#partials]
       property :navbar, default: OpenStructWithHashAccess.new(partials: {})
 
+      # @!attribute bookmark_icon_component
+      # @since v8.3.1
+      # component class used to render a document
+      # set to Blacklight::Icons::BookmarkIconComponent to replace checkbox with icon
+      property :bookmark_icon_component, default: nil
+
       # @!attribute index
       # General configuration for all views
       # @return [Blacklight::Configuration::ViewConfig::Index]
--- a/app/components/blacklight/document/bookmark_component.html.erb
+++ b/app/components/blacklight/document/bookmark_component.html.erb
@@ -10,7 +10,8 @@
             }) do %>
   <div class="checkbox toggle-bookmark">
     <label class="toggle-bookmark" data-checkboxsubmit-target="label">
-      <input type="checkbox" class="toggle-bookmark" data-checkboxsubmit-target="checkbox" <%= 'checked="checked"' if bookmarked? %>>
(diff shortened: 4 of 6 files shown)


app/assets/stylesheets/blacklight/_bookmark.scss
```diff
@@ -18,3 +18,16 @@ label.toggle-bookmark {
...
+.toggle-bookmark .blacklight-icons svg {
+  height: 1.25rem;
+  width: 1.25rem;
+  path {
+    fill:none;
+    stroke: var(--bs-primary);
+  }
+}
+
+.toggle-bookmark.checked .blacklight-icons path {
```
jcoyne, line 31: Can you explain why we're using a css class (`.checked`) rather than the `:checked` pseudo-class?

app/assets/stylesheets/blacklight/_bookmark.scss
```diff
@@ -18,3 +18,16 @@ label.toggle-bookmark {
 		display: none;
 	}
 }
+
+.toggle-bookmark .blacklight-icons svg {
+  height: 1.25rem;
+  width: 1.25rem;
+  path {
+    fill:none;
```
jcoyne, line 26:
```suggestion
    fill: none;
```
Please add a space here.

app/assets/stylesheets/blacklight/_bookmark.scss
dnoneill: Either will work I figured since we actually attach the class there is made sense.

app/components/blacklight/document/bookmark_component.rb
```diff
@@ -21,6 +21,13 @@ def bookmarked?
         helpers.bookmarked? @document
       end
 
+      def bookmark_icon
+        icon = helpers.blacklight_icon("bookmark")
+        return unless icon.include?('svg')
```
jcoyne, line 26: Can you add a comment in the code explaining why we need to check the return value of the `blacklight_icon` helper?  Should we add a test for both paths here?

dnoneill: I am not sure I understand the both paths question. What do you mean by that?

dnoneill: In the rspec?

dnoneill: Done

app/components/blacklight/document/bookmark_component.rb
```diff
@@ -21,6 +21,14 @@ def bookmarked?
         helpers.bookmarked? @document
       end
 
+      def bookmark_icon
+        icon = helpers.blacklight_icon("bookmark")
+        # Check if there is a bookmark icon since it is optional
```
jcoyne, line 26: Can you expand on this?  How is it optional?  Where is the option?

dnoneill: You have to have a bookmark.svg file in the assets/images/blacklight folder, which isn't the default. Maybe enabled or existing is better language?

jcoyne: I think saying that in the comments is helpful here. Otherwise there's no way for a user to find out about this feature.

spec/fixtures/bookmark.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 28">
```
jcoyne, line 1: Where is this coming from?  Do we have license to include it with Blacklight?

jcoyne (changes requested):

spec/fixtures/bookmark.svg
dnoneill: Shoot meant to grab the blacklight icons one. This is the one from geoblacklight.

app/components/blacklight/document/bookmark_component.rb
```diff
@@ -21,6 +21,14 @@ def bookmarked?
         helpers.bookmarked? @document
       end
 
+      def bookmark_icon
+        icon = helpers.blacklight_icon("bookmark")
```
jcoyne, line 25: This is a deprecated method and calling it is going to cause a deprecation warning. https://github.com/projectblacklight/blacklight/blob/e957632975c4e190ed94a812bee74d774a5459af/app/helpers/blacklight/icon_helper_behavior.rb#L12-L20

dnoneill: Without this I have to create a component which I haven't figured out a good way to switch between in tests. Do you have any ideas?

jcoyne: Can we make it a configuration setting like `config.bookmark.use_icon` and set it to false by default?  That's a pretty good way of having optional features as we use it for other things as well.

lib/blacklight/configuration.rb
```diff
@@ -150,6 +150,12 @@ def initialized_default_configuration?
       # @return [#partials]
       property :navbar, default: OpenStructWithHashAccess.new(partials: {})
 
+      # @!attribute bookmark_icon_component
+      # @since v5.8.0
```
jcoyne, line 154: This version isn't correct.

dnoneill: Oops. Thanks updated to v8.3.1

jcoyne (approved):

PR #3219, opened 2024-07-17
Add Configuration#fetch_many_documents_path 

cbeer:
... and use it when we're retrieving multiple documents by id.

By default, this uses the existing search behavior (maintaining backwards compatible), but allows applications to use other paths (e.g. the real-time `/get` or other path) that's more appropriate. Without this change, applications have been limited to using `qt` with old-style request handlers in Solr.

### DIFF
--- a/app/services/blacklight/search_service.rb
+++ b/app/services/blacklight/search_service.rb
@@ -148,7 +148,15 @@ module Blacklight
               .merge(blacklight_config.fetch_many_document_params)
               .merge(extra_controller_params)
 
-      solr_response = repository.search(query)
+      # find_many was introduced in Blacklight 8.4. Before that, we used the
+      # regular search method (possibly with a find-many specific `qt` parameter).
+      # In order to support Repository implementations that may not have a find_many,
+      # we'll fall back to search if find_many isn't available.
+      solr_response = if repository.respond_to?(:find_many)
+                        repository.find_many(query)
+                      else
+                        repository.search(query)
+                      end
 
       solr_response.documents
     end
--- a/lib/blacklight/abstract_repository.rb
+++ b/lib/blacklight/abstract_repository.rb
@@ -35,6 +35,12 @@ module Blacklight
       raise NotImplementedError
     end
 
+    # Find multiple documents by their ids
+    # @param [Hash] _params query parameters
+    def find_many(params, **kwargs)
+      search(params, **kwargs)
+    end
+
     ##
     # Execute a search query against a search index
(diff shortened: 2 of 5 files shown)


jcoyne (approved):

app/services/blacklight/search_service.rb
```diff
@@ -148,7 +148,11 @@ def fetch_many(ids, extra_controller_params)
               .merge(blacklight_config.fetch_many_document_params)
               .merge(extra_controller_params)
 
-      solr_response = repository.search(query)
+      solr_response = if repository.respond_to?(:find_all)
```
jcoyne, line 151: Won't all repositories that extend from `AbstractRepository` now respond to find_all?  Should we add a deprecation warning if they don't implement `find_all`?

lib/blacklight/abstract_repository.rb
```diff
@@ -35,6 +35,12 @@ def find(_id, _params = {})
       raise NotImplementedError
     end
 
+    # Find multiple documents by their ids
+    # @param [Hash] _params query parameters
+    def find_all(params)
```
jcoyne, line 40: `find_all` might not be the best name as it's only a subset we're actually finding, right?

app/services/blacklight/search_service.rb
cbeer: That was my hope, but I don't think we can guarantee downstream applications have done that.

I'm not sure we need a deprecation warning until we want to remove the fallback to `search` but I don't see a pressing need for that 🤷‍♂️

lib/blacklight/abstract_repository.rb
cbeer: Got a better suggestion? I was inspired by ActiveRecord and Enumerable that use(d) `find_all` to mean "find all the match some condition"

jcoyne: is `find_many` confusing? I know we are using it at a different level of abstraction, but it seems to accurately describe it's function.

app/services/blacklight/search_service.rb
jcoyne: If not a deprecation, would you mind dropping a brief comment in the code here explaining why we would want to check first?

lib/blacklight/abstract_repository.rb
cbeer: 🤷‍♂️ sure

jcoyne (approved):

PR #3230, opened 2024-07-18
Drop support for view_component 2

### LINKED ISSUE #3229: Drop view_component 2.x support

jcoyne:
Fixes #3229


thatbudakguy:
What's the context for this? Just to keep the version fresh, or does 2.x cause problems?

jcoyne:
Yeah, we just don't want to support two major versions of the dependency.

thatbudakguy (approved):

PR #3235, opened 2024-07-26
Close FacetFieldCheckboxesComponent span tag

### LINKED ISSUE #3234: Unclosed span on FacetFieldCheckboxesComponent
I noticed that the `span` in this component that contains the label and count is unintentionally unclosed, resulting in two opened spans instead of one closed span:
```ruby
          <span>
            <%= label_tag "f_inclusive_#{@facet_field.key}_#{idx}" do %>
              <span class="facet-label"><%= presenter.label %></span>
              <span class="facet-count"><%= t('blacklight.search.facets.count', number: number_with_delimiter(presenter.hits)) %></span>
            <% end %>
          <span>
```

Unclear whether this leads to any unexpected behavior/accessibility concerns, but it can be remedied quite easily.

body-clock:
Closes #3234


jcoyne:
Thank you. @body-clock

PR #3240, opened 2024-07-30
Add a variable to target facet spacing

jcoyne:
This allows adopters to easily customize this spacing without affecting the global spacing


sandbergja (approved):
Thanks, @jcoyne

PR #3246, opened 2024-08-01
Don't fix permissions since ruby 3.2.5 has been released

jcoyne:


jcoyne:
It appears 3.2.5 has not resolved the issue.

PR #3254, opened 2024-08-05
Fix docker compose usage

jcoyne:
docker-compose has been replaced by docker compose (no dash).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
         - ALPINE_RUBY_VERSION
     # mounting . is causing seg-fault on macosx
     #volumes:
-      #- .:/app
+    #- .:/app
     depends_on:
       - solr
     ports:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
         - ALPINE_RUBY_VERSION
     # mounting . is causing seg-fault on macosx
     #volumes:
-      #- .:/app
+    #- .:/app
     depends_on:
```
marlo-longley, line 11: If we're moving to the newer conventions, they changed the recommended filename from "docker-compose" to just "compose" 

https://docs.docker.com/compose/gettingstarted/#step-2-define-services-in-a-compose-file

jcoyne: This PR is not about conventions, but that they have completely removed `docker-compose` from this image.

jcoyne: I created https://github.com/projectblacklight/blacklight/issues/3256 to track this suggestion.

corylown (approved):

PR #3226, opened 2024-07-18
Remove bootstrap 4 support

jcoyne:
Fixes #3211

PR #3262, opened 2024-08-07
Use Bootstrap 5 Accordion for the facet list

jcoyne:
Screenshot 2024-08-07 at 10 04 19 AM

This was not available in Bootstrap 4.  It allows a lot more customization (via Bootstrap variables) and requires Blacklight itself to have less code.

### DIFF
--- a/app/assets/stylesheets/blacklight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/blacklight/_bootstrap_overrides.scss
@@ -1,27 +1,3 @@
-// Facet field headings and buttons
-.facet-field-heading {
-  border-bottom: 0;
-
-  button {
-    font-weight: $headings-font-weight;
-
-    &::after {
-      content: "❯";
-      float: right;
-      transform: rotate(90deg);
-    }
-
-    &.collapsed {
-      border-bottom: 0;
-
-      &::after {
-        transform: rotate(0deg);
-        transition: transform 0.1s ease;
-      }
-    }
-  }
-}
-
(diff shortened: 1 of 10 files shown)


jrochkind:
After this PR, does facet list still work with Bootstrap 4, or is this a clean break requiring Bootstrap 5?

If it's a clean break, and that's what the community wants, then please make sure that's clearly noted in release notes for version that includes this?  Also to avoid confusion probably remove  the config `bootstrap_version`, and possibly remove other support for bootstrap 4?

jcoyne:
@jrochkind this is only for BS 5 and it can be merged after https://github.com/projectblacklight/blacklight/pull/3226 is in place.

jcoyne:
This is how IU is styling this component:
Screenshot 2024-10-16 at 10 27 18 AM

jrochkind:
We have long done similar styling in Blacklight 7/Boostrap 4, adding styles to make the facet menu look like this even in that situation. 

Screenshot 2024-10-16 at 11 34 45 AM


https://digital.sciencehistory.org/catalog

cbeer:
If we're going to remove the whitespace and borders around each facet, I think we also need to change the background color of the accordion headers (like IU + SHI do) to visually distinguish between the header and the facet values.

jcoyne:
Good suggestion @cbeer I've adjusted it as per this screenshot:
Screenshot 2024-10-22 at 12 19 05 PM

jrochkind (approved):

PR #3215, opened 2024-06-25
Resolve "Adding bookmark without bookmark counter raises an error"

mamrey:
Closes #3214

PR #3224, opened 2024-07-18
Stop supporting Rails 6

jcoyne:
Fixes #3223

PR #3263, opened 2024-08-08
Bookmark counter spec

mamrey:
A flaky test from #3215 was removed in  #3245. 

This PR fixes and replaces said spec, testing the ability to add/remove bookmarks without the bookmark navbar element present. 

These changes also add an `after` hook to clean up modifications to `CatalogController` in the bookmark icon example

Ran the bookmark test suite 100 times, all green.

### DIFF
--- a/spec/features/bookmarks_spec.rb
+++ b/spec/features/bookmarks_spec.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 RSpec.describe "Bookmarks" do
+  before do
+    # Prevent any existing bookmarks from interfering
+    Bookmark.delete_all
+  end
+
   describe "navigating from the homepage" do
     it "has a link to the history page" do
       sign_in 'user1'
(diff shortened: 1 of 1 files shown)


spec/features/bookmarks_spec.rb
```diff
@@ -96,4 +107,26 @@
...
+
+    after do
+      CatalogController.blacklight_config.navbar.partials = default_nav_partials
+    end
+
+    it 'adds and removes bookmarks', :js do
+      visit solr_document_path('2007020969')
+      expect(page).to have_no_css('#bookmarks_nav')
+      check 'Bookmark'
+
```
jcoyne, line 126: Can you add a check here before navigating away from the page that shows it has been saved? Typically you can look for `In Bookmarks` in the page content.

jcoyne (approved):

spec/features/bookmarks_spec.rb
mamrey: Thanks, good call.

 I have added that check, but it fails the rails 6 check. I built an internal test app using rails 6.1.7.8 and encountered the same failure. When inspecting the app, it looks like the  `BookmarkToggle` event handler callback is fired twice in succession, so the checkbox checks and unchecks itself with one click. I haven't pinpointed the issue, but perhaps it's related to how the js assets are built/handled when using rails 6?

mamrey: Some additional context: 

In the rails `6.1.7.8` test app, the browser receives both a `blacklight.source.js` and `blacklight.debug.js` file that appear the same. The `clicked` function is called in both these js files according to the debugger. This has the effect of the checkbox being clicked twice in succession, which reflects the failing test.  While I am unfamiliar with the asset pipeline for rails 6 and BL, I will investigate [turning off debug mode](https://guides.rubyonrails.org/v4.0.0/asset_pipeline.html#turning-debugging-off).

sandbergja:
@mamrey now that main no longer supports rails 6.1 (as of #3224), would you mind rebasing this PR?  Thank you!

mamrey:
Thank you for further investigating the issue and moving things forward @sandbergja!

PR #3274, opened 2024-08-14
Use nav tag rather than section tag for pagination links

jcoyne:
This offers better accessibility and follows the example markup in https://getbootstrap.com/docs/5.3/components/pagination/\#disabled-and-active-states


jcoyne:
This depends on #3275  being merged first.

jrochkind (approved):

PR #3275, opened 2024-08-14
Provide aria-label for main navigation

jcoyne:


app/components/blacklight/top_navbar_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-expand-md navbar-dark bg-dark topbar" role="navigation">
+<nav class="navbar navbar-expand-md navbar-dark bg-dark topbar" aria-label="<%= aria_label %>">
```
jcoyne, line 1: We don't need the `role="navigtion"` because the role is implied https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav#technical_summary

jrochkind (approved):

tpendragon (approved):

PR #3277, opened 2024-08-14
allow show tools to be moved to the header, closes #2316

### LINKED ISSUE #2316: Move show page tools to page header area
Part of #2314.

The way the show page tools are currently displayed -- in a sidebar box -- is arguably less than ideal in a couple of ways. First, just visually, the display in a dedicated sidebar box makes the tools, which in most cases are likely to be used only occasionally by users, very prominent. More importantly, if a Blacklight-based application wants to use the show page sidebar to display additional application-specific features it's challenging to display them in a way that makes them more prominent to the user than the Blacklight tools.

A specific example of this is in GeoBlacklight, where the show page sidebar can be used to show the user several types of resources and tools related to the current item. With the way the Blacklight tools are currently displayed, this results in a cluttered sidebar where the geo-specific tools are mixed in with the generic tools and they all have equal weight in the sidebar:

Screen Shot 2020-07-24 at 4 11 19 PM

This ticket proposes moving the Blacklight tools out of a sidebar and to a set of buttons that goes in the page header area. This places them in a location where they feel de-emphasized but available, and provides more focused sidebar for applications like GeoBlacklight to place application-specific tools or features.

I'd suggest:

- Right-align the group of buttons on the same row as the pagination controls
- Style the buttons as `btn btn-outline-secondary btn-sm` 

Examples below.

## Item show page with no sidebar
Blacklight - no sidebar column

## Item show page with a sidebar containing an application-specific sidebar box
Blacklight - with sidebar column

## "md" (tablet) viewport width
I think the tools will still fit at this width, so no adjustment needed.

Screen Shot 2020-07-24 at 4 25 12 PM

## "sm" viewport width 
At the `sm` viewport width, the buttons should be collapsed:

Screen Shot 2020-07-24 at 4 26 11 PM

## "sm" viewport width - expanded
And when they are expanded:

- Add `bg-light` to give the expanded section a light gray background so it's more clear to the user what the expand action has done
- Left-align the buttons in a vertical list

Screen Shot 2020-07-24 at 4 27 08 PM

dnoneill:
Add a new function called add_show_header_tools_partial. This will allow the user to add tools to the page header. 

When the app is spun up nothing changes. Changes only go into effect if the user enables these settings.

Screenshot 2024-08-15 at 11 10 15 AM

add_show_header_tools_partial small screen
Screenshot 2024-08-15 at 11 06 45 AM

### DIFF
--- a/app/assets/stylesheets/blacklight/_bookmark.scss
+++ b/app/assets/stylesheets/blacklight/_bookmark.scss
@@ -3,6 +3,12 @@ label.toggle-bookmark {
 	min-width: 8.5em;
 }
 
+/* override for line 3.
+Creates weird spacing in toolbar when min-width is set to 8rem */
+.header-tools label.toggle-bookmark {
+  min-width: 2rem;
+}
+
 .bookmark-toggle {
 	.no-js & {
 		input[type="submit"] {
--- a/app/components/blacklight/document/page_header_component.html.erb
+++ b/app/components/blacklight/document/page_header_component.html.erb
@@ -1,0 +1,7 @@
+<%= render applied_params_component %>
+<div class="<%= header_container_classes %>">
+  <div class="<%= pagination_container_classes %>">
+    <%= render pagination_component %>
+  </div>
+  <%= render_header_tools %>
+</div>
--- a/app/components/blacklight/document/page_header_component.rb
+++ b/app/components/blacklight/document/page_header_component.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+module Blacklight
+  module Document
+    # Render the start over and prev/next displays
+    class PageHeaderComponent < Blacklight::Component
+      attr_reader :document, :blacklight_config, :search_context, :search_session
+
+      delegate :blacklight_config, to: :helpers
+
+      def initialize(document:, search_context:, search_session:)
+        super
+        @search_context = search_context
+        @search_session = search_session
+        @document = document
+      end
+
+      def render?
+        search_context.present? || search_session.present? || has_header_tools?
+      end
+
+      def applied_params_component
+        return unless blacklight_config.track_search_session.applied_params_component
+
+        blacklight_config.track_search_session.applied_params_component.new
+      end
+
+      def pagination_component
+        return unless blacklight_config.track_search_session.item_pagination_component
+
+        blacklight_config.track_search_session.item_pagination_component.new(search_context: search_context, search_session: search_session, current_document: document)
+      end
+
+      def has_header_tools?
+        header_actions.any? || show_header_tools_component
+      end
+
+      def pagination_container_classes
+        has_header_tools? ? 'col-12 col-md-6 ms-auto' : ''
+      end
+
+      def header_container_classes
+        has_header_tools? ? 'row pagination-search-widgets pb-2' : 'pagination-search-widgets'
+      end
+
+      def header_actions
+        actions = helpers.filter_partials(blacklight_config.view_config(:show).header_actions, { document: document })
+        actions.map { |_k, v| v }
+      end
+
+      def show_header_tools_component
+        blacklight_config.view_config(:show).show_header_tools_component
+      end
+
+      def default_action_component_render
+        render Blacklight::Document::ActionsComponent.new(document: document,
+                                                          tag: action_component_tag,
+                                                          classes: classes,
+                                                          link_classes: link_classes,
+                                                          actions: header_actions,
+                                                          url_opts: Blacklight::Parameters.sanitize(params.to_unsafe_h))
+      end
+
+      def action_component_tag
+        'div'
+      end
+
+      def classes
+        'd-inline-flex header-tools align-items-center col-12 col-md-6 ms-auto justify-content-md-end'
+      end
+
+      def link_classes
+        'btn btn-outline-primary ms-2'
+      end
+
+      def render_header_tools
+        return unless has_header_tools?
+
+        return render show_header_tools_component.new(document: document) if show_header_tools_component
+
+        default_action_component_render
+      end
+    end
+  end
+end
--- a/app/helpers/blacklight/component_helper_behavior.rb
+++ b/app/helpers/blacklight/component_helper_behavior.rb
@@ -37,6 +37,10 @@ module Blacklight
       filter_partials(blacklight_config.view_config(:show).document_actions, { document: document }.merge(options)).map { |_k, v| v }
     end
 
+    def filter_partials(partials, options)
+      partials.select { |_, config| blacklight_configuration_context.evaluate_if_unless_configuration config, options }
+    end
+
     private
 
     def render_filtered_partials(partials, options = {})
@@ -51,10 +55,6 @@ module Blacklight
         end
       end
       safe_join(content, "\n") unless block_given?
-    end
-
-    def filter_partials(partials, options)
-      partials.select { |_, config| blacklight_configuration_context.evaluate_if_unless_configuration config, options }
     end
   end
 end
--- a/app/views/catalog/_show_main_content.html.erb
+++ b/app/views/catalog/_show_main_content.html.erb
@@ -1,4 +1,4 @@
-<%= render blacklight_config.track_search_session.item_pagination_component.new(search_context: @search_context, search_session: search_session, current_document: @document) if blacklight_config.track_search_session.item_pagination_component %>
+<%= render blacklight_config.view_config(:show).document_header_component.new(document: @document, search_context: @search_context, search_session: search_session) %>
 <% @page_title = t('blacklight.search.show.title', document_title: document_presenter(@document).html_title, application_name: application_name).html_safe %>
 <% content_for(:head) { render_link_rel_alternates } %>
 
--- a/lib/blacklight/configuration.rb
+++ b/lib/blacklight/configuration.rb
@@ -201,6 +201,8 @@ module Blacklight
         # in Blacklight 9, the default show_tools_component configuration will
         # be Blacklight::Document::ShowToolsComponent
         show_tools_component: nil,
+        show_header_tools_component: nil,
+        document_header_component: Blacklight::Document::PageHeaderComponent,
         sidebar_component: Blacklight::Document::SidebarComponent,
         display_type_field: nil,
         # the "field access" key to use to look up the document display fields
@@ -211,7 +213,8 @@ module Blacklight
         route: nil,
         # partials to render for each document(see #render_document_partials)
         partials: [],
-        document_actions: NestedOpenStructWithHashAccess.new(ToolConfig)
+        document_actions: NestedOpenStructWithHashAccess.new(ToolConfig),
+        header_actions: NestedOpenStructWithHashAccess.new(ToolConfig)
       )
 
       # @!attribute action_mapping
@@ -558,6 +561,22 @@ module Blacklight
       add_action(show.document_actions, name, opts)
       klass && ActionBuilder.new(klass, name, opts).build
     end
+
+    # Add a partial to the show page header when rendering a document.
+    # @!macro partial_if_unless
+    #   @param name [String] the name of the document partial
+    #   @param opts [Hash]
+    #   @option opts [Class] :component draw a component
+    #   @option opts [String] :partial partial to draw if component is false
+    #   @option opts [Symbol,Proc] :if render this action if the method identified by the symbol or the proc evaluates to true. The proc will receive the action configuration and the document or documents for the action.
+    #   @option opts [Symbol,Proc] :unless render this action unless the method identified by the symbol or the proc evaluates to true. The proc will receive the action configuration and the document or documents for the action.
+    def add_show_header_tools_partial(name, opts = {})
+      opts[:partial] ||= 'document_action'
+
+      add_action(show.header_actions, name, opts)
+      klass && ActionBuilder.new(klass, name, opts).build
+    end
+
     # rubocop:enable Layout/LineLength
 
     # Add a tool for the search result list itself
--- a/app/components/blacklight/search_context/server_item_pagination_component.html.erb
+++ b/app/components/blacklight/search_context/server_item_pagination_component.html.erb
@@ -1,10 +1,7 @@
-<div class='pagination-search-widgets'>
+<div class="search-context page-links">
+  <%= link_to_previous_document %> |
 
-  <div class="page-links">
-    <%= link_to_previous_document %> |
+  <%= item_page_entry_info %> |
 
-    <%= item_page_entry_info %> |
-
-    <%= link_to_next_document %>
-  </div>
+  <%= link_to_next_document %>
 </div>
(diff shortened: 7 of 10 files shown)


jcoyne:
@dnoneill can you post some screenshots that also have the "more like this" component?

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,49 @@
...
+        @search_session = search_session
+        @blacklight_config = blacklight_config
+        @document = document
+      end
+
+      def render?
+        search_context.present? || search_session.present? || has_header_actions
+      end
+
+      def before_render
```
jcoyne, line 20: why is this `before_render` Can we make a method `has_header_actions?` (and remove the `attr_reader`)

app/helpers/blacklight/component_helper_behavior.rb
```diff
@@ -37,6 +37,10 @@ def document_actions(document, options: {})
       filter_partials(blacklight_config.view_config(:show).document_actions, { document: document }.merge(options)).map { |_k, v| v }
     end
 
+    def header_actions(document, options: {})
```
jcoyne, line 40: Is there a reason we need this helper?  Can we just have it in the component?  We've been trying to move away from helper methods.

app/helpers/blacklight/component_helper_behavior.rb
```diff
@@ -37,6 +37,10 @@ def document_actions(document, options: {})
...
+      filter_partials(blacklight_config.view_config(:show).header_actions, { document: document }.merge(options)).map { |_k, v| v }
```
jcoyne, line 41: Is it possible to do?
```suggestion
      filter_partials(blacklight_config.view_config(:show).header_actions, { document: document }.merge(options)).values
```

app/views/catalog/_show_main_content.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render blacklight_config.track_search_session.item_pagination_component.new(search_context: @search_context, search_session: search_session, current_document: @document) if blacklight_config.track_search_session.item_pagination_component %>
+<%= render Blacklight::Document::PageHeaderComponent.new(document: @document, blacklight_config:, search_context: @search_context, search_session:) unless blacklight_config.full_width_header? %>
```
jcoyne: Can we push this class into the config? Can we move the conditional into the component's `render?` method?

lib/blacklight/configuration.rb
```diff
@@ -156,6 +156,12 @@ def initialized_default_configuration?
       # set to Blacklight::Icons::BookmarkIconComponent to replace checkbox with icon
       property :bookmark_icon_component, default: nil
 
+      # @!attribute full_width_header?
+      # @since v8.3.1
+      # When true, doesn't place the header in col-9 documents but outside as own div
+      # @return [Boolean]
+      property :full_width_header?, default: false
```
jcoyne, line 163: I don't think config properties usually have question marks, do they?

lib/blacklight/configuration.rb
```diff
@@ -558,6 +565,22 @@ def add_show_tools_partial(name, opts = {})
...
+
+    # Add a partial to the header when rendering a document.
+    # @!macro partial_if_unless
+    #   @param name [String] the name of the document partial
+    #   @param opts [Hash]
+    #   @option opts [Class] :component draw a component
+    #   @option opts [String] :partial partial to draw if component is false
+    #   @option opts [Symbol,Proc] :if render this action if the method identified by the symbol or the proc evaluates to true. The proc will receive the action configuration and the document or documents for the action.
+    #   @option opts [Symbol,Proc] :unless render this action unless the method identified by the symbol or the proc evaluates to true. The proc will receive the action configuration and the document or documents for the action.
+    def add_header_tools_partial(name, opts = {})
```
jcoyne, line 577: I think we want to make this clear that this is the "show" header.  Not the header on the index view.

app/assets/stylesheets/blacklight/_bookmark.scss
```diff
@@ -37,4 +37,8 @@ label.toggle-bookmark {
   &.bookmark-unchecked {
     display: none;
   }
+}
+
+.header-tools label.toggle-bookmark {
+  min-width: 2rem;
```
jcoyne, line 43: What does this style do for us? What happens without it?

app/components/blacklight/document/page_header_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= render blacklight_config.track_search_session.applied_params_component.new if blacklight_config.track_search_session.applied_params_component %>
```
jcoyne, line 1: Can this be moved into a method on the component? Something like `<%= applied_params %>`

jcoyne (changes requested):

app/views/catalog/_show_main_content.html.erb
dnoneill: No because it is shows up in two different places. If it is full_width we want it to render, just somewhere else.

app/assets/stylesheets/blacklight/_bookmark.scss
dnoneill: The bookmark gets moved really far to the right. I am overriding another css class that I am assuming is to keep tools formatted properly https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/assets/stylesheets/blacklight/_bookmark.scss#L3

app/components/blacklight/document/page_header_component.rb
dnoneill: done.

app/helpers/blacklight/component_helper_behavior.rb
dnoneill: Nope causes an error. Something about how the ActionsComponent is set up doesn't like it.

lib/blacklight/configuration.rb
dnoneill: how do you like add_show_header_tools_partial?

app/helpers/blacklight/component_helper_behavior.rb
dnoneill: Done.

app/views/catalog/_show_main_content.html.erb
dnoneill: But yes we can push the class into config.

dnoneill:
@jcoyne I added your suggestions except for the unresolved ones. I also added tests for the PageHeaderComponent and added a show_header_tools_component to the blacklight config.

app/assets/stylesheets/blacklight/_bookmark.scss
jcoyne: Can you add a comment into the CSS file so that other people will understand the reasoning for this rule?

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,73 @@
...
+        @search_session = search_session
+        @document = document
+      end
+
+      def render?
+        search_context.present? || search_session.present? || has_header_tools?
+      end
+
+      def before_render
+        @blacklight_config = helpers.blacklight_config
```
jcoyne, line 21: Do we need this `before_render` method?  Can we just delegate `blacklight_config` to `helpers`?

lib/blacklight/configuration.rb
```diff
@@ -156,6 +156,12 @@ def initialized_default_configuration?
       # set to Blacklight::Icons::BookmarkIconComponent to replace checkbox with icon
       property :bookmark_icon_component, default: nil
 
+      # @!attribute full_width_header
+      # @since v8.3.1
+      # When true, doesn't place the header in col-9 documents but outside as own div
```
jcoyne, line 161: I'm not sure it's clear to say "in col-9 documents".  Can you be more descriptive here?

Is there a reason we need to support this flag? Can we do it just one way?  This sort of configuration makes it challenging to test as it adds a whole new dimension that we should be testing for.  Is it possible for just your application to override the correct components to get the layout as you need?

jcoyne (changes requested):

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,73 @@
...
+      def render_header_tools
+        return unless has_header_tools?
+
+        show_header_tools_component&.tap do |show_tools_component_class|
+          return render show_tools_component_class.new(document: document)
+        end
+
+        render Blacklight::Document::ActionsComponent.new(document: document,
+                                                          tag: 'div',
+                                                          classes: 'd-inline-flex header-tools align-items-center col-12 col-md-6 ms-auto justify-content-md-end',
```
jcoyne, line 66: Extracting these classes (as well as the component class itself) to a separate method makes it easier for people to configure via overriding.

app/components/blacklight/document/page_header_component.rb
dnoneill: We did not. I didn't know you could delegate in components without causing errors. Thanks for the tip.

app/components/blacklight/document/page_header_component.rb
dnoneill: Done.

app/assets/stylesheets/blacklight/_bookmark.scss
```diff
@@ -37,4 +37,10 @@ label.toggle-bookmark {
   &.bookmark-unchecked {
     display: none;
   }
+}
+
+/* override for line 3.
```
jcoyne, line 42: Can we move this rule up by line 3 so that it's easier to see the connection?

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      def initialize(document:, search_context:, search_session:)
+        super
+        @search_context = search_context
+        @search_session = search_session
+        @document = document
+      end
+
+      def render?
+        search_context.present? || search_session.present? || has_header_tools?
```
jcoyne, line 19: Is there a reason we need this method?  Won't `search_session` always exist, which means this is always truthful?  See https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/controllers/concerns/blacklight/search_context.rb#L56-L61

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,87 @@
...
+      def applied_params_component
+        return unless blacklight_config.track_search_session.applied_params_component
+
+        blacklight_config.track_search_session.applied_params_component.new
+      end
+
+      def pagination_component
+        return unless blacklight_config.track_search_session.item_pagination_component
+
+        blacklight_config.track_search_session.item_pagination_component.new(search_context: search_context, search_session: search_session, current_document: document)
```
jcoyne, line 31:
```suggestion
        blacklight_config.track_search_session.item_pagination_component.new(search_context:, search_session:, current_document: document)
```

jcoyne (changes requested):

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,87 @@
...
+      def has_header_tools?
+        header_actions.any? || show_header_tools_component
+      end
+
+      def pagination_container_classes
+        has_header_tools? ? 'col-12 col-md-6 ms-auto' : ''
+      end
+
+      def header_container_classes
+        has_header_tools? ? 'd-flex justify-content-between pagination-search-widgets pb-2 row justify-content-end' : 'pagination-search-widgets'
```
jcoyne, line 43: I don't understand why we use both grid classes `row` and flex classes here.  I don't understand why we'd want both `justify-content-between` and `justify-content-end`.  Don't those override each other?

app/components/blacklight/document/page_header_component.rb
dnoneill: I did this originally. Rubocop complains.

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+
+      def header_actions
+        actions = helpers.send(:filter_partials, blacklight_config.view_config(:show).header_actions, { document: document })
```
jcoyne, line 47: Please change the visibility of `filter_partials` or move it in here. Don't use `send` to override the encapsulation.

jcoyne (dismissed):

app/components/blacklight/document/page_header_component.rb
jcoyne: I see that in some `.erb` files you did this too.  So can we be consistent in the style we're using?

app/components/blacklight/document/page_header_component.rb
dnoneill: cruft.

app/components/blacklight/document/page_header_component.rb
dnoneill: Actually present? works for empty hashes, so I changed it back.

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,87 @@
...
+      def default_action_component_render
+        render Blacklight::Document::ActionsComponent.new(document: document,
+                                                          tag: tag,
+                                                          classes: classes,
+                                                          link_classes: link_classes,
+                                                          actions: header_actions,
+                                                          url_opts: Blacklight::Parameters.sanitize(params.to_unsafe_h))
+      end
+
+      def tag
```
jcoyne, line 64: Can you change the name of this method. I suspect it may be confusing with the `tag` heleper method https://api.rubyonrails.org/v5.1.7/classes/ActionView/Helpers/TagHelper.html#method-i-tag

jcoyne (changes requested):

app/components/blacklight/document/page_header_component.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+
+      def link_classes
+        'btn btn-outline-primary ms-2'
+      end
+
+      def render_header_tools
+        return unless has_header_tools?
+
+        show_header_tools_component&.tap do |show_tools_component_class|
```
jcoyne, line 79: Is this the same as `return render show_header_tools_component.new(document: document) if show_header_tools_component` ?  I think writing it that way is clearer.

dnoneill: True but this is the way it is in show_tools so I figured we would match? https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/app/components/blacklight/document/sidebar_component.rb#L20-L22

app/components/blacklight/document/page_header_component.rb
dnoneill: updated to action_component_tag

tpendragon (approved):

jcoyne (approved):

Issue #3238, opened 2024-07-29
Facet checkbox unselected when advanced search form receives incoming f_inlcusive parameters 


mamrey:
We can [pass existing parameters to the advanced search form](https://github.com/projectblacklight/blacklight/blob/a158b32f52098e5e3927be7cded5c495ded648d7/spec/features/advanced_search_spec.rb#L106) to populate the form with our existing search context. 

However, when the blacklight advanced search form component receives incoming `f_inclusive` facet parameters, the corresponding facet checkbox component remains unselected. Moreover, clicking "more" to reveal more facets, renders these incoming inclusive facets in the popup modal using the `FacetFieldInclusiveConstraintComponent.` Is this the expected behavior?

To address this we can update how we determine whether these [checkboxes are selected when rendered](https://github.com/projectblacklight/blacklight/blob/a158b32f52098e5e3927be7cded5c495ded648d7/app/components/blacklight/facet_field_checkboxes_component.html.erb#L11) in the [FacetItemPresenter class.](https://github.com/projectblacklight/blacklight/blob/main/app/presenters/blacklight/facet_item_presenter.rb#L4)

To reproduce:

- [Go to advanced search page with incoming f_inclusive parameters](https://demo.projectblacklight.org/catalog/advanced?op=must&clause[0][field]=all_fields&clause[0][query]=&clause[1][field]=title&clause[1][query]=&clause[2][field]=author&clause[2][query]=&clause[3][field]=subject&clause[3][query]=&f_inclusive[format][]=Book&sort=relevance&commit=Search)

- observe incoming "Book" format facet remains unselected

mamrey:
closed with #3278

PR #3278, opened 2024-08-15
Improve how adv search handles incoming params

### LINKED ISSUE #3237: Submitting an advanced search with existing parameters duplicates parameters in request
We can [pass existing parameters to the advanced search form](https://github.com/projectblacklight/blacklight/blob/a158b32f52098e5e3927be7cded5c495ded648d7/spec/features/advanced_search_spec.rb#L106) to populate the form with our existing search context. 

The BL advanced search form component [renders a hidden search state component](https://github.com/projectblacklight/blacklight/blob/a158b32f52098e5e3927be7cded5c495ded648d7/app/components/blacklight/advanced_search_form_component.html.erb#L11) that creates hidden control fields from these incoming parameters.

The form also populates the visible control fields with their corresponding incoming values. This leads to duplicate parameters in the request. 

Besides producing extra long URLs, this raises issues for facet display on the search results page, duplicating the facets selected and rendered. 

A possible solution would be to ensure that hidden inputs do not contain the same parameters as the visible inputs. 

To reproduce:

- Go to advanced search form with incoming params [here](https://demo.projectblacklight.org/catalog/advanced?op=must&clause[0][field]=all_fields&clause[0][query]=&clause[1][field]=title&clause[1][query]=Maine+&clause[2][field]=author&clause[2][query]=&clause[3][field]=subject&clause[3][query]=&f_inclusive[format][]=Book&sort=relevance&commit=Search)

```
https://demo.projectblacklight.org/catalog/advanced?op=must&clause[0][field]=all_fields&clause[0][query]=&clause[1][field]=title&clause[1][query]=Maine+&clause[2][field]=author&clause[2][query]=&clause[3][field]=subject&clause[3][query]=&f_inclusive[format][]=Book&sort=relevance&commit=Search
```
- Click search
- observe that the url has duplicate params
- observer duplicate facets selected in side panel
```
https://demo.projectblacklight.org/?sort=relevance&op=must&clause[0][field]=all_fields&clause[0][query]=&clause[1][field]=title&clause[1][query]=Maine+&clause[2][field]=author&clause[2][query]=&clause[3][field]=subject&clause[3][query]=&f_inclusive[format][]=Book&op=must&clause[0][field]=all_fields&clause[0][query]=&clause[1][field]=title&clause[1][query]=Maine+&clause[2][field]=author&clause[2][query]=&clause[3][field]=subject&clause[3][query]=&sort=relevance&commit=Search
```

mamrey:
closes #3237, #3238 

### Changes
- excludes adv search form parameters from hidden search fields to prevent duplicate query params 
- pre selects facet appropriate facet checkbox _only_ when receiving inclusive filter parameters
  - implements check using new `FacetCheckboxItemPresenter` that overrides `#selected?`
- increases spec coverage when prepopulating adv search form
- adds specs for new presenter 

### Questions
I have a couple lingering questions that would help clarify my understanding of the expected behavior of the adv search and provide additional context for a review: 

Currently, incoming exclusive filter parameters will pre-check the corresponding facet checkbox. This PR prevents this behavior, and now any incoming exclusive filters will continue to be added as a hidden field. What is the desired behavior around incoming exclusive filters? In my opinion, the facet checkboxes represent inclusive filters, so it's a bit misleading to have them pre selected with an incoming exclusive filter. Any thoughts on this?

Currently, we exclude incoming inclusive parameters from the displayed constraints on the adv search page. This PR does not change this, but is this desired behavior?

### DIFF
--- a/spec/presenters/blacklight/facet_checkbox_item_presenter_spec.rb
+++ b/spec/presenters/blacklight/facet_checkbox_item_presenter_spec.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Blacklight::FacetCheckboxItemPresenter, type: :presenter do
+  subject(:presenter) do
+    described_class.new(facet_item, facet_config, view_context, facet_field, search_state)
+  end
+
+  let(:facet_item) { Blacklight::Solr::Response::Facets::FacetItem.new(value: 'Book', hits: 30) }
+  let(:facet_config) { Blacklight::Configuration::FacetField.new(key: 'format') }
+  let(:view_context) { controller.view_context }
+  let(:filter_field) { Blacklight::SearchState::FilterField.new(facet_config, search_state) }
+  let(:facet_field) { Blacklight::Solr::Response::Facets::FacetField.new('format', [facet_item]) }
+  let(:params) { ActionController::Parameters.new(f_inclusive: { format: ["Book"] }) }
+  let(:blacklight_config) { Blacklight::Configuration.new }
+  let(:search_state) { Blacklight::SearchState.new(params, blacklight_config) }
+
+  before do
+    blacklight_config.add_facet_field 'format'
+  end
+
+  describe '#selected?' do
+    subject { presenter.selected? }
+
+    context 'with a matching inclusive filter' do
+      it 'returns true' do
+        expect(subject).to be true
+      end
+    end
+
+    context 'with an inclusive filter that does not match' do
+      let(:params) { ActionController::Parameters.new(f_inclusive: { format: ["Manuscript"] }) }
+
+      it 'returns true' do
+        expect(subject).to be false
+      end
+    end
+
+    context 'with a matching exclusive filter' do
+      let(:params) { ActionController::Parameters.new(f: { format: ["Book"] }) }
+
+      it 'returns false' do
+        expect(subject).to be false
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


spec/presenters/blacklight/facet_checkbox_item_presenter_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    context 'with a matching inclusive filter' do
+      it 'returns true' do
+        expect(subject).to be true
+      end
+    end
+
+    context 'with an inclusive filter that does not match' do
+      let(:params) { ActionController::Parameters.new(f_inclusive: { format: ["Manuscript"] }) }
+
+      it 'returns true' do
```
jcoyne, line 35: The description doesn't match the assertion here.

I prefer `is { is_expected.to be false }` in simple cases like this.  Then no description is needed.

jcoyne (dismissed):

spec/presenters/blacklight/facet_checkbox_item_presenter_spec.rb
mamrey: Good catch, thanks! Done.

tampakis:
> Currently, we exclude incoming inclusive parameters from the displayed constraints on the adv search page. This PR does not change this, but is this desired behavior?

I think the inclusive facet parameters should also carry over. Could we track adding that capability in a separate issue?

cgalarza:
> > Currently, we exclude incoming inclusive parameters from the displayed constraints on the adv search page. This PR does not change this, but is this desired behavior?
> 
> I think the inclusive facet parameters should also carry over. Could we track adding that capability in a separate issue?

@mamrey and I discussed this and we think we should only display parameters that are not already selected on the form in the displayed constraints at the top of the advanced search form. The inclusive facet parameters will be selected on the advanced search form so there isn't a need to display them again in the displayed constraints. @mamrey is going to write up a ticket explaining this issue in more detail.

tampakis (approved):

PR #3282, opened 2024-08-21
Remove unused parameter.

jcoyne:
`render` doesn't do anything with an `object` argument


jcoyne:
Actually, this gets assigned to the `paginate_compact` local.

PR #3283, opened 2024-08-22
move start over button to setting

dnoneill:
Allow the start over button to be set.

### DIFF
--- a/lib/blacklight/configuration.rb
+++ b/lib/blacklight/configuration.rb
@@ -151,6 +151,8 @@ module Blacklight
       # @return [#partials]
       property :navbar, default: OpenStructWithHashAccess.new(partials: {})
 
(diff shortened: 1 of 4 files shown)


jcoyne (commented):
This seems like a backwards incompatible change. Do I have that right?

lib/blacklight/configuration.rb
```diff
@@ -151,6 +151,8 @@ def initialized_default_configuration?
       # @return [#partials]
       property :navbar, default: OpenStructWithHashAccess.new(partials: {})
 
+      property :start_over_component, default: Blacklight::StartOverButtonComponent.new
```
jcoyne, line 154: We typically never put component instances in the configuration.

jcoyne (changes requested):

PR #3287, opened 2024-08-22
This style doesn't change anything

jcoyne:
It should be removed so that downstream apps can more easily customize the style of this element without fighting against Blacklight


sandbergja (approved):
Thanks, @jcoyne !

PR #3293, opened 2024-08-23
Remove custom styling for card-body

jcoyne:
It is unnecessary, and we can configure it using the bootstrap variables: 
`padding: var(--bs-card-spacer-y) var(--bs-card-spacer-x);`


jrochkind (approved):

jcoyne:
No longer needed after switching to accodion widget for facets.

PR #3304, opened 2024-08-29
fix show label in metadatadocument

### LINKED ISSUE #3303: When viewing a show page, the @show = false and it is trying to get the translation for blacklight.search.index..label
This line seems to be false
https://github.com/projectblacklight/blacklight/blob/69373f202753f04ec6aca179fdf8fa01248a9edf/app/components/blacklight/metadata_field_component.rb#L14

which means this line https://github.com/projectblacklight/blacklight/blob/69373f202753f04ec6aca179fdf8fa01248a9edf/app/components/blacklight/metadata_field_component.rb#L19 never gets hit

dnoneill:
closes #3303

### DIFF
--- a/spec/components/blacklight/document_component_spec.rb
+++ b/spec/components/blacklight/document_component_spec.rb
@@ -158,6 +158,26 @@ RSpec.describe Blacklight::DocumentComponent, type: :component do
       expect(rendered).to have_content 'embed'
     end
 
+    context 'show view with custom translation' do
+      let!(:original_translations) { I18n.backend.send(:translations).deep_dup }
+
+      before do
+        controller.action_name = "show"
+        I18n.backend.store_translations(:en, blacklight: { search: { show: { label: "testing:%{label}" } } })
+      end
+
+      after do
+        I18n.backend.reload!
+        I18n.backend.store_translations(:en, original_translations[:en])
+      end
+
+      it 'renders with show-specific metadata with correct translation' do
+        expect(rendered).to have_css 'dl.document-metadata'
+        expect(rendered).to have_css 'dt', text: 'testing:ISBN'
+        expect(rendered).to have_css 'dd', text: 'Value'
+      end
+    end
+
     context 'with configured metadata component' do
       let(:custom_component_class) do
         Class.new(Blacklight::DocumentMetadataComponent) do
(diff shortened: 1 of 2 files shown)


marlo-longley (approved):

jcoyne (changes requested):
Are you able to add a test to this PR that shows the fix?

dnoneill:
Yes. Do you want me to include screenshots. I tested it locally to make sure it was working.

jcoyne:
@dnoneill I'd like an automated test please. That way we don't have any accidental regressions.

dnoneill:
Gotcha. Can do.

spec/components/blacklight/document_component_spec.rb
```diff
@@ -158,6 +158,19 @@ def call
       expect(rendered).to have_content 'embed'
     end
 
+    context 'show view with custom translation' do
+      before do
+        controller.action_name = "show"
+        I18n.backend.store_translations(:en, blacklight: { search: { show: { label: "testing:%{label}" } } })
```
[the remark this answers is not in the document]
dnoneill: Fixed.

spec/components/blacklight/document_component_spec.rb
```diff
@@ -158,6 +158,26 @@ def call
       expect(rendered).to have_content 'embed'
     end
 
+    context 'show view with custom translation' do
+      let(:original_translations) { I18n.backend.send(:translations).deep_dup }
```
jcoyne, line 162: Don't we need to do this prior to adding the new translation? Won't you need `let!` to do that?

jcoyne: I'm trying to understand how this is working to restore the original traslations.  `let` blocks are called lazily, so here we're not resolving the original translations until the `after` block.  Don't we need them cached first?

dnoneill: I originally tried setting it in before do like this:
before do
@original_translations = I18n.backend.send(:translations).deep_dup 
end

and then accessing it in after with @original_translations. Unfortunately rubocop started complaining and suggested using let. I moved it to the context because my assumption (which would be wrong) is it would be okay because context variables are set before before do is done.

jcoyne: If you use `let!` instead of `let` then it does get the value eagerly, and it should work.

jcoyne (approved):

PR #3311, opened 2024-09-03
Ignore Rubocop false positive on empty literal

jcoyne:
See https://github.com/rubocop/rubocop/issues/13189


sandbergja (approved):
Thanks, @jcoyne

PR #3312, opened 2024-09-03
Each key makes it's own list

jcoyne:
This fixes a misunderstanding of how Hash.new(obj) works. That means each key gets the same instance of the list rather than a new list


sandbergja (approved):
Thanks, @jcoyne

PR #3315, opened 2024-09-05
adding option for single page count

### LINKED ISSUE #3314: Enable display of pagination component in the case where we have only one result
Currently the pagination component does not show if the record is the only result for a search.  We would like to enable the pagination component to display if there is only one result, in which case it would only show "1 of 1" without any previous or next links. 

See https://github.com/sul-dlss/earthworks/issues/1367 .

hudajkhan:
What this pull request does:

Adds a check to see if there is only one search result, and shows "1 of 1" in the view without showing previous or next links.
Adds a test for this particular case. 

Closes #3314 .
Addresses a ticket in EarthWorks: https://github.com/sul-dlss/earthworks/issues/1367

### DIFF
--- a/app/components/blacklight/search_context/server_item_pagination_component.rb
+++ b/app/components/blacklight/search_context/server_item_pagination_component.rb
@@ -12,7 +12,11 @@ module Blacklight
       end
 
       def render?
-        @search_context.present? && (@search_context[:prev] || @search_context[:next]) && (@search_session['document_id'] == @current_document_id)
+        return false unless (@search_session['document_id'] == @current_document_id)
+
+        return true if total == 1
+        
+        @search_context.present? && (@search_context[:prev] || @search_context[:next]) 
       end
 
       ##
@@ -20,25 +24,34 @@ module Blacklight
       #
       # @return [String]
       def item_page_entry_info
-        t('blacklight.search.entry_pagination_info.other', current: number_with_delimiter(count),
+        t('blacklight.search.entry_pagination_info.other', current: current_count,
                                                            total: number_with_delimiter(total),
                                                            count: total).html_safe
       end
 
       def link_to_previous_document(previous_document = nil, classes: 'previous', **link_opts)
+        return if total == 1
+
         previous_document ||= @search_context[:prev]
         link_opts = session_tracking_params(previous_document, count - 1, per_page: per_page, search_id: search_id).merge(class: classes, rel: 'prev').merge(link_opts)
+       
         link_to_unless previous_document.nil?, raw(t('views.pagination.previous')), url_for_document(previous_document), link_opts do
           tag.span raw(t('views.pagination.previous')), class: 'previous'
         end
       end
 
       def link_to_next_document(next_document = nil, classes: 'next', **link_opts)
+        return if total == 1
+        
         next_document ||= @search_context[:next]
         link_opts = session_tracking_params(next_document, count + 1, per_page: per_page, search_id: search_id).merge(class: classes, rel: 'next').merge(link_opts)
         link_to_unless next_document.nil?, raw(t('views.pagination.next')), url_for_document(next_document), link_opts do
           tag.span raw(t('views.pagination.next')), class: 'next'
         end
+      end
+
+      def current_count
+        (total == 1)? total : number_with_delimiter(count)
       end
 
       private
--- a/spec/components/blacklight/search_context/server_item_pagination_component_spec.rb
+++ b/spec/components/blacklight/search_context/server_item_pagination_component_spec.rb
@@ -23,6 +23,17 @@ RSpec.describe Blacklight::SearchContext::ServerItemPaginationComponent, type: :
     end
   end
 
+  context 'when there is exactly one search result with no next or previous document' do
+    let(:search_context) { {} }
+    let(:search_session) { { 'document_id' => current_document_id, 'total' => '1' } }
+
+    it "renders single page count" do
+      expect(render.to_html).to include '<strong>1</strong> of <strong>1</strong>'
+      expect(render.css('span.previous').to_html).to be_blank
+      expect(render.css('span.next').to_html).to be_blank
+    end
+  end
+
   context 'when there is next and previous' do
     let(:search_context) { { next: next_doc, prev: prev_doc } }
     let(:prev_doc) { SolrDocument.new(id: '777') }


app/components/blacklight/search_context/server_item_pagination_component.rb
```diff
@@ -12,35 +12,48 @@ def initialize(search_context:, search_session:, current_document:)
       end
 
       def render?
-        @search_context.present? && (@search_context[:prev] || @search_context[:next]) && (@search_session['document_id'] == @current_document_id)
+        return false unless @search_session['document_id'] == @current_document_id
+
+        return true if total == 1
+
+        @search_context.present? && (@search_context[:prev] || @search_context[:next])
```
thatbudakguy, line 19: is this effectively the same as:
```suggestion
        @search_context.present?
```

thatbudakguy: Right – my guess is that case (you have a context, but no prev or next) is the same thing as having a single result. But just a guess!

hudajkhan: Separately I have to remove the "|" separating the different sections, so that will be another commit.

app/components/blacklight/search_context/server_item_pagination_component.rb
```diff
@@ -12,35 +12,52 @@ def initialize(search_context:, search_session:, current_document:)
...
+      end
+
+      def display_separator
+        '|' unless total == 1
```
dnoneill, line 23: I would make this display separator a translation.

hudajkhan: I have tried to do that.

spec/components/blacklight/search_context/server_item_pagination_component_spec.rb
```diff
@@ -23,6 +23,18 @@
     end
   end
 
+  context 'when there is exactly one search result with no next or previous document' do
```
jcoyne, line 26: What if someone goes directly to a show page without any session history.  Will they still see this?

jcoyne:
For anyone else who is interested in the history, this conditional originated in the initial git commit 15 years aog: https://github.com/projectblacklight/blacklight/commit/cfc474389df64ffb61b69f8a2127b93ac19d07ab#diff-c732a210586cc78a766262d89f052850316eef0ddeb8a1fc63020dff9a77a85bR1

app/components/blacklight/search_context/server_item_pagination_component.rb
```diff
@@ -12,35 +12,52 @@ def initialize(search_context:, search_session:, current_document:)
       end
 
       def render?
-        @search_context.present? && (@search_context[:prev] || @search_context[:next]) && (@search_session['document_id'] == @current_document_id)
+        return false unless @search_session['document_id'] == @current_document_id
+
+        return true if total == 1
```
jcoyne, line 17: I believe we need to check for `@search_context` here. Otherwise it will display on pages with no search context.

jcoyne (changes requested):

hudajkhan:
Thanks for the comments @jcoyne. I should be able to look at this today/tomorrow.

spec/components/blacklight/search_context/server_item_pagination_component_spec.rb
hudajkhan: After the most recent changes, I went directly to an item page url in a separate browser and that did not show the "1 of 1" content.

hudajkhan:
@jcoyne I believe I addressed your comments. Thanks!

app/components/blacklight/search_context/server_item_pagination_component.rb
```diff
@@ -12,35 +12,50 @@ def initialize(search_context:, search_session:, current_document:)
       end
 
       def render?
-        @search_context.present? && (@search_context[:prev] || @search_context[:next]) && (@search_session['document_id'] == @current_document_id)
+        return false unless @search_context.present? && @search_session['document_id'] == @current_document_id
+
+        total == 1 || (@search_context[:prev] || @search_context[:next])
```
jcoyne, line 17:
```suggestion
        total == 1 || @search_context[:prev] || @search_context[:next]
```

app/components/blacklight/search_context/server_item_pagination_component.rb
```diff
@@ -12,35 +12,50 @@ def initialize(search_context:, search_session:, current_document:)
...
+      end
+
+      def display_separator
```
jcoyne, line 20: Can we move the calls to this method out of the template and put them in `link_to_previous_document` and `link_to_next_document`? Then we don't need to have a conditional on `total == 1` in this method.

app/components/blacklight/search_context/server_item_pagination_component.rb
```diff
@@ -12,35 +12,50 @@ def initialize(search_context:, search_session:, current_document:)
...
+
         next_document ||= @search_context[:next]
         link_opts = session_tracking_params(next_document, count + 1, per_page: per_page, search_id: search_id).merge(class: classes, rel: 'next').merge(link_opts)
         link_to_unless next_document.nil?, raw(t('views.pagination.next')), url_for_document(next_document), link_opts do
           tag.span raw(t('views.pagination.next')), class: 'next'
         end
       end
 
+      def current_count
+        total == 1 ? total : number_with_delimiter(count)
```
jcoyne, line 56: Do we need a conditional here? 
```suggestion
        number_with_delimiter(count)
```

jcoyne (changes requested):

app/components/blacklight/search_context/server_item_pagination_component.rb
```diff
@@ -20,7 +20,7 @@ def render?
       end
 
       def display_separator
-        '|' unless total == 1
+        t('views.pagination.separator') unless total == 1
```
jcoyne, line 21: Do we really want to configure the delimiter separately for each language that we want to support?  This seems to me like something better suited for Blacklight::Config rather than an i18n.

jcoyne (changes requested):

app/components/blacklight/search_context/server_item_pagination_component.rb
jcoyne: On second thought, that may break compatibility for someone who is expecting this not to have a delimiter in it.  Disregard this.

PR #3318, opened 2024-09-09
Don't add unnecessary bookmark routes

jcoyne:
The "Show" route is not necessary.


tampakis (approved):
Approved during community call 9/18.

PR #3322, opened 2024-09-19
Run prettier on config/locales for consistency

jcoyne:
Now all values that use quotes use double-quotes


package.json
```diff
@@ -22,6 +22,7 @@
   },
   "homepage": "https://github.com/projectblacklight/blacklight#readme",
   "devDependencies": {
+    "prettier": "3.3.3",
```
thatbudakguy, line 25: was this intentional?

jcoyne: 🤷 should I just install it globally?

thatbudakguy (approved):

PR #3327, opened 2024-09-23
Deprecate the start_over partial. Render the component instead

jcoyne:


cbeer:
Unlike #3333, I think this one is ok, right?

jcoyne:
Yes. I think the local in the partial is safer.

PR #3330, opened 2024-09-23
Call the constraints component directly

jcoyne:
Deprecate the catalog/_constraints.html.erb partial


cbeer:
I don't think this change would be backwards compatible (like #3333), right?

jcoyne:
@cbeer Can you explain the situation in which it would not be?

jcoyne:
@cbeer is this better?

PR #3333, opened 2024-09-23
Deprecate render_document_sidebar_partial and catalog/show_sidebar pa…

jcoyne:
…rtial

Use the component configuration instead


sandbergja (approved):
Thanks, @jcoyne !

cbeer:
Won't this break applications that have overridden those methods or partials?

jcoyne:
@cbeer yes, you are right.  Is there a better way to deprecate this call without breaking any overrides?

PR #3341, opened 2024-09-27
Consolidate generated additions to the User class

jcoyne:
Give the string_display_key a default value


sandbergja (approved):
Thanks, @jcoyne !

PR #3343, opened 2024-09-27
Allow the `view_config` to be passed into the `document_presenter` helper and `DocumentComponent` constructor

cbeer:
This allows downstream applications (and e.g. gems like spotlight) a little more flexibility in how they use these.


app/components/blacklight/document_component.rb
```diff
@@ -137,6 +138,12 @@ def before_render
       end
     end
 
+    def view_config
+      @view_config ||= presenter&.view_config || Blacklight::Configuration::ViewConfig.new
```
jcoyne, line 142: Is there really a case where presenter is `nil`?  Wouldn't we also need to deal with that in the `delegate :field_presenters`?

jcoyne:
Can you explain why downstream apps need more flexibility?   In order to do what? 

It seems that this would allow the presenter.view_config to be different than `DocumentComponent#view_config` and that seems like a lot of complexity.

app/components/blacklight/document_component.rb
cbeer: I don't know. Someone thought so in all the `renders_one` blocks above, but not in the `before_render`.

cbeer:
Spotlight is a good example where its use of document presenters and components doesn't neatly align with Blacklight's limited notion of search results and single document pages, and where `document_presenter_class` is really a bad fit:

jcoyne:
@cbeer It seems like what Spotlight has right now is working... so what do you want it to look like?

cbeer:
Yes, spotlight, blacklight-gallery and who-knows-what-else could all provide their own `document_presenter` implementations to just get the document presenter instance they actually want. It seems like an anti-pattern to force each gem to do that when we could have a standardized implementation here for just a tiny change.

jcoyne:
So the main thing here is the addition of `view_config&.document_presenter_class` in `#document_presenter`?  That seems like a good change.  Do you think we should deprecate the call to `document_presenter_class` helper method? Do we still need two ways to do this?

Issue #3350, opened 2024-10-05
Make the bookmark counter badge work for multiple elements


kirkkwang:
Currently, if we want to add another element that displays the bookmark count, it doesn't tick up or down when the bookmark is checked or unchecked.  It would be nice to have the ability to add more than one bookmark count element and have the same functionality.

![bl-bookmark](https://github.com/user-attachments/assets/afa2f6be-1336-4dd2-b0af-b05ae1d327e2)

cbeer:
Was this fixed by #3351?

kirkkwang:
@cbeer ah yes it was, i can close

PR #3351, opened 2024-10-05
Change bookmarkCounter to find multiple selectors

kirkkwang:
This commit will change the bookmarkCounter function to look for multiple selectors instead of one.  This way if we want to put the bookmark counter elsewhere it will also be dynamic.

Ref:
- https://github.com/projectblacklight/blacklight/issues/3350

bl-bookmark-pr

### DIFF
--- a/app/javascript/blacklight/checkbox_submit.js
+++ b/app/javascript/blacklight/checkbox_submit.js
@@ -37,7 +37,9 @@ export default class CheckboxSubmit {
       this.labelTarget.removeAttribute('disabled')
       this.checkboxTarget.removeAttribute('disabled')
       this.updateStateFor(!this.checked)
-      if (this.bookmarkCounter()) this.bookmarkCounter().innerHTML = json.bookmarks.count
+      this.bookmarksCounter().forEach(counter => {
+        counter.innerHTML = json.bookmarks.count;
+      });
     }).catch((error) => {
       this.handleError(error)
     })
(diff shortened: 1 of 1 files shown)


app/javascript/blacklight/checkbox_submit.js
```diff
@@ -37,7 +37,11 @@ export default class CheckboxSubmit {
       this.labelTarget.removeAttribute('disabled')
       this.checkboxTarget.removeAttribute('disabled')
       this.updateStateFor(!this.checked)
-      if (this.bookmarkCounter()) this.bookmarkCounter().innerHTML = json.bookmarks.count
+      if (this.bookmarkCounter()) {
```
sandbergja, line 40: I think we can remove this `if`, since `querySelectorAll` returns a truthy value whether it finds anything or not.

app/javascript/blacklight/checkbox_submit.js
```diff
@@ -64,7 +68,7 @@ export default class CheckboxSubmit {
   }
 
   bookmarkCounter() {
```
sandbergja, line 70:
```suggestion
  bookmarkCounters() {
```

Could we rename the method name, since it could potentially be more than one counter?

sandbergja (commented):
Thanks, @kirkkwang !  Looks like a good change to me, I just had two suggestions.

app/javascript/blacklight/checkbox_submit.js
kirkkwang: yes, you're right!

app/javascript/blacklight/checkbox_submit.js
kirkkwang: Ah yes, that's true, I've made the changes

kirkkwang:
@sandbergja thank you so much! I've made the suggested changes.

sandbergja (approved):
Thanks, @kirkkwang !

PR #3354, opened 2024-10-09
Swap order of skip links so main content comes first

thatbudakguy:
Some accessibility checkers, like SiteImprove, regard it as a
failure that the 'skip to main' link is not the first focusable
element on the page.

This swaps the order of the 'skip to main' and 'skip to search'
links so that 'skip to main' comes first, which is in line with
the WCAG 2.2 guidelines
(https://www.w3.org/WAI/WCAG22/Techniques/general/G1.html).

Related: https://github.com/sul-dlss/earthworks/issues/1431


jcoyne:
While this may satisfy automated accessibility checkers, I suspect the first interaction people will use in a Blacklight site is the "Search".  So I wonder if this would cause a regression in usability for keyboard users.

PR #3364, opened 2024-10-15
Use a valid command to test for the existence of docker compose

jcoyne:
docker compose -v was not a valid command, it produces the docker compose help text


sandbergja (approved):
I was wondering about that in the output.  Thanks, @jcoyne !

PR #3365, opened 2024-10-15
Precompile assets for propshaft

jcoyne:
Propshaft's server does not run when `config.eager_load = true`, which is the default behavior when running in CI.  See https://github.com/rails/propshaft/issues/211


sandbergja (approved):
Thanks, @jcoyne !

PR #3369, opened 2024-10-17
Backport #3367 to 8.x (Add a way to filter facets in the facets "more" modal)

sandbergja:
Backport the "more" facet modal filter feature to the 8.x series (see #3367 for the 9.x version)

### DIFF
--- a/app/components/blacklight/search/facet_suggest_input.html.erb
+++ b/app/components/blacklight/search/facet_suggest_input.html.erb
@@ -1,9 +1,9 @@
-<label for="facet-suggest-<%= facet.key %>">
+<label for="facet_suggest_<%= facet.key %>">
   <%= I18n.t('blacklight.search.facets.suggest.label', field_label: presenter&.label) %>
 </label>
-<input class="facet-suggest form-control"
-       id="facet-suggest-<%= facet.key %>"
-       data-facet-field="<%= facet.key %>"
-       name="facet_suggest_<%= facet.key %>"
-       placeholder="<%= I18n.t('blacklight.search.form.search.placeholder') %>">
-</input>
+<%= text_field_tag "facet_suggest_#{facet.key}",
+  nil,
+  class: "facet-suggest form-control",
+  data: {facet_field: facet.key},
+  placeholder: I18n.t('blacklight.search.form.search.placeholder')
+%>
--- a/app/components/blacklight/search/facet_suggest_input.html.erb
+++ b/app/components/blacklight/search/facet_suggest_input.html.erb
@@ -1,0 +1,9 @@
+<label for="facet-suggest-<%= facet.key %>">
+  <%= I18n.t('blacklight.search.facets.suggest.label', field_label: presenter&.label) %>
+</label>
+<input class="facet-suggest form-control"
+       id="facet-suggest-<%= facet.key %>"
+       data-facet-field="<%= facet.key %>"
+       name="facet_suggest_<%= facet.key %>"
+       placeholder="<%= I18n.t('blacklight.search.form.search.placeholder') %>">
+</input>
--- a/app/controllers/concerns/blacklight/catalog.rb
+++ b/app/controllers/concerns/blacklight/catalog.rb
@@ -83,7 +83,12 @@ module Blacklight::Catalog
     @facet = blacklight_config.facet_fields[params[:id]]
     raise ActionController::RoutingError, 'Not Found' unless @facet
 
-    @response = search_service.facet_field_response(@facet.key)
+    query_fragment = params[:query_fragment] || ''
+    @response = if query_fragment.present?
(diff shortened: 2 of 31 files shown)


app/views/catalog/facet.html.erb
```diff
@@ -6,6 +6,7 @@
   <% end %>
 
   <% component.with_title { facet_field_label(@facet.key) } %>
+  <%= render Blacklight::Search::FacetSuggestInput.new(facet: @facet, presenter: @presenter) %>
```
cbeer, line 9: Should we make this opt-in with facet configuration in order to keep this backwards compatible?

app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -83,7 +83,12 @@ def facet
     @facet = blacklight_config.facet_fields[params[:id]]
     raise ActionController::RoutingError, 'Not Found' unless @facet
 
-    @response = search_service.facet_field_response(@facet.key)
+    query_fragment = params[:query_fragment] || ''
+    @response = if query_fragment.present?
+                  search_service.facet_suggest_response(@facet.key, params[:query_fragment])
+                else
+                  @response = search_service.facet_field_response(@facet.key)
```
cbeer, line 90:
```suggestion
                  search_service.facet_field_response(@facet.key)
```

app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -83,7 +83,12 @@ def facet
     @facet = blacklight_config.facet_fields[params[:id]]
     raise ActionController::RoutingError, 'Not Found' unless @facet
 
-    @response = search_service.facet_field_response(@facet.key)
+    query_fragment = params[:query_fragment] || ''
+    @response = if query_fragment.present?
```
cbeer, line 87: Do we need `query_fragment`?

```suggestion
    @response = if params[:query_fragment].present?
```

app/components/blacklight/search/facet_suggest_input.html.erb
```diff
@@ -0,0 +1,9 @@
+<label for="facet-suggest-<%= facet.key %>">
+  <%= I18n.t('blacklight.search.facets.suggest.label', field_label: presenter&.label) %>
+</label>
+<input class="facet-suggest form-control"
```
cbeer, line 4: Should we use e.g. `text_field_tag` to make sure the attribute values get appropriately escaped?

app/controllers/concerns/blacklight/catalog.rb
sandbergja: 🤦🏻 thanks for catching that!

app/javascript/blacklight/debounce.js
```diff
@@ -0,0 +1,15 @@
+// Usage:
+// ```
+// const basicFunction = (entry) => console.log(entry)
+// const debounced = debounce(basicFunction("I should only be called once"));
+//
+// debounced // does NOT print to the screen becase it is invoked again less than 200 milliseconds later
```
maxkadel, line 6:
```suggestion
// debounced // does NOT print to the screen because it is invoked again less than 200 milliseconds later
```

maxkadel, line 6: 🎉

app/views/catalog/facet.html.erb
sandbergja: Added as a configuration option, great idea, thanks!

config/locales/blacklight.ar.yml
```diff
@@ -111,6 +111,9 @@ ar:
         sort:
           count: ترتيب رقمي
           index: ترتيب أبجدي
+        suggest:
```
maxkadel, line 114: Is it expected that these be in the target language? Or is it ok to start with English and then have a future edit with the real translation?

maxkadel (commented):
This looks awesome! Thank you @sandbergja! 

My only hesitation is about the translations and what our expectations are around including language translations, or just making sure the "no translation" error doesn't render in any languages.

config/locales/blacklight.ar.yml
sandbergja: Thanks for pointing this out.  It looks like I used the [`add_missing` task from i18n-tasks](https://github.com/glebm/i18n-tasks?tab=readme-ov-file#add-missing-keys), when I should have used [`translate_missing`](https://github.com/glebm/i18n-tasks?tab=readme-ov-file#translate-missing-keys), which grabs a preliminary translation from Google Translate.

maxkadel (approved):
Thanks @sandbergja! Looks good! Happy to know about the "translate missing" task as well

maxkadel (approved):

seanaery:
Thanks @sandbergja 💯  -- this is such a useful feature and I'm glad to see it'll be available in 8.x!

PR #3381, opened 2024-10-22
Show the item page information even if there's only one result

### LINKED ISSUE #3314: Enable display of pagination component in the case where we have only one result

cbeer:
fixes #3314

This is an alternative to https://github.com/projectblacklight/blacklight/pull/3315 trying  to address jcoyne's comment: https://github.com/projectblacklight/blacklight/pull/3315#discussion_r1761250210


jcoyne (approved):
Thank you @cbeer!

PR #3386, opened 2024-10-23
Revert + revise "Deprecate render_document_sidebar_partial and catalog/show_sidebar partial"

cbeer:
This reverts commit a1f8edf8880542ae27264586163b19d1517d91a9.

This commit breaks backwards compatibility for applications that have overridden render_document_sidebar_partial (or show_sidebar), like applications like the geoblacklight example in the commit comment:

https://github.com/geoblacklight/geoblacklight/blob/7d3c31c7af3362879b97e2c1351a2496c728c59c/app/helpers/blacklight_helper.rb#L7

(Although geoblacklight 5 uses a custom sidebar component, requiring  some work in the downstream gem/app seems like a big problem to me, so maybe this isn't ripe for deprecation in 8.x or needs a more sophisticated deprecation path.)


jcoyne (approved):
Thank you @cbeer

PR #3398, opened 2024-10-24
Arrange HTML so that the search bar is part of the main element

### LINKED ISSUE #3252: Restructure semantic HTML
Currently the HeaderComponent contains the search bar.  However, I think it would be better semantics to move the search bar into the `<main>` https://github.com/projectblacklight/blacklight/blob/main/app/views/layouts/blacklight/base.html.erb#L38-L40

Then we could have a structure like this:
```html
<body>
  <header>
  </header>
  <main>
    <search>
       <form><!-- search bar goes here --></form>
       <section><!-- selected options --></section>
       <section><!--facets--></section>
       <section><!--search results--></section>
    </search>
  </main>
 </body>
 ```

jcoyne:
Fixes #3252

### DIFF
--- a/app/components/blacklight/header_component.html.erb
+++ b/app/components/blacklight/header_component.html.erb
@@ -1,2 +1,1 @@
 <%= top_bar %>
-<%= search_bar %>
--- a/app/components/blacklight/header_component.rb
+++ b/app/components/blacklight/header_component.rb
@@ -6,10 +6,6 @@ module Blacklight
       component.new(blacklight_config: blacklight_config)
     }
 
-    renders_one :search_bar, lambda { |component: Blacklight::SearchNavbarComponent|
-      component.new(blacklight_config: blacklight_config)
-    }
-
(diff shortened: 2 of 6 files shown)


cbeer:
I'm not sure I understand the motivation for this change, but this seems like it could be incredibly disruptive to downstream applications are require a lot more overriding to recreate fairly common layout patterns:

Screenshot 2024-10-24 at 10 41 09

Screenshot 2024-10-24 at 10 41 24


Screenshot 2024-10-24 at 10 42 11

Screenshot 2024-10-24 at 10 42 56

jcoyne:
I agree that this may be a challenge for adopters to take on.  I believe this provides better semantics for accessibility. I believe this was raised by Darcy when looking at Geoblacklight. 

While implementing this it felt like our existing layouts (with content_for) made this a challenge to implement easily.

dbranchini:
Let's close as won't do.

PR #3422, opened 2024-10-28
update link_to_search to work with only one/no search_field

### LINKED ISSUE #3302: "Skip to search" link doesn't work if you only have one search_field
If in the configuration you only call `add_search_field` once, then the `search_field` form component is hidden.  However the "Skip to search" link still points at that hidden element.

See https://github.com/geoblacklight/geoblacklight/issues/1580

dnoneill:
closes #3302

### DIFF
--- a/app/components/blacklight/skip_link_component.rb
+++ b/app/components/blacklight/skip_link_component.rb
@@ -3,11 +3,17 @@
 module Blacklight
   class SkipLinkComponent < Blacklight::Component
     def link_to_search
-      link_to t('blacklight.skip_links.search_field'), '#search_field', class: link_classes
+      link_to t('blacklight.skip_links.search_field'), search_id, class: link_classes
     end
 
     def link_to_main
       link_to t('blacklight.skip_links.main_content'), '#main-container', class: link_classes
+    end
+
+    def search_id
+      return '#search_field' if helpers.blacklight_config.search_fields.values.many? { |field_def| helpers.should_render_field?(field_def) }
+
+      '#q'
     end
 
     def link_classes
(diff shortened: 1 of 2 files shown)


app/components/blacklight/skip_link_component.rb
```diff
@@ -3,7 +3,8 @@
 module Blacklight
   class SkipLinkComponent < Blacklight::Component
     def link_to_search
-      link_to t('blacklight.skip_links.search_field'), '#search_field', class: link_classes
+      search_id = helpers.blacklight_config.search_fields.keys.length < 2 ? '#q' : '#search_field'
```
cbeer, line 6:
```suggestion
      search_id = helpers.blacklight_config.search_fields.many? ? '#search_field' : '#q'
```

app/components/blacklight/skip_link_component.rb
```diff
@@ -3,7 +3,8 @@
 module Blacklight
   class SkipLinkComponent < Blacklight::Component
     def link_to_search
-      link_to t('blacklight.skip_links.search_field'), '#search_field', class: link_classes
+      search_id = helpers.blacklight_config.search_fields.many? ? '#search_field' : '#q'
```
cbeer, line 6: I'm not sure it's enough to check whether multiple fields are configured; don't we need to reproduce the rest of the logic that's used in the SearchBarComponent?

https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/search_bar_component.html.erb#L13

I think it'd be something like:
```suggestion
      search_id = helpers.blacklight_config.search_fields.values.select { |field_def| helpers.should_render_field?(field_def) }.many? ? '#search_field' : '#q'
```

(but maybe it'd be cleaner to extract `search_id` to its own method or something?

PR #3423, opened 2024-10-30
Use a labeled <aside> for show tools sidebar

thatbudakguy:
The show page sidebar is currently displayed
inside a `<section>`; this changes it to use the `<aside>` element,
which is arguably more appropriate for a sidebar.

The search sidebar had an aria-label already set, but the show
sidebar did not, which also meant that it had no accessible name
despite having a landmark role.

Thus this also adds a configurable aria-label for the show page sidebar
similar to the one for the search sidebar.

### DIFF
--- a/config/locales/blacklight.en.yml
+++ b/config/locales/blacklight.en.yml
@@ -245,3 +245,7 @@ en:
     pagination_compact:
       next: Next &raquo;
       previous: "&laquo; Previous"
+  layouts:
+    catalog_result:
+      aria:
+        sidebar: Tools
--- a/app/views/layouts/catalog_result.html.erb
+++ b/app/views/layouts/catalog_result.html.erb
@@ -3,9 +3,9 @@
     <%= yield %>
   </section>
 
-  <section class="<%= show_sidebar_classes %>">
+  <aside class="<%= show_sidebar_classes %>" aria-label="<%= t('.aria.sidebar') %>">
     <%= content_for(:sidebar) %>
-  </section>
+  </aside>
 <% end %>
 
 <%= render template: "layouts/blacklight/base" %>
--- a/config/locales/blacklight.ar.yml
+++ b/config/locales/blacklight.ar.yml
@@ -222,6 +222,10 @@ ar:
       aria:
         container_label: التنقل الرئيسي
     welcome: مرحبًا!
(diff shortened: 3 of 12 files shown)


thatbudakguy:
This PR results in the following accessible names for the sidebars in VoiceOver (the sidebar landmark is highlighted):

Search:

Screenshot 2024-10-30 at 13 39 45

Show:

Screenshot 2024-10-30 at 13 39 26

jcoyne:
I don't think it actually fits the aside definition:
> The complementary [landmark role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#3._landmark_roles) is used to designate a supporting section that relates to the main content, yet can stand alone when separated.

I don't think the facets can stand alone, they must be displayed in context of the search results.

thatbudakguy:
> I don't think the facets can stand alone, they must be displayed in context of the search results.

Is this also true of the show tools/show page sidebar, do you think? Or just search facets?

jcoyne:
@thatbudakguy I think the tools and show could arguably be an aside.

thatbudakguy:
Updated to only change the show page sidebar.

config/locales/blacklight.en.yml
```diff
@@ -245,3 +245,7 @@ en:
     pagination_compact:
       next: Next &raquo;
       previous: "&laquo; Previous"
+  layouts:
+    catalog_result:
+      aria:
+        sidebar: Sidebar
```
jcoyne, line 251: Is "Sidebar" the best we can do here? How about "Tools"?  Additionally, once a term is chosen, can you run the i18n tools to fix the translations for the other languages rather than keeping them English?

jcoyne (changes requested):

config/locales/blacklight.en.yml
thatbudakguy: By "fix" do you mean using the google translate backend to pre-fill translations?

jcoyne: Yes.

thatbudakguy: Updated to use "Toolbar" and its translated versions.

jcoyne (approved):

PR #3442, opened 2024-11-11
ClausePresenter: don't attempt to call #display_label on a non-existant field configuration

### LINKED ISSUE #3441: ClausePresenter throws a NoMethodError when user tries to search a non-existant field
### Steps to recreate

1. Start blacklight locally
2. Go to http://localhost:3000/?clause[0][field]=not_a_real_field&clause[0][query]=cats
3. Note that you get `NoMethodError (undefined method 'display_label' for nil:NilClass)`

Every other part of the stack is able to handle cases in which the field does not exist, I think that this class should match.

Alternatively, if we do want to throw an error in cases that the user tries to search a non-existant field, I think it should be a more intuitive error in a different part of the stack.

sandbergja:
closes #3441

### DIFF
--- a/app/presenters/blacklight/clause_presenter.rb
+++ b/app/presenters/blacklight/clause_presenter.rb
@@ -13,7 +13,7 @@ module Blacklight
     end
 
     def field_label
-      field_config.display_label('search')
+      field_config&.display_label('search')
     end
 
     ##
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
This handles the failure silently, but shouldn't this result in a 401 Bad Request?

PR #3450, opened 2024-11-21
Set search bar border rounded class for number of configured search fields

### LINKED ISSUE #3448: Search box should have rounded left/start border when only one search field is configured
Current behavior:
Screenshot 2024-11-21 at 8 35 24 AM

Expected behavior:
Screenshot 2024-11-21 at 8 56 22 AM

There are two problems. When only one search field is configured the `SearchBarComponent` sets the `rounded-left` class on the `input` element. It also needs to set the `rounded-start` class for Bootstrap 5. But this only fixes the problem when autocomplete is turned off. If autocomplete is turned on the border is set on the `autocomplete` element so the `rounded-left` and `rounded-start` classes need to be set there as well for this to work when autocomplete is on.

Border set on autocomplete element:
Screenshot 2024-11-21 at 8 54 27 AM

corylown:
Fixes #3448 

With one configured search field:

Before:
Screenshot 2024-11-21 at 8 35 24 AM

After:
Screenshot 2024-11-21 at 9 08 58 AM

### DIFF
--- a/app/components/blacklight/search_bar_component.rb
+++ b/app/components/blacklight/search_bar_component.rb
@@ -60,6 +60,12 @@ module Blacklight
       blacklight_config.advanced_search.enabled
     end
 
+    def rounded_border_class
+      return 'rounded-0' if search_fields.length > 1
+
+      'rounded-start'
+    end
+
     private
 
     def blacklight_config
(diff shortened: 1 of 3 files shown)


app/components/blacklight/search_bar_component.rb
```diff
@@ -60,6 +60,12 @@ def advanced_search_enabled?
       blacklight_config.advanced_search.enabled
     end
 
+    def rounded_border_class
+      return 'rounded-0' if search_fields.length > 1
+
+      'rounded-left rounded-start'
```
jcoyne, line 66: Isn't rounded-left a bootstrap 4 class?  The main branch no longer has support for Bootstrap 4.

jcoyne (changes requested):

jcoyne (approved):

PR #3451, opened 2024-11-21
Remove axe exceptions for unused typeahead and bootstrap 4

jcoyne:


seanaery (commented):
I definitely approve pruning the unused code here -- good catch. Though I do kinda like keeping `be_accessible` as a clearer alias for `be_axe_clean`. There may be good reasons not to do that though -- perhaps e.g., just because it passes Axe automated tests doesn't guarantee that it is accessible.

seanaery (approved):

PR #3466, opened 2024-12-10
[8.x] Import and export modalForm.js so it is included in the rollup build

sandbergja:
Before this commit, modalForm.js was not included in the rollup build, which meant that applications with forms in the modal would not work properly, even if they had the required `data-blacklight-modal=trigger`

To confirm the fix, you can run:
```
npm install && npm run prepare && grep triggerFormSelector app/assets/javascripts/blacklight/blacklight.js
```

If you get a match, that means that the file is included in the javascript build and applications can use it.

This affects both the release-8.x branch and main.  It is blocking our upgrade to Blacklight 8 (see https://github.com/pulibrary/orangelight/issues/4567).  We will also open up a pull request for main.

### DIFF
--- a/app/javascript/blacklight/index.js
+++ b/app/javascript/blacklight/index.js
@@ -1,6 +1,7 @@
 import BookmarkToggle from 'blacklight/bookmark_toggle'
 import ButtonFocus from 'blacklight/button_focus'
 import Modal from 'blacklight/modal'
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thank you @sandbergja and @christinach 🍄 🐬

PR #3470, opened 2024-12-13
Remove hover style from non-controllable element

### LINKED ISSUE #3469: Hover style applied to constraint values make it appear like a link/button, but there is no action.
It is better to remove the hover style so it doesn't confuse the user as to why the control doesn't work when clicked.

Normal style
Screenshot 2024-12-13 at 9 26 29 AM

Hover style
Screenshot 2024-12-13 at 9 26 22 AM

jcoyne:
The hover style indicates to the user that the currently hovered element is a clickable control. However in this case it is not a control, so removing the hover style improves UX.

Fixes #3469

### DIFF
--- a/app/assets/stylesheets/blacklight/_constraints.scss
+++ b/app/assets/stylesheets/blacklight/_constraints.scss
@@ -27,10 +27,6 @@
       margin-left: 0.25rem;
       transition: background-color 0.15s ease-in-out;
     }
-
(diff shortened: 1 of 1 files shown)


seanaery (approved):
This is great -- thanks for tackling it! I noticed this happening a while back and also found it to be an undesirable effect, but just never got around to filing an issue.

PR #3485, opened 2025-01-03
[#3484] Add missing argument to Hash default proc

### LINKED ISSUE #3484: Accessing Null Facet Fields through @response.aggregations can be very slow
When you try to access a facet field that is not in the solr response, e.g. `@response.aggregations['my-new-field']`, Blacklight returns a null `Blacklight::Solr::Response::FacetField` object.

@maxkadel , @kevinreiss, @christinach , and I were investigating a performance issue in our advanced search form, and found that every time we try to do `@response.aggregations['my-new-field']` on a field not in the solr response, the code spends a lot of time in`Hash#inspect` mysteriously.  It turns out that there is a subtle bug in the `default_aggregations` method that causes another method called `facet_field_aggregation_options` to receive a huge hash of all facets and their values, rather than the field name string that it expects.  It tries to interpolate the hash in a few places the field_name should go via `Hash#inspect`, and it can get pretty expensive if you have a lot of facet values to deal with.

PR coming up soon!

sandbergja:
The proc to fill in a default value for a Hash accepts two arguments: the hash itself and then the key.  Prior to this commit, our proc was only using the Hash, but was treating it as if it was a key.  This had the potential to cause some serious performance problems when solr returned many facets or many facet values, when a huge hash was treated as a string and interpolated into other strings via `Hash#inspect`.

Adding this missing argument ensures that we send the key/field_name, rather than the entire hash.  For the provided test case on my laptop, this commit speed up null facet object generation up from 45 iterations/second to 18,700 iterations/second.  It also doubles the performance of our advanced search form.

This commit also adds the rspec-benchmark gem as a dev dependency -- I'm definitely open to figuring out a different way to test this if the additional dependency is not desired.

Closes #3484


spec/spec_helper.rb
```diff
@@ -24,6 +24,12 @@
 require 'rsolr'
 require 'blacklight'
 
+require 'rspec-benchmark'
+
+RSpec.configure do |config|
+  config.include RSpec::Benchmark::Matchers
```
maxkadel, line 30: Would it be faster to load individual tests if we load this in the individual test files, rather than for the whole suite? Example from [gem documentation](https://github.com/piotrmurach/rspec-benchmark):

```ruby
RSpec.describe "Performance testing" do
  include RSpec::Benchmark::Matchers
end
```

maxkadel (approved):
Thanks @sandbergja! 

I have one comment, but it does not need to block this from going in.

Having seen loading single tests get slow in other applications, I'm starting to think more conservatively about loading dependencies for the whole suite.

It's really cool to learn about rspec-benchmark! I'm also excited about the advanced search form loading faster!

maxkadel (approved):
Thanks @sandbergja!

PR #3488, opened 2025-01-09
fix paginator next_page_tag

dnoneill:
next_page_position is not a variable. this made it so that next_page button never rendered. Replaced it with the correct variable next_button_position


jcoyne:
It looks like this was introduced in #3349

PR #3503, opened 2025-01-30
[backport 8.x] Strip markup from html_safe selected facet values in the html title.

seanaery:
Backports #3464 to 8.x
- E.g., span/data-attributes rendered for selected range facet values by blacklight_range_limit


sandbergja (approved):
Thanks, @seanaery !

Issue #3507, opened 2025-02-05
Facet suggest: filtering facets, choosing a value, then attempting to filter again does not work


sandbergja:
Steps to recreate locally on the release-8.x branch:

1. Follow [these steps to set up blacklight locally](https://github.com/projectblacklight/blacklight/wiki/Testing-and-Developing-Blacklight#step-by-step-with-more-control) 
2. Edit the subject facet field configuration to the following in .internal_test_app/app/controllers/catalog_controller.rb:
    ```
    config.add_facet_field 'subject_ssim', label: 'Topic', limit: 3, index_range: 'A'..'Z', suggest: true
    ```
3. In .internal_test_app, run `bundle exec rails s`
4. In your browser, go to http://localhost:3000
5. In the topic facet, click the "More >>" link
6. Type "dr" and select Japanese Drama
8. Note that the URL contains `only_values=true` and `query_fragment=dr`
9. In the topic facet, click the "More >>" link again
10. Note that the modal is not usable, with no facet values, not even an "X" to dismiss the modal! 😞

Here is what the modal looks like for me:

![A very small modal says Japanese drama, has an x, and then the number 2](https://github.com/user-attachments/assets/981f85b3-7a4f-43ff-9878-72cdb5433b5c)

Removing `only_values=true` and `query_fragment=dr` from the URL fixes it for me.  I believe that the links in the facet modal should not include those parameters.

maxkadel:
Closed by #3509

PR #3509, opened 2025-02-06
Strip out new facet-only request keys

maxkadel:
By stripping out the new facet-only request keys, used for doing faster Solr queries for facet counts during facet suggestions, we can perform a second facet-suggest without issue.

image

Connected to #3507


sandbergja (approved):
Thanks, @maxkadel !

PR #3510, opened 2025-02-06
[Backport 8.x] Backport #3507 - Strip out new facet-only request keys

maxkadel:
This is a backport of #3509

Before this backport, when someone used the facet suggest form, the parameters intended only to help the Javascript show relevant facets were included in the index url, so that when someone tried to add another facet using the same feature, they could only see their original facet.

See issue #3507


sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #3519, opened 2025-03-06
[#3518] Remove incomplete implementation of menu aria role from Blacklight::System::DropdownComponent

### LINKED ISSUE #3518: Remove menu and menuitem roles from Blacklight::System::DropdownComponent
### Issue

The `Blacklight::System::DropdownComponent` implements the aria menu roles partially, but not fully.  In particular, [we have not implemented the necessary keyboard interactions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#keyboard_interactions).  This makes navigation with a screen reader like NVDA much less convenient.  Rather than providing a partially implemented aria menu role, we should remove the role and not interfere with the builtin functionality.

### More details

We previously addressed some of the issues with the partial implementation in #2222, but have not yet fully implemented the aria menu role.

The menu role causes NVDA to go into focus mode, which turns off many useful keyboard shortcuts so as not to conflict with the keyboard interactions that an element with the menu role should provide.  However, since we haven't supplied the keyboard interactions, users get neither the keyboard interactions that would come with NVDA's browse mode, nor the keyboard interactions promised when we declare an element to have the menu role.

For some concrete examples: without the menu and menuitem roles, an NVDA user can open the results per page dropdown, then press "k" and "Shift+K" to navigate between the links in the dropdown, or "v" and "Shift+V" to navigate only between the links they've visited before.  However, with the role=menu, these options are not available, and the user has to just stick with Tab and Shift+Tab.

sandbergja:
Closes #3518

### DIFF
--- a/app/components/blacklight/system/dropdown_component.html.erb
+++ b/app/components/blacklight/system/dropdown_component.html.erb
@@ -1,7 +1,7 @@
 <%= content_tag :div, id: @id, class: @classes.join(' ') do %>
   <%= button %>
 
-  <div class="dropdown-menu" role="menu">
+  <div class="dropdown-menu">
     <% options.each do |option| %>
       <%= option %>
     <% end %>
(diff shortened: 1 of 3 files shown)


jcoyne:
Wow, that was added 11 years ago: https://github.com/projectblacklight/blacklight/commit/8718de5d17abe2e1432bcfe2b76275c629c82331

Issue #3522, opened 2025-03-14
Modal facet filter input needs separation from elements below it


seanaery:
This is most apparent with a facet that has `index_range: true` and displays A-Z browse. 
<img width="795" alt="Image" src="https://github.com/user-attachments/assets/cbbb26ca-f3eb-49f5-8ea7-75076200bfec" />

<img width="797" alt="Image" src="https://github.com/user-attachments/assets/460ad324-7fe5-4efc-946e-2e8b4677368a" />

PR #3525, opened 2025-03-14
Add some bottom margin to the facet-suggest filter in the modal. Clos…

seanaery:
…es #3522.
Screenshot 2025-03-14 at 3 41 29 PM


sandbergja (approved):
Thanks, @seanaery!

PR #3526, opened 2025-03-17
Use the repository configured in blacklight.yml

jcoyne:
This allows us to use a custom repository without having to configure it in two places


seanaery:
OK, I see, and there are already specs for `repository_class` in `spec/lib/blacklight_spec.rb` that verify that it works with with different adapter keys like `elastic_search`. Good DRY here.

seanaery (approved):

Issue #3524, opened 2025-03-14
Modal facet nav has odd pagination & sort above the header


seanaery:
The `facet.html.erb` template that renders the modal component ([code](https://github.com/projectblacklight/blacklight/blob/main/app/views/catalog/facet.html.erb)) is currently rendering the `facet_pagination.html.erb` partial twice: once above the modal title and once below the facet values. This does not appear to be intentional.
<img width="881" alt="Image" src="https://github.com/user-attachments/assets/0c02d3dc-d635-400b-ae21-6815e99ff286" />

### Notes
* The top nav seems to have been present in the markup for a long time but had previously been hidden via CSS ([here](https://github.com/projectblacklight/blacklight/blob/main/app/assets/stylesheets/blacklight/_modal.scss#L5-L8)).
* The inline comment `/* Doesn't display in a modal, but would display on a new page (e.g. without Javascript) */` indicates it was never intended to display in the modal but at least at some point was intended to appear on a facet page like https://demo.projectblacklight.org/catalog/facet/subject_ssim 
* The top nav started appearing after #3445 -- the addition of `d-flex` gives the element `display: flex !important` which overrides its `display: none`.
* Problem was originally noted in the demo video in #3453

jcoyne:
Yeah, it seems like we need to make that `d-flex` added in #3445 conditional.

seanaery:
@jcoyne Right, that could work. Another alternative would be to actually intentionally render the top pagination regardless of whether you're in the modal or on the facet browse page (with some minor restyling). Here's a branch and a quick demo of an idea for that:
https://github.com/projectblacklight/blacklight/pull/3530

Couple benefits of having the `facet_pagination` partial visible at the top: 1) it's easier to page-through values without having to scroll down; 2) a user actually sees right away that they can toggle the sort without having to scroll.

PR #3530, opened 2025-03-17
WIP: demo idea for facet pagination within modal or on facet browse p…

seanaery:
…age. Possible solution for #3524.


seanaery:
Here's a 30sec demo of this idea, which would preserve the top pagination for both the modal facet browse and non-modal page, just moving it under the new facet filter element.


https://github.com/user-attachments/assets/77a7e1d1-392c-45c6-9410-955eccf22a0d

seanaery:
Closing in favor of #3548 as that is a more refined solution.

Issue #3532, opened 2025-03-20
Facet browse page prev/next button style is incorrect when disabled


seanaery:
On a facet browse page (e.g., `/catalog/facet/subject_ssim`) the pagination links are not styled correctly when in a disabled state (see border). 

<img width="995" alt="Image" src="https://github.com/user-attachments/assets/6fe927b2-f689-40f3-8034-1e11b63b397c" />

There's some css that removes this border when displayed in the default modal UI, but it'd be cleaner to just use the `btn-link` class as is already done for a clickable prev/next link.

PR #3533, opened 2025-03-20
Fix facet prev/next nav link appearance when the link is disabled. Fi…

seanaery:
…xes #3532

- BS5 btn classes are not meant for span tags (see https://getbootstrap.com/docs/5.3/components/buttons/#button-tags)
- BS5 disabled buttons should not have an href, and should use aria-disabled="true" (see https://getbootstrap.com/docs/5.3/components/buttons/#disabled-state)

### DIFF
--- a/app/assets/builds/blacklight.css
+++ b/app/assets/builds/blacklight.css
@@ -434,9 +434,6 @@ main {
 .modal-content .blacklight-modal-close {
   display: block;
 }
-.modal-content .prev_next_links.btn-group .btn {
-  border: 0;
-}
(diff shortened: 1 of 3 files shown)


seanaery:
Screenshots of the disabled links with this change in action, both in-modal and on facet page (disregard the top-rendered nav, will be addressed in future PR):
Screenshot 2025-03-20 at 12 53 11 PM
Screenshot 2025-03-20 at 12 53 04 PM

tpendragon (approved):

PR #3553, opened 2025-04-03
Ensure facet filter gray box (in modal or facet browse page) doesn't …

### LINKED ISSUE #3552: Facet modal looks odd when filtering/suggest is turned off
Missed in #3548 -- when a facet field has been configured to _not_ support live-filtering modal input, it should not show the gray box that contains that element. 

E.g., if you configure a field like this (with `suggest: false`):
```
config.add_facet_field 'subject_ssim', label: 'Topic', limit: 20, suggest: false
```
...it currently displays like this (see undesired blank gray box):
Image

Note too the effect of `index_range` config being present or absent, since the range nav appears in this box if applicable. I.e., we should only render that box if there's content to display in it.

seanaery:
…appear if it's empty. Fixes #3552.

- this area will be empty for facets configured with suggest: false
- uses a simple :has CSS rule rather than complicate the template rendering

### DIFF
--- a/app/assets/stylesheets/blacklight/_facets.scss
+++ b/app/assets/stylesheets/blacklight/_facets.scss
@@ -205,3 +205,9 @@ $facet-toggle-height: $facet-toggle-width !default;
       $facet-toggle-width auto no-repeat;
   }
 }
+
+/* Facet browse pages & modals
+-------------------------------------------------- */
+.facet-filters:not(:has(*)) {
+  display: none !important;
+}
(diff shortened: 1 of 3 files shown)


seanaery:
With `suggest: false`...
Screenshot 2025-04-03 at 3 18 19 PM

With `suggest: false`, but `index_range: true` and in `A-Z Sort`...
Screenshot 2025-04-03 at 3 18 29 PM

With `suggest: true`...
Screenshot 2025-04-03 at 3 20 05 PM

With `suggest: true` and `index_range: true` and in `A-Z Sort`...
Screenshot 2025-04-03 at 3 20 11 PM
d

jrochkind (approved):

app/assets/stylesheets/blacklight/_facets.scss
```diff
@@ -205,3 +205,9 @@ $facet-toggle-height: $facet-toggle-width !default;
       $facet-toggle-width auto no-repeat;
   }
 }
+
+/* Facet browse pages & modals
+-------------------------------------------------- */
+.facet-filters:not(:has(*)) {
+  display: none !important;
```
jcoyne, line 212: Why does this have to be `!important`?

jcoyne (changes requested):

jcoyne (approved):

PR #3562, opened 2025-04-07
Darken modal backdrops, to match Bootstrap opacity. Fixes #3561.

### LINKED ISSUE #3561: Blacklight modal backdrops are not dark enough
When a user opens a Blacklight modal (e.g., a facet `more` link, or an item show tools link like `Email`), the backdrop behind the modal only darkens a very slight amount, and this difference can be imperceptable to users.

In 2024 usability testing, multiple users didn't notice that a modal had appeared after clicking a facet `more` link (_Source: https://crkn-rcdr.gitbook.io/user-centered-design#facet-pop-ups_): 
> During Task 4, 16.67% of users did not notice the pop-up right away and were not expecting their attention to need to shift to the center of the page after clicking more.

This is a side-effect of #2671, which replaced Bootstrap modals with native HTML5 `<dialog>` elements to more readily work accommodate different major versions of Bootstrap. We have made some subsequent refinements (e.g., #2815 & #2785 ) to get Blacklight modals to appear and behave more like Bootstrap modals.

Though implemented differently, **both** Bootstrap 4 & 5 use a black backdrop with **50% opacity** to darken the page behind a modal.

The native HTML5 `<dialog>` used by Blacklight uses a black backdrop with only **10% opacity** via the `::backdrop` pseudo-element. This is not defined within Blacklight; it's the default browser style, at least in the few browsers I have tested. I.e.: `dialog:-internal-dialog-in-top-layer::backdrop { rgba(0, 0, 0, 0.1) }`.

It seems it'd be valuable for Blacklight to emulate Bootstrap here and use 50% opacity black out of the box, while at the same time make this style easy to override for an implementer regardless of the version of Bootstrap they're using.

seanaery:
- Uses black w/0.5 opacity by default to match BS4 & BS5 modals (native dialog uses 0.1)
- Adds --bl-modal-backdrop-bg CSS variable for downstream customization

### DIFF
--- a/app/assets/builds/blacklight.css
+++ b/app/assets/builds/blacklight.css
@@ -18,6 +18,8 @@
   --bl-constraint-remove-hover-bg: #bb2d3b;
   --bl-constraint-remove-hover-border-color: #bb2d3b;
   --bl-field-name-color: var(--bs-secondary-color);
+  /* emulate Bootstrap backdrop bg & opacity */
+  --bl-modal-backdrop-bg: rgba(0, 0, 0, 0.5);
 }
 
 .page-link {
@@ -30,6 +32,9 @@ dialog.modal[open] {
   border: none;
   max-height: unset;
   max-width: unset;
+}
+dialog.modal[open]::backdrop {
+  background-color: var(--bl-modal-backdrop-bg);
 }
 
 /* Generic layout stuff */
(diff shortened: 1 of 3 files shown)


seanaery:
Before:
Screenshot 2025-04-07 at 2 04 55 PM

After:
Screenshot 2025-04-07 at 2 04 15 PM

jcoyne (approved):

PR #3569, opened 2025-04-16
[7.x] Fix testing in CI

seanaery:
- backports same Solr 9.8 compatibility fix that was previously backported to 8.x
- bumps versions of ruby & rails tested via CI to supported versions, drops EOL versions


maxkadel:
Superseded by https://github.com/projectblacklight/blacklight/pull/3580

PR #3570, opened 2025-04-17
Separate into reusable workflows so we can schedule CI for release branches

maxkadel:
See this [workflow run](https://github.com/projectblacklight/blacklight/actions/runs/14522877941) for an example of what will happen if one of the release-8.x CI runs fails.

In order to re-use workflows, separated existing configuration into reusable workflows. 

TODO: Allow more of the matrix to be configurable, especially the "include" stanza. Once the release-7.x branch is green, this will be needed to add that branch to this configuration.

It could also be worth it to put these actions into their own repo in the projectblacklight namespace, and try to configure some of the plugins to use them as well, which could help with compatibility snarls, but that's a bigger project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,16 @@
+on:
+  workflow_call:
+    inputs:
+      ref:
+        required: false
+        type: string
+        default: ''
+jobs:
+  build:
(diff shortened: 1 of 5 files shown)


jcoyne (approved):
Great job @maxkadel

PR #3578, opened 2025-04-22
Split ci workflows

maxkadel:
This PR will allow us to point to the CI configuration on the release-8.x branch from the main branch and run CI on the release-8.x branch weekly from there.

Read the matrix from its own file - this will make extending to the release-7.x branch easier.


sandbergja (approved):
Thanks, @maxkadel !  I like the solution of having a branch-specific matrx.json that enumerates which versions of Ruby, Rails, etc. it supports.

PR #3593, opened 2025-04-30
Compare document with bookmark using string compairson of class

jcoyne:
Two classes that are of different versions, which can be triggered by changing the source code when using hot reloading, are not equal.  By casting them to_s, then we make them equal even at different versions.


seanaery (commented):
Code change looks good and thanks for documenting with a note. Minor issue but there's one typo in the note ("compairson").

body-clock (approved):

PR #3597, opened 2025-05-09
Extract FacetSearchBuilder

jcoyne:

### DIFF
--- a/app/controllers/concerns/blacklight/catalog.rb
+++ b/app/controllers/concerns/blacklight/catalog.rb
@@ -9,6 +9,7 @@ module Blacklight::Catalog
   include Blacklight::Configurable
   include Blacklight::SearchContext
   include Blacklight::Searchable
+  include Blacklight::Facetable
 
   # The following code is executed when someone includes blacklight::catalog in their
   # own controller.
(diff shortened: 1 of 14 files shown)


jrochkind:
I am having the problem @cbeer mentioned above:

> Doesn't the FacetSearchBuilder need to stay in sync with the regular SearchBuilder in order to get the same facet result set?

It's causing me a lot of trouble trying to upgrade to BL9. 

@jcoyne , can you supply any context about your motivation in doing this PR, to understand the goals?  Perhaps there is a change that can meet everyone's needs, if we underestand them.

barmintor:
> Maybe it was supposed to simplify code? But this PR has a net gain of ~900 lines of code, so maybe not.

This is not a helpful comment: There were 1276 added lines in this PR, but 929 of those lines were in a single new spec file (spec/models/blacklight/solr/facet_search_builder_behavior_spec.rb). That leaves the delta on the rest of the app as +347, -378, largely from extracting reused classes/modules.

It did/does seem to me to simplify things for apps that do a lot of customization around facet searching, but in our apps we don't have a class hierarchy of search builders, they are all composed with configured filter chains.

PR #3598, opened 2025-05-09
Add new rubocops and regnerate todo

jcoyne:


.rubocop.yml
```diff
@@ -1,4 +1,4 @@
-require:
```
cgalarza, line 1: Should we pin `rubocop` to 1.72 or greater in the gemspec? I am not sure if this syntax would work with older version of `rubocop`.

cgalarza (approved):

PR #3599, opened 2025-05-09
Extract FilterQueryBuilder to its own class

jcoyne:
SearchBuilderBehavior is way too big and has too many responsibilities.  This removes a chunk of them to its own class responsible for building only filter queries.


lib/blacklight/solr/search_builder_behavior.rb
```diff
@@ -128,25 +128,19 @@ def add_facet_fq_to_solr(solr_parameters)
...
 
       search_state.filters.each do |filter|
-        if filter.config.filter_query_builder
-          filter_query, subqueries = filter.config.filter_query_builder.call(self, filter, solr_parameters)
-
-          Array(filter_query).each do |fq|
-            solr_parameters.append_filter_query(fq)
-          end
-          solr_parameters.merge!(subqueries) if subqueries
+        filter_query_builder_class_or_proc = filter.config.filter_query_builder || DefaultFilterQueryBuilder
```
sandbergja, line 131: Nice, I like that this is configurable

sandbergja (approved):
Thanks, @jcoyne !

PR #3602, opened 2025-05-13
Retain keyboard focus on bookmark checkbox

### LINKED ISSUE #3601: Bookmark checkboxes not retaining keyboard focus
When using a keyboard to bookmark (or remove a bookmark) on an item you use the “tab” key to navigate to the bookmark checkbox and then press the space bar to select/deselect the bookmark checkbox.  Pressing space for a second time causes the screen to scroll by a screen height instead of selecting/deselecting the bookmark checkbox again.

The focus on the checkbox is lost when the bookmark form "autosubmit" takes place.

Steps to reproduce:

Tab to focus on checkbox
Image

Space bar to check checkbox - loses keyboard focus
Image

Focus should be retained on checkbox
Image

paulhagon:
Retains keyboard focus on the bookmark checkbox when selecting/deselecting a bookmark. Fixes #3601

It's my first input to blacklight core so I hope I've followed all the correct procedures.

### DIFF
--- a/app/javascript/blacklight-frontend/checkbox_submit.js
+++ b/app/javascript/blacklight-frontend/checkbox_submit.js
@@ -36,6 +36,7 @@ export default class CheckboxSubmit {
     }).then((json) => {
       this.labelTarget.removeAttribute('disabled')
       this.checkboxTarget.removeAttribute('disabled')
+      this.checkboxTarget.focus()
       this.updateStateFor(!this.checked)
       this.bookmarksCounter().forEach(counter => {
         counter.innerHTML = json.bookmarks.count;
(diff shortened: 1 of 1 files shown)


jcoyne (changes requested):
Would you please add a comment here about why we need this `.focus()` call? I'm not sure it's obvious.

paulhagon:
Sorry for the wiat on a reply - bit of a timezone difference here in Australia.  We had an accessibility audit carried out on our [Catalogue](https://catalogue.nla.gov.au) and this was one of the issues raised.

The default behaviour of a checkbox when using a keyboard is to use the space bar to check/uncheck & the focus is retained on the checkbox when the state changes. For the bookmark checkbox the event listener that auto submits when the checkbox state changes takes the focus away from the checkbox. This creates a usability & accessibility issue if a user wishes to uncheck again, they have to navigate back to the checkbox (using shift+tab, or tab) to uncheck it again.  This keeps the behaviour the same as a default checkbox.

jcoyne:
@paulhagon I think you misunderstand me. I see the problem and am able to duplicate it. I see that your solution works.  Additionally, I think that adding a comment in the code, that explains why we need `focus()` will be useful so that someone in the future doesn't say, "This line doesn't seem to be necessary. Why is it here? I think I can just remove it.".

jcoyne (approved):

PR #3606, opened 2025-05-13
Modernized view component testing

jcoyne:
Since we no longer support view_component 2

This removes the support methods (`#controller` & `#render_inline_to_capybara_node`) we had in `ViewComponentTestHelpers`


sandbergja (approved):
Thanks, @jcoyne !

PR #3609, opened 2025-05-14
Extract FacetFiltersComponent so it's display can be customized

jcoyne:


app/components/blacklight/search/facet_filters_component.rb
```diff
@@ -0,0 +1,27 @@
...
+    def facet
+      @presenter.facet_field
+    end
+
+    attr_reader :classes, :presenter
+
+    delegate :display_facet, to: :presenter
+
+    def render?
+      facet.suggest != false || render_index_navigation?
```
dnoneill, line 20:
```suggestion
      facet.suggest != false || render_index_navigation?
      facet.suggest.present? || render_index_navigation?
```

dnoneill: It seems like FacetSuggestInput has the possibility of facet being nil. Also shouldn't the render? statement be removed from FacetSuggestInput if it is getting moved?

dnoneill: I don't know. I just know what the current code has: https://github.com/projectblacklight/blacklight/blob/4a9f4263a534e573718d99c68cfeb7e8771badaa/app/components/blacklight/search/facet_suggest_input.rb#L16

dnoneill:
Can you add a test for app/components/blacklight/search/facet_filters_component.rb

app/components/blacklight/search/facet_filters_component.rb
jcoyne: I thought the current code was wrong, so I fixed it.

jcoyne:
@dnoneill sure, test added.

app/components/blacklight/search/facet_filters_component.rb
dnoneill: @jcoyne can we still use facet.suggest.present? instead of != false or are we fine with facet.suggest = '' and we want that?

jcoyne: Facet suggest is a boolean field.  We'd never expect it to have strings.

dnoneill: Then shouldn't the line be facet.suggest if it is a boolean. I am a little confused why there is a check for it against false

jcoyne: No, because it could also be nil.

dnoneill: so it should render if facet.suggest == nil?

jcoyne: Right. it has to match this: https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/facets/suggest_component.rb#L17-L18

dnoneill:
I am having trouble getting blacklight up locally to do a double check of the code. I am getting the error `LoadError: cannot load such file -- view_component/rails/tasks/view_component.rake (LoadError)`

PR #3612, opened 2025-05-14
Remove css that is not applied

jcoyne:
The bootstrap padding has higher precidence, so we can remove this and allow consumers to set as they see fit.


mamrey (approved):
Good catch, I see the bootstrap padding precedence mentioned.

PR #3613, opened 2025-05-14
Get rid of unnecessary _facet_pagination template

jcoyne:
call the component directly


jcoyne:
@cgalarza you are right.  I've moved the class name to a config variable, so that you can just set it one time and you don't have to override the whole `catalog/facet` template.

cgalarza (approved):

cgalarza:
@jcoyne Thanks for making that addition!

PR #3614, opened 2025-05-14
Move facet components to Facets namespace

jcoyne:
Adding a namespace for all the Facet commponents to go to


app/components/blacklight/facet_field_no_layout_component.rb
```diff
@@ -1,14 +1,6 @@
...
-    renders_one :label
-    renders_one :body
-
-    def initialize(**); end
-
-    def call
-      body.to_s
-    end
-  end
+  class FacetFieldNoLayoutComponent < Facets::FieldComponent; end
```
cgalarza, line 4: Should this be?

```suggestion
  class FacetFieldNoLayoutComponent < Facets::NoLayoutComponent; end
```

app/components/blacklight/facet_field_checkboxes_component.rb
```diff
@@ -1,24 +1,6 @@
...
-      return [] unless @facet_field.paginator
-
-      return to_enum(:presenters) unless block_given?
-
-      @facet_field.paginator.items.each do |item|
-        yield Blacklight::FacetCheckboxItemPresenter.new(item, @facet_field.facet_field, helpers, @facet_field.key, @facet_field.search_state)
-      end
-    end
-  end
+  class FacetFieldCheckboxesComponent < Facets::FieldComponent; end
```
cgalarza, line 4:
```suggestion
  class FacetFieldCheckboxesComponent < Facets::CheckboxesComponent; end
```

cgalarza (changes requested):
I really appreciate this quality of life improvement! Since this change is targeted for v9, I would also be in favor of removing the deprecation warnings and making this change part of the migration instructions for v9.

jcoyne:
@cgalarza the deprecation warnings are just added here. If we removed them, then nobody would ever see them.

PR #3615, opened 2025-05-14
Remove unused erb file

jcoyne:


sandbergja (approved):
Thanks for catching this @jcoyne !

PR #3627, opened 2025-05-29
Support Rails 8.0 in Blacklight 7.x

mjgiarlo:
Upgrading one of our apps to BL8 has been a timesink, and we don't want it stuck on Rails 7.2, so this is a quick spike to see if the 7.x release branch can work atop Rails 8.0.

### DIFF
--- a/lib/generators/blacklight/controller_generator.rb
+++ b/lib/generators/blacklight/controller_generator.rb
@@ -30,7 +30,7 @@ module Blacklight
       route <<-EOF
   concern :searchable, Blacklight::Routes::Searchable.new
 
-  resource :catalog, only: [:index], as: 'catalog', path: '/catalog', controller: 'catalog' do
+  resource :catalog, only: [], as: 'catalog', path: '/catalog', controller: 'catalog' do
     concerns :searchable
   end
       EOF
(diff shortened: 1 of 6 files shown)


lib/generators/blacklight/controller_generator.rb
```diff
@@ -30,7 +30,7 @@ def inject_blacklight_routes
       route <<-EOF
   concern :searchable, Blacklight::Routes::Searchable.new
 
-  resource :catalog, only: [:index], as: 'catalog', path: '/catalog', controller: 'catalog' do
+  resource :catalog, only: [], as: 'catalog', path: '/catalog', controller: 'catalog' do
```
mjgiarlo, line 33: Found this change in a commit in `main` that fixed CI for Rails 8.0

spec/spec_helper.rb
```diff
@@ -48,8 +48,11 @@
   # When we're testing the API, only run the api tests
   config.filter_run api: true if ENV['BLACKLIGHT_API_TEST']
 
-  # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
-  config.fixture_path = Rails.root.join("spec/fixtures")
+  if Rails.version.to_f >= 7.1
+    config.fixture_paths = [Rails.root.join("spec/fixtures")]
+  else
+    config.fixture_path = Rails.root.join("spec/fixtures")
+  end
```
mjgiarlo, line 55: Found this change in a commit in `main`. Without this change, the call to `#fixture_path=` raises an exception in Rails 8.

PR #3632, opened 2025-05-30
Stop looking in the Solr response to determine the request params

jcoyne:
We already know which parameters we requested.  This supports elasticsearch, which does not echo the request params

### DIFF
--- a/lib/blacklight/solr/response/params.rb
+++ b/lib/blacklight/solr/response/params.rb
@@ -15,17 +15,7 @@ module Blacklight::Solr::Response::Params
     header['params'] || request_params
   end
 
-  def start
-    search_builder&.start || single_valued_param(:start).to_i
-  end
-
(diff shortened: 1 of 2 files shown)


lib/blacklight/solr/response/params.rb
```diff
@@ -15,17 +15,7 @@ def params
...
-  end
-
-  def rows
-    search_builder&.rows || single_valued_param(:rows).to_i
-  end
-
-  def sort
-    search_builder&.sort || single_valued_param(:sort)
-  end
+  delegate :start, :rows, :sort, to: :search_builder
```
barmintor, line 18: Doesn't this require search_builder to exist? I think it can still be nil. Should the initializer for B::S::Response create one if it's missing from the params (ie, if a hash or nil is passed as second arg)?

PR #3636, opened 2025-05-30
Refactor Solr::Response::Group spec to use SearchBuider

jcoyne:
This is more like what is used in production


barmintor (approved):
See also #3637

PR #3637, opened 2025-05-30
Use SearchBuilder in AdvancedSearchFormComponent spec

jcoyne:
This is more like how a Solr::Response is constructed in production


barmintor (approved):
In reviewing it looks like RDoc should be updated for B:S:Response.initialize, but that's a separate issue.

PR #3645, opened 2025-06-18
Avoid transforming the size of the icon

jcoyne:
This is something a downstream user can do if they want it bigger, but we don't want to force someone to override blacklight.  They may also choose to use bi-x-lg instead.

### DIFF
--- a/app/assets/builds/blacklight.css
+++ b/app/assets/builds/blacklight.css
@@ -445,8 +445,3 @@ main {
   padding-top: 1rem;
   margin-top: 1rem;
 }
-
-.remove .bi {
-  height: 1em;
-  width: 1em;
-}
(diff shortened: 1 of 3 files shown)


seanaery (approved):
I don't really mind these `1em` rules here, but am OK with removing them. They do provide the benefit of scaling the `X` icon proportionately if someone has customized the `--bs-btn-font-size` or `--bs-body-font-size` variable. On the other hand, removing these rules doesn't seem to have a visible effect in the browsers I've tested (i.e., they end up rendered as 16x16 regardless of the extra css), and as you indicated, if someone wants to style the icons downstream, it's better to have fewer layers of styles to have to override. I wasn't aware of utility classes or suffixes like `-lg` that'd work to size blacklight icons (other than like `w-100` or `h-100`?)... so if that's possible it could be useful to document, maybe as part of 9.0 release notes/upgrade docs.

PR #3646, opened 2025-06-18
add bookmarks keyboard navigation listener

### LINKED ISSUE #3555: Bookmark elements are not keyboard navigable
In either a search results/index view or an item show page, a user can't currently reach the bookmark element via keyboard to check or uncheck it. Tabbing through the UI, the bookmark checkbox never receives focus.

Image

dnoneill:
closes #3555 

Also fixes the checked state not being updated properly.

This is actually also an issue for checkboxes. Without this JS change the page will reload when the user uses keyboard navigation to bookmark an item. The screenshot below is when I used the checkboxes to bookmark the 4th result.
Screenshot 2025-06-18 at 4 49 55 PM

Screenshot 2025-06-23 at 3 53 11 PM
Screenshot 2025-06-23 at 3 53 07 PM

### DIFF
--- a/app/components/blacklight/document/bookmark_component.html.erb
+++ b/app/components/blacklight/document/bookmark_component.html.erb
@@ -9,8 +9,8 @@
                inprogress: t('blacklight.search.bookmarks.inprogress')
             }) do %>
   <div class="toggle-bookmark">
-    <label class="toggle-bookmark-label" data-checkboxsubmit-target="label">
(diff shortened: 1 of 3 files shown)


dnoneill:
@seanaery if you put the tabindex=0 on the label the focus ring will go over the entire label which I don't necessarily think is a bad thing.
Screenshot 2025-06-23 at 3 00 49 PM
Screenshot 2025-06-23 at 3 00 44 PM

I am not sure it should only be when there is an icon. As I pointed out in the PR when the user "clicks" using the checkbox the page reloads and then the user has to tab from the start.

PR #3647, opened 2025-06-18
fix remove alignment for selected facet

dnoneill:
Alignment seems to be slightly off in safari
Before:
Screenshot 2025-06-18 at 5 33 15 PM

After:
Screenshot 2025-06-18 at 5 32 48 PM

### DIFF
--- a/app/components/blacklight/constraint_layout_component.rb
+++ b/app/components/blacklight/constraint_layout_component.rb
@@ -25,7 +25,7 @@ module Blacklight
     def remove_button
       return unless @remove_path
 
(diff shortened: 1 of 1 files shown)


seanaery:
As far as I can tell, these classes doesn't seem to impact the appearance of the `X` at all in core Blacklight. Even with #3644 & #3645 in place (which slightly alter the size of the rendered svg), toggling the presence of these classes doesn't seem to move the X even a pixel in either direction in the browsers I've tested (Chrome, Safari, Firefox). Is there a specific case where a local customization (e.g., of the X image or fonts/styles in the constraints) would be helped along by giving the wrapper span these flexbox classes?

PR #3661, opened 2025-06-27
Use a larger Bootstrap close icon for constraint removal. Fixes #3660.

### LINKED ISSUE #3660: Default close/remove icon is too small
The `Blacklight::Icons::RemoveComponent.svg` is a 16x16 svg image of an `X` icon. The likely intent for that icon is for it to render about the size of the text it would appear near (unless given additional style properties). Our default font size is 16px.

The problem is, the image path inside the 16x16 box is mostly padding, so the X itself only takes up about 8x8 pixels when rendered naturally. This was previously remedied by some CSS rules to enlarge the image, which were difficult to override in a downstream app and were recently removed via #3644 & #3645 . This may also relate to problems noted in #3647

A good solution would be to create a new version of the `X` svg that -- like the rest of the [blacklight icon components](https://github.com/projectblacklight/blacklight/tree/main/app/components/blacklight/icons) -- uses the available space in its containing box. The default Bootstrap 5 close icon might also be preferred: https://getbootstrap.com/docs/5.3/components/close-button/ 

Image

Image

seanaery:
- the bi-x-lg icon fills the SVG bounding box and can be more easily scaled/aligned as needed in downstream apps

### DIFF
--- a/app/components/blacklight/icons/remove_component.rb
+++ b/app/components/blacklight/icons/remove_component.rb
@@ -7,8 +7,8 @@ module Blacklight
     #   Blacklight::Icons::RemoveComponent.svg = '<svg>your SVG here</svg>'
     class RemoveComponent < Blacklight::Icons::IconComponent
       self.svg = <<~SVG
-        <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-x" viewBox="0 0 16 16">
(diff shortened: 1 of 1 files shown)


seanaery:
### Before
Screenshot 2025-06-26 at 3 23 53 PM


### After
Screenshot 2025-06-27 at 8 30 06 AM

PR #3690, opened 2025-07-16
Facet field label

jcoyne:


body-clock (approved):
Looks good! This cleans things up significantly for some of the work I'm doing with constraints.

PR #3693, opened 2025-07-18
add doc id for bookmark event

dnoneill:
The bookmark page currently doesn't up and remove bookmarks if they are deselected on the page. I think this would be a good update to blacklight and it is something Stanford wants now. I plan to bring this up during a call but for now this is the building block of making that functionality work.

### DIFF
--- a/app/components/blacklight/document/bookmark_component.html.erb
+++ b/app/components/blacklight/document/bookmark_component.html.erb
@@ -6,7 +6,8 @@
              data: {
                present: t('blacklight.search.bookmarks.present'),
                absent: t('blacklight.search.bookmarks.absent'),
-               inprogress: t('blacklight.search.bookmarks.inprogress')
+               inprogress: t('blacklight.search.bookmarks.inprogress'),
+               doc_id: @document.id
             }) do %>
   <div class="toggle-bookmark">
     <label class="toggle-bookmark-label" data-checkboxsubmit-target="label" <% if bookmark_icon %>tabindex="0"<% end %>>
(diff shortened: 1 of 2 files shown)


app/components/blacklight/document/bookmark_component.html.erb
```diff
@@ -6,7 +6,8 @@
              data: {
                present: t('blacklight.search.bookmarks.present'),
                absent: t('blacklight.search.bookmarks.absent'),
-               inprogress: t('blacklight.search.bookmarks.inprogress')
+               inprogress: t('blacklight.search.bookmarks.inprogress'),
+               doc_id: @document.id
```
jcoyne, line 10: Looks good, but let's go with `document_id` as that's what's more commonly used throughout the codebase.

app/javascript/blacklight-frontend/checkbox_submit.js
```diff
@@ -44,7 +44,7 @@ export default class CheckboxSubmit {
         counter.innerHTML = json.bookmarks.count;
       });
 
-      var e = new CustomEvent('bookmark.blacklight', { detail: { checked: this.checked } });
+      var e = new CustomEvent('bookmark.blacklight', { detail: { checked: this.checked, document_id: this.formTarget.dataset.documentId } });
       window.dispatchEvent(e)
```
cbeer, line 48:
```suggestion
      this.formTarget.dispatchEvent(e)
```
Instead of sending the document id, could we dispatch the event from the form and let consumers do whatever they need to do?

PR #3698, opened 2025-07-24
[8.x Backport] Restore scrolling and execute other modal hide behavior on modal dismiss by "escape" key

cjcolvar:
Backport #3694 to 8.x

### DIFF
--- a/app/javascript/blacklight/modal.js
+++ b/app/javascript/blacklight/modal.js
@@ -162,6 +162,14 @@ const Modal = (() => {
       else if (e.target.matches(`${modal.modalSelector}`) || e.target.closest('[data-bl-dismiss="modal"]'))
         modal.hide()
     })
+
+    // Make sure user-agent dismissal of html 'dialog', etc `esc` key, triggers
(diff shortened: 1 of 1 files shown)


jrochkind:
@seanaery @cjcolvar -- I think this backport may have somehow broke modals in BL 8.x. 

Don't know why CI didn't catch it. 

If anyone had time to test manually to confirm or reject what I'm seeing, would be helpful for me! Trying to figure out what's going on.

jrochkind:
I'm getting 

```
Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at modal.setupModal (modal.js:168:17)
    at modal.js:213:9
    at modal.js:216:1
```

That's `modal.target()` in this PR being `undefined`.

jrochkind:
Oh wait no, it's simply that some pages in my app do not have a modal on them. And before running this JS was a no-op, now it's an error that prevents other JS from executing too. :(

Not caught my CI because all tests have a modal on the page. 

Is a problem in main too. OK, will have to resolve.

PR #3708, opened 2025-07-30
Avoid ViewComponent::Base#set_slot

jcoyne:
This is deprecated, and unavailable in ViewComponent 4.0.  We don't have any reason to do it anyway


jcoyne:
@cbeer yes, but that isn't available in view_component 3, which is what we're running here.

PR #3710, opened 2025-07-30
Remove unused positional arguments from component lambdas

jcoyne:


body-clock:
Could it be the case that someone using their own component (set in the `view_config`) is using some extra arguments that will no longer be passed on?

body-clock (approved):

PR #3711, opened 2025-07-31
Fix facet suggest catalog path

mksndz:
If you've changed the resource path names for CatalogController routes...

```ruby
resource :catalog, as: 'catalog', path: '/items'
```

...the facet suggest lookups won't work. [The `/catalog/` path is hard-coded into the FacetSuggest object](https://github.com/projectblacklight/blacklight/pull/3711/files#diff-f12939c2b9cc757ac9564932c71adea5ee785945e77d9764333368d2e551af5aL21). 

This MR extracts the base catalog actions path from the `facetSearchContext` that is already read in by the javascript.

I struggled to add a spec to cover this change. I could not find precedent for specs covering changed route names for catalog controller actions. Errors in this change would most likely be caught by existing specs covering facet suggest actions.


jcoyne:
Thank you @mksndz

PR #3732, opened 2025-08-19
Push the default id for the show page back to the partial.

cbeer:
It seems nicer to explicitly set the `id` where we want it.


seanaery (approved):
Good catch; I agree that this is better.

PR #3743, opened 2025-08-28
Resolve '3739 render multiple dd elements for multivalued fields'

mamrey:
This PR refactors the Blacklight rendering pipeline and related components to normalize return values to arrays.

Previously, by default, all values were joined into a single string (multivalues combined with to_sentence). As a result, metadata fields, even those with multiple values, were always rendered with a single `dd` element.

Now, the pipeline, field presenter, and document presenter consistently return arrays of values. The `MetadataFieldLayoutComponent` renders each value individually, so multivalued fields produce multiple `dd`s.

Furthermore, the join step now considers the context of the request format. Previously, we joined for all request formats. Now, the join step requires an html request format. No longer joining multiple values into a single string makes our default `json` response easier to parse and consume.

Key changes

- `FieldPresenter#render`, `DocumentPresenter#field_value`, and `Rendering::Terminator` now always return arrays.
- The join step is optional and no longer runs by default. Opt in with:
  - a `join: true` field config,
  - a `join: true` option passed to the field presenter
  - `separator_options `on the field config.
- [As done previously for fields with a single value](https://github.com/projectblacklight/blacklight/blob/09170e0882baf503c356fd86119cc5dc566ac83d/app/presenters/blacklight/rendering/join.rb#L8), we consistently rely on [rails erb output safety](https://edgeguides.rubyonrails.org/active_support_core_extensions.html#output-safety) when handling both single and multiple field values
- Logic added to `MetadataFieldLayoutComponent` to render multiple `dd` elements vertically with appropriate offset.
- Ensure document presenter `#heading` and `#html_title` continue to join titles and return a string
- Updated specs to expect arrays instead of strings.
- Describe 'join' contexts in field presenter and rendering pipeline specs

Screenshots

multivalued language field rendered on frontend
Screenshot 2025-08-28 at 2 56 40 PM

multiple `dd` elements for a single `dt`
Screenshot 2025-08-28 at 2 57 11 PM

multivalued json output
Screenshot 2025-08-28 at 2 58 03 PM

Backward compatibility
- Apps that expect strings from the default rendering pipeline will need to adjust (e.g., join values explicitly or iterate).
- Apps that want to continue joining metadata fields will need to configure the fields with the 'join' option.
  -  Perhaps raising this higher in the configuration may be helpful to preserve legacy behavior? (e.g. `config.index.join` or `config.show.join`)

Special Thanks

- @mksndz, who offered feedback and ideas for a first draft of these changes and whose work in our apps inspired this and #3728
- @seanaery and @jcoyne for providing really helpful examples of the many ways blacklight apps may achieve such a markup.

### DIFF
--- a/spec/presenters/blacklight/document_presenter_spec.rb
+++ b/spec/presenters/blacklight/document_presenter_spec.rb
@@ -29,13 +29,13 @@ RSpec.describe Blacklight::DocumentPresenter do
   end
 
   describe '#field_value' do
-    let(:field_presenter) { instance_double(Blacklight::FieldPresenter, render: 'xyz') }
+    let(:field_presenter) { instance_double(Blacklight::FieldPresenter, render: ['xyz']) }
     let(:field_config) { Blacklight::Configuration::DisplayField.new }
     let(:options) { { a: 1 } }
 
     it 'calls the field presenter' do
       allow(Blacklight::FieldPresenter).to receive(:new).with(request_context, doc, field_config, options).and_return(field_presenter)
-      expect(presenter.field_value(field_config, options)).to eq 'xyz'
+      expect(presenter.field_value(field_config, options)).to eq ['xyz']
     end
 
     it 'can be configured to use an alternate presenter' do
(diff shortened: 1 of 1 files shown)


seanaery (approved):
This looks really great to me, @mamrey -- thanks especially for the revisions/additions to the specs, and the screenshots demonstrating the changes. I'll approve and give just a bit more time before merging in case any other developers still want to review.

PR #3750, opened 2025-10-03
require view_component/version in engine to avoid path-dependence

adjam:
versions 3 and 4 of the `view_component` gem do not autoload `VERSION`, which can lead to uninitialized constant errors when running Blacklight-based apps in some environments.


jcoyne:
I believe this was resolved by #3729

PR #3757, opened 2025-10-27
CI on Rails 8.1.0 now that it's out instead of 8.1.0.rc1

jrochkind:


seanaery (approved):

jrochkind:
Aha, view_component 3 can't be used iwth Rails 8.1, so can't be tested with it, out of our hands, fair enough. Need to fix that.

PR #3760, opened 2025-10-29
Add a blacklight_config.default_facet_suggest, to allow turn off new facet suggest feature by default

jrochkind:
Doing this involved a bit of yak-shaving:

* DRY'd two components that were checking this to use FacetFieldPresenter#suggest consistently, to put logic there
* Changed name to #suggest? since it's boolean -- backwards incompat, but to little-used api and fine if we get it in before 9.0
* Relevant test needs a good view_context, which revealed FacetFieldPresenter's attempt to have default for search_state initializer arg was not working, fixed. (not totally sure why this was not causing problems before)

### DIFF
--- a/app/components/blacklight/facets/filters_component.rb
+++ b/app/components/blacklight/facets/filters_component.rb
@@ -21,7 +21,7 @@ module Blacklight::Facets
     delegate :display_facet, to: :presenter
 
     def render?
-      facet.suggest != false || render_index_navigation?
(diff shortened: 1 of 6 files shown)


jrochkind:
oh crap, I just realized what was going on with that initialier arg, and I didn't really have to change it, just had to change test setup.  Already merged, but will try to PR undo of that unnecessary deck chair shift.

PR #3761, opened 2025-10-30
bump versions for 9.0.0-beta8 release

jrochkind:


jrochkind:
My understanding is that community norms allow self-merge of version-bump-only PRs. Please let me know if I'm wrong and I won't do it again!

PR #3762, opened 2025-10-30
Revert "Extract FacetSearchBuilder"

jrochkind:
This would revert #3597

PR created  for discussion/demo, to see if it still reverts fairly cleanly and green CI. it should probably not be merged without disucssoin with original author and reviewer (@jcoyne and @barmintor). 

My argument:

SearchBuilder is generated into local application becuase it's an *intended* site of local customization, a "seam". Historically, it's where many plugins that want to interact with the search process have injected themselves. 

With the split to a separate FacetSearchBuilder -- any customization in SearchBuilder (or many, and any I can think of?) used in a given controller, *needs* to be duplicated in the corresponding FacetSearchBuilder on that same controller. If it is not, when you press 'more' to get the 'more facets' dialog/page will display the wrong facets, not those corresponding to the search you just came from.  (This was pointed out by @cbeer on original PR too in June). 

So this is a very jagged edge, waiting to trip people up. Any plugins that are used to injecting into SearchBuilder I guess have to inject into both places, or you'll get a potentially hard-to-notice bug in the facets displayed in more-facets. 

In real world apps making extensive use of this (intended, provided) seam for local customization, it can be pretty cumbersome.  In my BL8 app, I have a:

* `SearchBuilder` (with customization)
* `CollectionSearchBuilder < SearchBuilder`. (same app-wide customization from search builder, plus more)
* `FeaturedTopicSearchBuilder < SearchBuilder` (similar)

In the move to a BL9 with the current "Extract FacetSearchBuilder' architecture, if I want to keep it DRY between the places that need to match to avoid faulty behavior, I need to introduce a bunch of modules, as well as all the paired search builder classes. 

* `LocalSearchBuilderBehavior` (concern)
* `SearchBuilder` (`include LocalSearchBuilderBehavior`)
* `FacetSearchBuilder` (`include LocalSearchBuilderBehavior`)

_

* `LocalCollectionSearchBuilderBehavior` (concern)
* `CollectionSearchBuilder` (`include LocalCollectionSearchBuilderBehavior`) `< SearchBuilder`
* `CollectionFacetSearchBuilder` (`include LocalCollectionSearchBuilderBehavior`) < `FacetSearchBuilder`

_

* `LocalFeaturedTopicSearchBuilderBehavior` (concern)
* `FeaturedTopicSearchBuilder` (`include LocalFeaturedTopicSearchBuilderBehavior`) `< SearchBuilder`
* `FeaturedTopicFacetSearchBuilder` (`include LocalFeaturedTopicSearchBuilderBehavior`) `< FacetSearchBuilder`

i've gone from three local classes/modules to 9, with a much more confusing relationship between them. And every place I set (eg) `blacklight_config.search_builder_class = CollectionSearchBuilder`, I _need_ to make sure to _also_ set `blacklight_config.facet_search_builder_class = CollectionFacetSearchBuilder` -- or I'll have faulty behavior. For each pair. 

I think I get (more or less) the philosophical issue with single-responsibility principle here... but prinicples like that are meant to serve the cleanliness and maintainability of our code, right?

I'd like to suggest reverting this for now, until/unless the existing architecture is causing real pain and/or a solution can be found to "SRP" that provides a better developer DX, without the jagged edges and gotchas and need for boilerplate code.

### DIFF
--- a/.github/matrix.json
+++ b/.github/matrix.json
@@ -8,11 +8,13 @@
   "additional_name": [""],
   "include": [
     {
+      "ruby": "3.4",
       "rails_version": "8.1.1",
       "additional_engine_cart_rails_options": "--css=bootstrap --js=esbuild",
       "additional_name": "| esbuild"
     },
     {
+      "ruby": "3.4",
       "rails_version": "8.0.4",
       "additional_engine_cart_rails_options": "--css=bootstrap",
       "view_component_version": "~> 3.0",
@@ -21,7 +23,7 @@
     {
       "ruby": "3.3",
       "rails_version": "8.0.4",
-      "additional_engine_cart_rails_options": "--css=bootstrap"
+      "additional_engine_cart_rails_options": "--css=bootstrap",
       "additional_name": "| Rails 8.0 + Ruby 3.3"
     },
     {
@@ -37,12 +39,14 @@
       "additional_name": "| Solr 8.11.2 and Ruby 3.2"
     },
     {
+      "ruby": "3.4",
       "rails_version": "8.1.1",
       "api": "true",
       "additional_engine_cart_rails_options": "--api --skip-yarn",
       "additional_name": "| API"
     },
     {
+      "ruby": "3.4",
       "rails_version": "7.2.3",
       "additional_engine_cart_rails_options": "-a propshaft --css=bootstrap --js=esbuild",
       "additional_name": "| Rails 7.2 + Propshaft, esbuild"


jrochkind:
@barmintor Here are the SearchBuilders in my app:

[SearchBuilder](https://github.com/sciencehistory/scihist_digicoll/blob/9c896dcd3791365c9b9596cd027f25588852b3bb/app/models/search_builder.rb)

[WithinCollectionSearchBuilder](https://github.com/sciencehistory/scihist_digicoll/blob/9c896dcd3791365c9b9596cd027f25588852b3bb/app/models/search_builder/within_collection_builder.rb)

[WithinFeaturedTopicSearchBuilder](https://github.com/sciencehistory/scihist_digicoll/blob/9c896dcd3791365c9b9596cd027f25588852b3bb/app/models/search_builder/within_featured_topic_builder.rb)

In the current architecture, I don't think there's any way around each one needing a paired FacetSearchBuilder, with identical logic, which is always configured next to it?

jcoyne:
I just want to add context, that this was added to support ElasticSearch, which could potentially allow us to do hybrid searching with embeddings.   The FacetSearchBuilder part of this is something that I don't believe ElasticSearch can support, so it makes sense to have one module that can be turned off when not using solr.

barmintor:
> @barmintor Here are the SearchBuilders in my app:
> 
> [SearchBuilder](https://github.com/sciencehistory/scihist_digicoll/blob/9c896dcd3791365c9b9596cd027f25588852b3bb/app/models/search_builder.rb)
> 
> [WithinCollectionSearchBuilder](https://github.com/sciencehistory/scihist_digicoll/blob/9c896dcd3791365c9b9596cd027f25588852b3bb/app/models/search_builder/within_collection_builder.rb) (supports per-collection landing pages that let you search within a collection)
> 
> [WithinFeaturedTopicSearchBuilder](https://github.com/sciencehistory/scihist_digicoll/blob/9c896dcd3791365c9b9596cd027f25588852b3bb/app/models/search_builder/within_featured_topic_builder.rb) (supports landing pages that let you search within constructed "featured topics", defined by a list of subject headings, inclusive).
> 
> In the current architecture, I don't think there's any way around each one needing a paired FacetSearchBuilder, with identical logic, which is always configured next to it?

As I mentioned earlier, I would have handled this situation by having the relevant controllers configure the extra step that controller requires in the processing chain. Handling it via subclasses is not illegitimate - There's More Than One Way To Do It - but the maintainers need to decide in this case what type of design they want to optimize for. Searching within facets is a frequently requested feature here, so the design direction of #3597 was valuable to me. We also don't subclass SearchBuilder. And the general direction of supporting ElasticSearch is very valuable to me, as I think the vector support is much better (as is the migration path to OpenSearch).

jrochkind:
I also find it quite rude that, while I intentionally set this to "draft" knowing it should not simply be merged with one approver when I know it is a matter of debate -- others have proceeded to merge PR's on top that are dependent on this decision, with one approver, making this PR now in conflict when it merged simply when I made it.

barmintor:
@jcoyne @body-clock #3766 undoes one of the mechanisms that I am arguing makes #3597 acceptable, I agree that we should revert that change while this PR is under discussion.

ebenenglish:
I think it's worth noting that subclassing `SearchBuilder` has been a pattern described in the Blacklight wiki for quite some time:
https://github.com/projectblacklight/blacklight/wiki/Extending-or-Modifying-Blacklight-Search-Behavior#extending-the-searchbuilder

Whether people find it elegant or not, it's likely to have been used by many, many implementations at this point. I personally find this pattern very straightforward to use, and it gets the job done (setting Solr params for specific controllers/views) with a minimal amount of code. Or maybe I'm just used to it.

What would be helpful to me in this discussion is an idea of whether there are some changes that could make this less painful, which I think is what @barmintor means by "refining the merged work to facilitate migration"?

I'm totally on board with supporting ElasticSearch, are there other potential ways to due this that are less disruptive? Are suggestions like the one @jrochkind makes here https://github.com/projectblacklight/blacklight/pull/3762#issuecomment-3482381867 feasible?

barmintor:
> I think it's worth noting that subclassing SearchBuilder has been a pattern described in the Blacklight wiki for quite some time:
https://github.com/projectblacklight/blacklight/wiki/Extending-or-Modifying-Blacklight-Search-Behavior#extending-the-searchbuilder

@ebenenglish that documentation (introduced in [2015](https://github.com/projectblacklight/blacklight/wiki/Extending-or-Modifying-Blacklight-Search-Behavior/726f94d394c5490e798b0ece984fd6d8d99fdf5f) during the v6.0.0.pre3 cycle and last edited in 2021 during the v7.21.0 cycle) doesn't describe subclassing the generated SearchBuilder, it describes adding methods to the generated SearchBuilder - the amount of overhead in migrating such an app vs one that has a number of SearchBuilder subclasses is one of the issues here.

Subclassing the generated SearchBuilder might be inferred to be a customization strategy from the presence the configuration property for `search_builder_class`, but that attribute was introduced for v5.10.0 in [this commit](https://github.com/projectblacklight/blacklight/commit/50961c0da2c587df12039df3c0a9e52fc50a7761) so that different repository implementations could have appropriate mappings of BL query params to the backend query, whatever it may be, as indicated by the documenting comment in Blacklight::Configuration "class for converting Blacklight parameters to request parameters for the repository_class".

seanaery:
Thanks; merging #3779 will address the CI issues here...

PR #3764, opened 2025-11-03
Update supported Rails versions

jcoyne:
https://rubyonrails.org/2025/10/29/new-rails-releases-and-end-of-support-announcement

### DIFF
--- a/.github/matrix.json
+++ b/.github/matrix.json
@@ -1,65 +1,51 @@
 {
   "ruby": ["3.4"],
-  "rails_version": ["7.1.5.1", "7.2.2.1"],
+  "rails_version": ["8.1.1"],
   "bootstrap_version": [null],
   "view_component_version": ["~> 4.0"],
   "api": [null],
-  "additional_engine_cart_rails_options": [""],
+  "additional_engine_cart_rails_options": ["--css=bootstrap"],
   "additional_name": [""],
   "include": [
     {
-      "ruby": "3.4",
-      "rails_version": "8.1.0",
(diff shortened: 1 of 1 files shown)


jrochkind:
I worry that we have an increasingly split install base among Blacklight users, with very very few people using the latest versions, and that dropping old Rails aggressively with official EOL will contribute to this further splitting. 

There are costs to the sustainability of BL of having such a split install base, and I think supporting recently EOL'd Rails has a fairly low cost to BL maintainance generally while making it easier to upgrade to new BL versions, so reducing pressures that lead to split install base. 

The fact that Rails released the last bugfix relase of 7.0 and 7.1 just now suggests that they expect people to still be using htem -- the EOL means they won't be releasing any subsequent bugfixes though. 

Just an opinion, I am personally on recent Rails and not personally effected, except to the extent that I think the increasingly split BL install base makes it harder on the community in general. But if I am, as I expect, a minority of one again amongst those interested in speaking up, I'll just stand aside!  I am curious if we could get opinions shared by anyone who is *not* on the latest Rails or Blacklight already.

seanaery:
I see where you're coming from here @jrochkind but especially since this PR 1) is into `main`, targeting 9.0.0; and 2) keeps Rails 7.2 in the mix, I'm good with the change. Maybe 7.1 can linger just a bit more in the release-8.x branch? If someone's currently using BL8 with Rails 7.0/7.1, upgrading to Rails 7.2 during a BL9 upgrade would likely be one of the more trivial parts of the upgrade.

PR #3765, opened 2025-11-03
Add documentation about SearchBuilder and FacetSearchBuilder

jcoyne:


jrochkind:
Obv dependent on a decision on previous #3762

PR #3769, opened 2025-11-04
Redesign facet suggest input styling/layout

jrochkind:
Note that the previous one had a typo/mistake in i18n label -- there was already an i18n key entered for this, but it was using a different one, now it's using the intended one.

# Before

blacklight-stock-before

# After

Screenshot 2025-11-04 at 5 11 23 PM

## input is type=search

So on some browsers (Chrome and Safari, but not Firefox), you get a "erase input" `x` icon, which works very nicely with the JS to immediately reset to unfiltered display. 

Screenshot 2025-11-04 at 5 11 38 PM

### DIFF
--- a/app/components/blacklight/facets/filters_component.rb
+++ b/app/components/blacklight/facets/filters_component.rb
@@ -3,7 +3,7 @@
 module Blacklight::Facets
   class FiltersComponent < Blacklight::Component
     # @param [Blacklight::FacetFieldPresenter] presenter
-    def initialize(presenter:, classes: 'facet-filters card card-body bg-light p-3 mb-3 border-0',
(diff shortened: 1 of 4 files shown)


jrochkind:
Not sure if the little spyglass on the end needs an aria label... or should just be omitted, maybe it's too fussy and unnecessary?

Any opinions?

jrochkind:
Thanks! i guess blacklight already puts aria-hidden on the SVG, so I guess it's fine. 

https://github.com/projectblacklight/blacklight/blob/f8673e19148c73f46bc887a35ac29311d32fa927/app/components/blacklight/icons/search_component.rb#L10

PR #3774, opened 2025-11-12
Switch facet suggest request to have query in URL query params, instead of path, with proper escaping

jrochkind:
It was failing prior on queries including `.` or `/`.  It may have been possible to fix this with proper escaping and leave it in path component of URL (may also have to fight with rails routing to get `.` in path component) -- but it's much more straightoforward to put it in query component after the `?`, and let existing escaping happening take care of it.   This is a URL only used by JS fetch that shouldn't ordinarily be seen by the user, so aesthetics prob not a concern.

The Blacklight CI sample data has a `topic_facet` value with a period in it (`Iran. Vizārat-i Kishvar`), and my own local data does too. I haven't seen any with `/`, but of course it's possible and should work.

We _could_ have made the `query_fragment` required for the Rails route -- but it was optional before, and current usage perhaps requires it being optional!

### DIFF
--- a/app/javascript/blacklight-frontend/facet_suggest.js
+++ b/app/javascript/blacklight-frontend/facet_suggest.js
@@ -16,10 +16,13 @@ const FacetSuggest = async (e) => {
 
     // Drop facet.page so a filtered suggestion list will always start on page 1
     url.searchParams.delete('facet.page');
+    // add our queryFragment for facet filtering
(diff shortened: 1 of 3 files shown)


jrochkind:
> could you add a new feature spec

Like that these punctuation can now be searched? 

Sure, I guess the existing fixture makes that pretty do-able.

seanaery (approved):

PR #3778, opened 2025-11-19
Test on ruby 4

jcoyne:


jcoyne:
It's working now with -preview3

seanaery (approved):

PR #3813, opened 2026-03-27
Remove bundler version

jcoyne:
Let it float to whatever is the latest.


jrochkind:
For record keeping, when/why the pin was added in first place is here:

https://github.com/projectblacklight/blacklight/pull/3794

Curious if that problem has been fixed in some other way or is no longer a problem with latest bundlers... sadly, failure on this PR may mean no change. :(

jcoyne:
Yes. I thought it may work, but it seems like it doesn't change it.

PR #3817, opened 2026-04-23
Bl7 rails 8 upgrade

dkinzer:
I need to upgrade my 7.x site to the latest version of rails but in order to this I needed to upgrade bl-7.x to use the latest version of rails.


thatbudakguy:
Was this accomplished by https://github.com/projectblacklight/blacklight/pull/3874?

PR #3826, opened 2026-05-13
Run safe autocorrect on Style lint issues

jcoyne:


cjcolvar (approved):
Thanks for cleaning these up!

PR #3840, opened 2026-05-22
Remove unused method

jcoyne:


sandbergja (approved):
Thanks, @jcoyne !

cbeer:
Why do we think this is unused?

It was introduced in https://github.com/projectblacklight/blacklight/commit/30ef7802a40c6720d794657f281bcc0d374b52c7 and is still being called in https://github.com/projectblacklight/blacklight/blob/main/lib/blacklight/search_builder.rb#L31, as far as I can tell.

And, we're willing to change the public API without any deprecation?

jcoyne:
I thought SearchBuilder#scope was calling the method in the catalog_controller (https://github.com/projectblacklight/blacklight/blob/main/app/controllers/concerns/blacklight/controller.rb#L31) . There didn't seem to be any test coverage for this method at all.

PR #3841, opened 2026-05-22
Allow SearchService#search_state to be mutated

jcoyne:
This allows us to adjust the search_state and run a query again, without having to reinitialize the object.  

This can be used in a report where you may want to iterate through a bunch of pages of results to construct a CSV.

### DIFF
--- a/app/services/blacklight/search_service.rb
+++ b/app/services/blacklight/search_service.rb
@@ -14,6 +14,8 @@ module Blacklight
     # The blacklight_config + controller are accessed by the search_builder
     attr_reader :blacklight_config, :context
 
+    attr_writer :search_state
+
     def search_builder
       search_builder_class.new(self)
     end


cbeer:
I find mutating `search_state` pretty surprising. I wonder if updating `search_results` to accept either a `search_state` or the `search_builder` would get you there?

tpendragon:
@jcoyne Can you speak to the use case motivating wanting to do this? Like is a user asking for something and this makes that feature easier?

jcoyne:
@tpendragon I was thinking of paging through a result set to write a report.

So, initialize the SearchService once, then call search_results, update the page in the search_state, then call search_results again, repeat until done.

In the existing code, I have to have a different SearchService instance for each page of results.

jcoyne:
> I find mutating search_state pretty surprising. I wonder if updating search_results to accept either a search_state or the search_builder would get you there?

@cbeer  yes, I think that approach would work just as well.

jrochkind (approved):

jrochkind:
Do we need a new test ensuring as supported tested API?   I approved anyway, up to you.

PR #3843, opened 2026-05-22
Extract retrieve_* methods to Blacklight::Searchable for consistency.

cbeer:


app/controllers/concerns/blacklight/searchable.rb
```diff
@@ -24,6 +24,24 @@ def search_service
...
+  end
+
+  # This method may be overridden to customize search behavior.
+  # @return [Blacklight::Document] the solr document object
+  def retrieve_document(id)
+    search_service.fetch(id)
+  end
+
+  # This method may be overridden to customize search behavior.
+  # @return [Array<Blacklight::Document>] the solr document object
```
cjcolvar:
```suggestion
  # @return [Array<Blacklight::Document>] an array of solr document objects
```

cjcolvar (approved):
This makes sense to me!

PR #3848, opened 2026-05-22
Define SearchState#query_boolean_operator (instead of reaching around…

cbeer:
… to the raw parameters)

Suggestions for a better name encouraged.


jcoyne (approved):
I think this is a fine name.

tpendragon (approved):

PR #3851, opened 2026-05-22
Push search_state out of the SearchBuilder initializer 

cbeer:


jcoyne (commented):
I really like this idea.

jcoyne (approved):

PR #3854, opened 2026-05-22
Allow Blacklight::SearchService#search_results to take params as an argument

cbeer:
This may better support use cases like #3841 (although I'm not sure I understand that case), and with some additional work might let us remove the SearchService abstraction.

### DIFF
--- a/app/services/blacklight/search_service.rb
+++ b/app/services/blacklight/search_service.rb
@@ -25,13 +25,16 @@ module Blacklight
     #                         SearchBuilder to be used. Block should return SearchBuilder to be used.
     #                         This is used in blacklight_range_limit
     # @return [Blacklight::Solr::Response] the solr response object
-    def search_results
-      builder = search_builder.with(search_state)
-      builder.page = search_state.page
-      builder.rows = search_state.per_page
+    def search_results(params: nil)
+      unless params
+        builder = search_builder.with(search_state)
+        builder.page = search_state.page
+        builder.rows = search_state.per_page
 
-      builder = yield(builder) if block_given?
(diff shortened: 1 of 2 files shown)


app/services/blacklight/search_service.rb
```diff
@@ -17,21 +17,24 @@ def initialize(config:, search_state:, search_builder_class: config.search_build
...
     # Fetch query results from solr
     # @yield [search_builder] optional block yields configured SearchBuilder,  caller can modify or create new
     #                         SearchBuilder to be used. Block should return SearchBuilder to be used.
     #                         This is used in blacklight_range_limit
     # @return [Blacklight::Solr::Response] the solr response object
-    def search_results
-      builder = search_builder.with(search_state)
-      builder.page = search_state.page
-      builder.rows = search_state.per_page
+    def search_results(params: nil)
```
jcoyne, line 28: I was thinking more of passing in the search_state, rather than any params.   I think we should keep ActionController::Parameters out of this class.

cbeer: This isn't ActionController::Parameters, it's a SearchBuilder instance (named `params` to match what we're sending to `repository` 🤷‍♂️  )

jcoyne: I'm suggesting that what we are sending to repository is confusing and we should not perpetuate a confusing pattern.

cbeer: I guess I understand `params` here to be "the HTTP query params for the request we're making", and that doesn't seem confusing and I'm not sure what else we'd call it.

PR #3855, opened 2026-05-22
Construct searches using the search builder directly from the controller.

cbeer:
I'm not there's advantage to having `Blacklight::SearchService` construct the search builder internally (other than it probably made sense at the time, before we encapsulated the params with SearchState).

If we hoist the search builder up to the controller (via the `Searchable` concern), maybe there's less magic?

Bypassing `SearchService#facet_field_response`  and `#facet_suggest_response` are potentially backwards-incompatible, but I couldn't find any indication that it's a method people are actually overriding downstream.

Includes #3856, #3843, and #3854

### DIFF
--- a/app/controllers/concerns/blacklight/catalog.rb
+++ b/app/controllers/concerns/blacklight/catalog.rb
@@ -80,11 +80,13 @@ module Blacklight::Catalog
     @facet = blacklight_config.facet_fields[params[:id]]
     raise ActionController::RoutingError, 'Not Found' unless @facet
 
-    @response = if params[:query_fragment].present?
-                  search_service.facet_suggest_response(@facet.key, params[:query_fragment])
-                else
-                  search_service.facet_field_response(@facet.key)
-                end
+    builder = search_builder.with(search_state).facet(@facet.key)
+    if params[:query_fragment].present?
+      builder = builder.facet_suggestion_query(params[:query_fragment])
+    end
+
+    @response = retrieve_search_results(params: builder)
+
     # @display_facet is a Blacklight::Solr::Response::Facets::FacetField
     @display_facet = @response.aggregations[@facet.field]
 
--- a/app/controllers/concerns/blacklight/searchable.rb
+++ b/app/controllers/concerns/blacklight/searchable.rb
@@ -42,6 +42,11 @@ module Blacklight::Searchable
     search_service.fetch(Array(ids))
   end
 
+  # @return [Blacklight::SearchBuilder]
+  def search_builder
+    blacklight_config.search_builder_class.new(self, blacklight_config: blacklight_config).with(search_state)
+  end
+
   # Override this method on the class that includes Blacklight::Searchable to provide more context to the search service if necessary.
   # For example, if your search builder needs to be aware of the current user, override this method to return a hash including the current user.
   # Then the search builder could use some property about the current user to construct a constraint on the search.
--- a/app/controllers/concerns/blacklight/searchable.rb
+++ b/app/controllers/concerns/blacklight/searchable.rb
@@ -26,8 +26,8 @@ module Blacklight::Searchable
 
   # This method may be overridden to customize search behavior.
   # @return [Blacklight::Solr::Response] the solr response object
-  def retrieve_search_results
-    search_service.search_results
+  def retrieve_search_results(params: nil)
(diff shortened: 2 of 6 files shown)


barmintor:
@jrochkind observes in Slack:
"SearchBuilder has had more or less the same API for a long time, and a lot of things in local apps and dependent *-lights depend on it, so beware there are downsides to changing it in backwards incompatible ways"

ArcLight uses a [mixin](https://github.com/projectblacklight/arclight/blob/main/app/models/concerns/arclight/search_behavior.rb) that does not override the initializer.

Spotlight uses two: an [access control mixin](https://github.com/projectblacklight/spotlight/blob/main/app/models/concerns/spotlight/access_controls_enforcement_search_builder.rb) that does not override the initializer, and a [category browse mixin](https://github.com/projectblacklight/spotlight/blob/main/app/models/concerns/spotlight/browse_category_search_builder.rb) that also does not.

GeoBlacklight likewise uses a [mixin](https://github.com/geoblacklight/geoblacklight/blob/412d72535503860f5d101511eb36b39af6e090ca/app/models/concerns/geoblacklight/suppressed_records_search_behavior.rb) without overriding the initializer.

All of these expect SearchBuilder.default_processor_chain to be an appendable list of symbols identifying methods to be added to the processing chain for params. None of them directly instantiate a SearchBuilder outside their specs, but all of them instantiate a SearchBuilder without the `blacklight_config` kwarg in the specs for their mixins.

barmintor:
ArcLight doesn't appear to interact with SearchService at all.
GeoBlacklight references SearchService in [component spec](https://github.com/geoblacklight/geoblacklight/blob/412d72535503860f5d101511eb36b39af6e090ca/spec/components/geoblacklight/homepage_feature_facet_component_spec.rb).
Spotlight has a helper method that reproduces the logic from Blacklight::Searchable.search_service, but I only see references to it in Spotlight controller mixins.

cbeer:
@barmintor : I think you're observing that this change is backwards-compatible (and/or having done some due diligence in some complicated apps to assure that, even if it was in some way, it doesn't impact real-world use)?

barmintor:
@cbeer that's right - per Jonathan's concern, I think things are mostly ok, but there is a spec in GeoBlacklight that will fail and apps that use the Spotlight helper might have an issue (though I think it may be a vestigial method).

jcoyne (approved):

app/controllers/concerns/blacklight/searchable.rb
```diff
@@ -24,6 +24,36 @@ def search_service
     search_service_class.new(config: blacklight_config, search_state: search_state, **search_service_context)
   end
 
+  # This method may be overridden to customize search behavior.
+  # @return [Blacklight::Solr::Response] the solr response object
+  def retrieve_search_results(params: nil)
```
jcoyne, line 29: Should we change this kwarg to `builder` since we're always passing a SearchBuilder here?

app/controllers/concerns/blacklight/catalog.rb
```diff
@@ -80,11 +80,13 @@ def facet
...
-                  search_service.facet_suggest_response(@facet.key, params[:query_fragment])
-                else
-                  search_service.facet_field_response(@facet.key)
-                end
+    builder = search_builder.with(search_state).facet(@facet.key)
+    if params[:query_fragment].present?
+      builder = builder.facet_suggestion_query(params[:query_fragment])
+    end
+
+    @response = retrieve_search_results(params: builder)
```
jcoyne, line 88: This conflates search results (documents) from facet results. Can we keep these two concerns separated?

cbeer: Is there any value in doing that? I find the current state incredibly confusing and it's not clear to me what behavior the seam is intending to provide that isn't better suited for the search builder.

app/controllers/concerns/blacklight/searchable.rb
cbeer: I'm just matching the argument name to the repository for consistency.

app/controllers/concerns/blacklight/catalog.rb
jcoyne: I think that it is confusing, and having one method handle the one case makes it less so.

PR #3858, opened 2026-05-26
Move SearchService#previous_and_next_document_params into the SearchB…

cbeer:
…uilder.

### DIFF
--- a/app/services/blacklight/search_service.rb
+++ b/app/services/blacklight/search_service.rb
@@ -69,11 +69,24 @@ module Blacklight
 
     # Get the previous and next document from a search result
     # @return [Blacklight::Solr::Response, Array<Blacklight::SolrDocument>] the solr response and a list of the first and last document
-    def previous_and_next_documents_for_search(index, request_params, extra_controller_params = {})
-      p = previous_and_next_document_params(index)
-      new_state = request_params.is_a?(Blacklight::SearchState) ? request_params : Blacklight::SearchState.new(request_params, blacklight_config)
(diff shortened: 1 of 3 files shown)


lib/blacklight/search_builder.rb
```diff
@@ -162,6 +162,26 @@ def processed_parameters
       end
     end
 
+    def for_previous_and_next_documents(index, window = 1)
+      document_pagination_params = blacklight_config.document_pagination_params.dup
+
+      if document_pagination_params.empty?
+        document_pagination_params[:fl] = blacklight_config.document_model.unique_key
```
jcoyne: This adds solr specifics to the SearchBuilder.  I'd like to keep the Repository specific code in Solr::Repository, so that we may add an Elasticsearch adapter.

cbeer: This code faithfully reflects the extracted code. I think a logical next step is to remove the solr-specific elements into the solr-specific search builder.

jcoyne (approved):

app/services/blacklight/search_service.rb
```diff
@@ -122,27 +135,11 @@ def solr_opensearch_params(field)
...
-
-      if index > 0
-        solr_params[:start] = index - window # get one before
-        solr_params[:rows] = (2 * window) + 1 # and one after
-      else
-        solr_params[:start] = 0 # there is no previous doc
-        solr_params[:rows] = 2 * window # but there should be one after
-      end
-
-      solr_params[:facet] = false
```
jcoyne, line 143: Was this removal intentional?

PR #3861, opened 2026-05-28
Simplify search params

jcoyne:

### DIFF
--- a/lib/blacklight/solr/repository.rb
+++ b/lib/blacklight/solr/repository.rb
@@ -16,21 +16,31 @@ module Blacklight::Solr
     end
 
     # Find multiple documents by their ids
-    # @param [Hash] _params query parameters
-    def find_many(params)
-      search(params: params, path: blacklight_config.fetch_many_documents_path)
+    # @param [Blacklight::SearchBuilder] builder query parameters
+    def find_many(builder)
+      search(builder: builder, path: blacklight_config.fetch_many_documents_path)
     end
 
     ##
     # Execute a search query against solr
-    # @param [Hash,Blacklight::SearchBuilder] params solr query parameters
+    # @param [Hash] params solr query parameters - deprecated
+    # @param [Blacklight::SearchBuilder] builder solr query parameters
     # @param [String] path solr request handler path
-    def search pos_params = nil, path: nil, params: nil, **kwargs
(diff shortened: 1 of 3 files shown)


lib/blacklight/solr/repository.rb
```diff
@@ -23,14 +23,17 @@ def find_many(params)
...
-    # @param [Hash,Blacklight::SearchBuilder] params solr query parameters
+    # @param [Blacklight::SearchBuilder] params solr query parameters
     # @param [String] path solr request handler path
     def search pos_params = nil, path: nil, params: nil, **kwargs
       if pos_params
         Blacklight.deprecation.warn("Passing positional arguments to search() is deprecated. Use the params kwarg instead.")
       end
 
-      request_params = (params || pos_params || {}).reverse_merge(kwargs).reverse_merge({ qt: blacklight_config.qt })
+      if params && !params.is_a?(Blacklight::SearchBuilder)
```
cbeer: I really don't like passing the search builder this deep into the stack. If nothing else, it seems like it'd make performing arbitrary queries really annoying.

jcoyne: We already are passing the search builder here. I'm just trying to align the parameter names with the object.

cbeer: I don't think that's the intent, it's just that SearchBuilder is sufficiently hash-like. Forcing us to pass a SearchBuilder all the way down the stack seems really annoying and I'm not sure what it gets us.

jcoyne: I'm not forcing a SearchBuilder to be passed. That work was already done. I've just updated the unit tests to reflect how we actually use it.

cbeer:
-1. Again, I think this is a really bad change. I don't think there's any need for a SearchBuilder object down in repository and will make sending arbitrary requests much harder/annoying. 

That you can pass a `SearchBuilder` object isn't the point; it's sufficiently hash-like that it works. If that's confusing, I'd much rather we call `to_h` on the `SearchBuilder` object before we pass it down. I don't think that step is necessary, but if it makes things less confusing, that's fine with me.

jcoyne:
I think one of the major problems we have with complexity in Blacklight is that we're supporting two paths through the application, using the same variable names, but with different expected types.  Since Blacklight itself is only using one of those paths, why not deprecate (with the intent to remove) the other path?

PR #3866, opened 2026-05-29
Remove solr-specific fields from Blacklight::SearchService

jcoyne:
This supports a agnostic backend.

### DIFF
--- a/lib/blacklight/search_builder.rb
+++ b/lib/blacklight/search_builder.rb
@@ -162,8 +162,12 @@ module Blacklight
       end
     end
 
+    # An abstract method intended to be overriden with the repository specific SearchBuilderBehavior.
+    #
+    # @return [Hash]
     def default_document_pagination_params
-      { fl: blacklight_config.document_model.unique_key }
+      Blacklight.logger.warn("#{self.class}#default_document_pagination_params is expected to be overridden, but was not.")
+      {}
     end
 
     def for_previous_and_next_documents(index, window = 1)
--- a/lib/blacklight/solr/search_builder_behavior.rb
+++ b/lib/blacklight/solr/search_builder_behavior.rb
@@ -100,6 +100,18 @@ module Blacklight::Solr
       solr_parameters.append_boolean_query(:must, bool_query)
     end
 
+    # Solr-specific request parameters for fetching documents with a
+    # minimal field set and without faceting. Used by
+    # Blacklight::SearchService when paging to the previous/next document.
+    #
+    # @return [Hash]
+    def default_document_pagination_params
+      {
+        fl: blacklight_config.document_model.unique_key,
+        facet: false
+      }
+    end
+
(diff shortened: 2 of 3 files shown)


cbeer:
I have an alternative, existing PR https://github.com/projectblacklight/blacklight/pull/3858 that pushes this into the search builder instead of adding yet more configuration.

jcoyne:
@cbeer I merged #3858 and then rebased this.

lib/blacklight/solr/repository.rb
```diff
@@ -42,6 +42,19 @@ def suggestions(request_params)
       Blacklight::Suggest::Response.new suggest_results, request_params, suggest_handler_path, suggester_name
     end
 
+    # Solr-specific request parameters for fetching documents with a
```
cbeer: I guess I don't understand why the repository needs to be involved in parameter mapping. I thought that was the job of the search builder.

jcoyne: The SearchBuilder is not repository specific. We need a way to do this same on other repositories, like Elasticsearch, that don't use the `qt` parameter.

cbeer: I thought the whole point of e.g. Blacklight::Solr::SearchBuilderBehavior to provide those service-specific mappings. Doing mappings in multiple places seems like it'll get confusing.

jcoyne: That is also an option. I find that the mixin approach is a bit more confusing and harder to test than the delegation approach.

jcoyne:
@cbeer I've moved it to `Blacklight::Solr::SearchBuilderBehavior`

lib/blacklight/search_builder.rb
```diff
@@ -162,10 +162,6 @@ def processed_parameters
       end
     end
 
-    def default_document_pagination_params
-      { fl: blacklight_config.document_model.unique_key }
-    end
-
```
cbeer, line 168: I intentionally put this in SearchBuilder to ensure backwards compatibility. Removing it seems like it could throw a no method error.

jcoyne: It seems incongruous to me that a Solr specific implementation is in `Blacklight::SearchBuilder`, but if we had an ElasticSearch implementation, then it would be in `Blacklight::ElasticSearch::SearchBuilderBehavior`.   It would also be extremely unlikely that there are any Blacklight 9 apps that exist that don't already `include Blacklight::Solr::SearchBuilderBehavior`, right?

cbeer: Is it? https://github.com/sul-dlss/SearchWorks/blob/main/app/models/eds/search_builder.rb is certainly one.

jcoyne: In that case it's not using Solr, so `fq` doesn't work, right?  Would it be enough to add a no-op implementation that warns and a separate Solr implementation?

PR #3870, opened 2026-05-29
Set the number of rows using the SearchBuilder

jcoyne:
This enables us to switch to an alternative repository backend

### DIFF
--- a/app/services/blacklight/search_service.rb
+++ b/app/services/blacklight/search_service.rb
@@ -152,11 +152,20 @@ module Blacklight
     def fetch_many(ids, extra_controller_params)
       extra_controller_params ||= {}
 
+      requested_rows = extra_controller_params.delete(:rows)
+      if requested_rows
+        Blacklight.deprecation.warn("Passing :rows to fetch_many is deprecated. Create an issue in blacklight if you see this warning. " \
+                                    "Otherwise, :rows will be ignored in Blackight 10")
+      else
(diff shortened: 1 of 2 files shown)


app/services/blacklight/search_service.rb
```diff
@@ -148,11 +148,13 @@ def previous_and_next_document_params(*, **)
     def fetch_many(ids, extra_controller_params)
       extra_controller_params ||= {}
 
+      extra_controller_params.delete(:rows)
       query = search_builder
               .with(search_state)
               .where(blacklight_config.document_model.unique_key => ids)
               .merge(blacklight_config.fetch_many_document_params)
               .merge(extra_controller_params)
+      query.rows(ids.count)
```
cjcolvar: It seems like it was possible to pass a `rows` value that was different than the length of the id array.  How about retaining that for backwards compatibility?
```suggestion
      requested_rows = extra_controller_params.delete(:rows)
query = search_builder
.with(search_state)
.where(blacklight_config.document_model.unique_key => ids)
.merge(blacklight_config.fetch_many_document_params)
.merge(extra_controller_params)
      query.rows(requested_rows || ids.count)
```

jcoyne: Yes, it was possible to do that, but is there a use case for that?

cjcolvar (approved):
👍 Looks good!

PR #3874, opened 2026-06-05
Add support for Rails 8.1 to Blacklight 7

tpendragon:
This will save my team from months of maintenance work in a time when we really need to be pushing on iterating features for product fit and migration use cases.

I believe it's fully backwards compatible, initial tests are showing this opens the door to upgrading old Geoblacklight, Spotlight, and vanilla Blacklight applications.

### DIFF
--- a/.github/matrix.json
+++ b/.github/matrix.json
@@ -39,6 +39,15 @@
       "rails_version": "8.0.2"
     },
     {
+      "ruby": "3.4",
+      "rails_version": "8.1.3"
+    },
+    {
+      "ruby": "3.4",
+      "rails_version": "8.1.3",
+      "bootstrap_version": "~> 4.0"
+    },
+    {
       "ruby": "3.3",
       "rails_version": "7.1.5.1",
       "view_component_version": "~> 2.74",
--- a/lib/generators/blacklight/templates/solr/conf/solrconfig.xml
+++ b/lib/generators/blacklight/templates/solr/conf/solrconfig.xml
@@ -47,7 +47,6 @@
           full_title_tsim
           short_title_tsim
           alternative_title_tsim
-          active_fedora_model_ssi
           title_tsim
           author_tsim
           subject_tsim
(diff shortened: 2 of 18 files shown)


.github/matrix.json
```diff
@@ -38,6 +38,15 @@
...
       "rails_version": "8.0.2"
     },
+    {
+      "ruby": "3.4",
+      "rails_version": "8.1.3"
+    },
+    {
+      "ruby": "3.4",
+      "rails_version": "8.1.3",
+      "bootstrap_version": "~> 4.0"
```
orangewolf, line 48: is 4 the right version here? My understanding was that BL 7 required Boostrap ~> 5.0

lib/generators/blacklight/templates/solr/conf/solrconfig.xml
```diff
@@ -23,7 +23,7 @@
 
   <dataDir>${solr.data.dir:}</dataDir>
 
-  <requestHandler name="search" class="solr.SearchHandler" default="true">
+  <requestHandler name="/select" class="solr.SearchHandler" default="true">
```
orangewolf, line 26: is this and the af_model change related to the rails 8 upgrade?

.github/matrix.json
tpendragon: It is! Blacklight 7 can support Bootstrap 4 and 5.

Which thank goodness, a bunch of my apps are still on 4.

lib/generators/blacklight/templates/solr/conf/solrconfig.xml
tpendragon: No, but CI doesn't pass without them since new versions of Solr require it. The af_model change was just me pulling down the new schema from main.

orangewolf: that makes sense then

orangewolf (approved):
This is working for me locally with a fresh Rails 8.1 / Blacklight build.

jcoyne (approved):

PR #3877, opened 2026-06-23
Dark mode support

thatbudakguy:
- **Add dark mode theme switcher**
  This adds a new component to the main navbar for switching between
  themes, based on Bootstrap's docs and provided example for how
  to implement a dark mode toggle.
  
  Also adds related JavaScript for detecting and saving the user's
  preference in localStorage, and updating the switcher to indicate
  which theme is active.
  

- **Update styles for dark mode compatibility**
  This tweaks the existing styles to handle cases where colors
  were not set up to handle dark mode.

  Also replaces the logo image with two SVG versions; one light
  and one dark.

- **Make dark mode configurable**
  This adds a configuration property called dark_mode_support for
turning dark mode off entirely, in case folks need more time to
migrate their components and styles before enabling the theme
switcher. It hides the theme switcher and uses light mode.


jcoyne:
can we just do this:
```css
@media (prefers-color-scheme: dark) {
    :root {
        --bs-body-color: white;
        --bs-body-bg: darkgray;
        /* etc. etc. */
    }
  }
}

```

And not have a switcher?

thatbudakguy:
That's what the "auto" mode in the switcher does, which I think is the default if you haven't set a preference.

In general I think the approach of just detecting the user's browser/OS preference is fine, but I've heard that it doesn't work for everyone – some people want to be able to control that preference on a per-app basis, because the main content of the app affects how they read (is it a long page of text? are there lots of fiddly UI elements? etc).

Enabling it without a switcher could also lead to confusion if someone gets the dark theme but doesn't know why and can't opt back into the light theme.

seanaery:
Thanks so much @thatbudakguy for working on this! A few thoughts:
* I like that you're using [Bootstrap 5's recommended conventions](https://getbootstrap.com/docs/5.3/customize/color-modes/) for color modes here
* I also like the combo of auto-detection (to respect a user's system settings) plus providing an ability to toggle and for that manual selection to stick.
* It'd be helpful to include a set of representative screenshots in the PR for a change like this that modifies the appearance of the UI so drastically
* The feature specs you've added are good, but since dark mode adds more considerations for accessibility I think it's important to add accompanying accessibility specs. There's a basic set here ( https://github.com/projectblacklight/blacklight/blob/main/spec/features/axe_spec.rb ) that could likely be extended to test these assertions in dark mode.

thatbudakguy:
Thanks for the review! I'll look at adding some additional tests; checking the contrast for both dark and light mode is probably a good idea.

thatbudakguy:
# Screenshots

## Switcher

Screenshot 2026-07-15 at 09 34 48

## Home

<table>
<tr>
<td>
home_light
</td>
<td>
home_dark
</td>
</tr>
</table>

## Index

<table>
<tr>
<td>
index_dark
</td>
<td>
index_light
</td>
</tr>
</table>

## Show

<table>
<tr>
<td>
show_dark
</td>
<td>
show_light
</td>
</tr>
</table>

thatbudakguy:
I've changed the a11y tests so they now all run separately on both dark and light mode, and fixed the resulting contrast errors via a few targeted bootstrap overrides.

Also added a new configuration property called `dark_mode_support` that defaults to `true`; if it's `false` we just don't render the theme switcher component, and we don't attempt to check the user's system preference and set the appropriate class on the `<html>` tag, so everything stays in light mode.

seanaery (approved):
This looks good to me, and addresses the areas discussed in the 7/15/2026 BL community call (default to on, adds a feature-flag in config, fixes & tests for contrast accessibility). I'll approve without merging for now, to leave some space for other folks to review if they'd like to, especially institutions already using BL 9.0, to verify that this would smooth out the anticipated 9.1 update experience ( @mksndz et. al ).

jcoyne (approved):

PR #3881, opened 2026-06-30
Allow using Rails default authentication in lieu of Devise

jcoyne:
We no longer need to have a Devise dependency as rails now has a generator for authenticated sessions.


seanaery:
Discussed briefly at Jul 15, 2026 community meeting. OK to proceed.

seanaery (approved):

PR #3883, opened 2026-07-17
[8.x] Get 8.x passing again in CI (address failed checks caused by updates to Rubocop, Solr, ViewComponent)

seanaery:
- fix rubocop violations
- add Solr 10 support
- fix a helper spec that had started failing with view_component 3.x with 3.25.0+


lib/generators/blacklight/templates/solr/conf/solrconfig.xml
```diff
@@ -257,7 +257,7 @@
 
   <requestHandler name="/replication" class="solr.ReplicationHandler" startup="lazy" />
 
-  <requestDispatcher handleSelect="true" >
```
cjcolvar, line 260: FWIW `main` still has this `handleSelect="true"` in its solrconfig.  Maybe it should be removed it from there too?

spec/helpers/blacklight_helper_spec.rb
```diff
@@ -172,6 +172,12 @@
     end
 
     it "ignores missing templates" do
+      # In ViewComponent 3.25.0+, collection components rendered from a plain
+      # ActionView template fall back to controller.view_context for helpers
+      # instead of the stubbed helper. Workaround not needed on 2.x or 4.x.
+      if ViewComponent::VERSION::MAJOR == 3
+        allow(controller).to receive(:view_context).and_return(helper)
+      end
```
cjcolvar, line 180: Nice catch!

cjcolvar (commented):
Thanks for working on this! I see the build is green but I'm slightly concerned about a drift in configuration between main and the 8.x branch.  Would it be possible to backport #3806 and #3819 so the solr configs and rubocop configs stay aligned?

seanaery:
Thanks @cjcolvar for the good feedback. I have made more changes in response (including cherry-picking a few relevant commits from main) to address those drift concerns and get this more aligned with what’s currently on `main`. That of course has turned this into a much bigger PR (now 78 files changed) :-) 

### Solr Configs
- `schema.xml` is now identical to what's on main.
- `solrconfig.xml` is now almost identical to main. I left in the BL8 `/advanced` requestHandler, as BL9 included a lot of Advanced Search refactoring that likely shouldn’t be backported. And yes, `<requestDispatcher handleSelect="true">` should be removed from main, as that has been obsolete since Solr 7.

### Rubocop
- `.rubocop.yml` is now identical to what’s on main. 

### Test Matrix
- I added some Ruby 3.4 & Rails 8.1 scenarios to the matrix and dropped Ruby 3.1.
- I also added a scenario that verifies that a Rails 8.x + BL8 app generation still works without the `--css=bootstrap` option; this required some shims in the `ImportmapGenerator` to ensure the Sprockets fallbacks for CSS actually work.
- Since BL8 is naturally more permissive than BL9 on some dependency versions (esp. Bootstrap 4-5; ViewComponent 2-4; Sprockets still semi-supported) it was a little tricky to get all tests passing, but it is now green again.

cjcolvar (approved):
Amazing work! ✨ 

This definitely was a lot more than I was originally thinking when I suggested the cherry-picking.  Thanks for devoting this much to it!  I think this puts the community in a good position for helping those who are still on 8.x (including me 🙂).

lib/blacklight/open_struct_with_hash_access.rb
```diff
@@ -53,19 +53,19 @@ def deep_dup
...
       if method_name.nil? && block_given?
         if b.arity.zero?
-          instance_eval(&block)
+          instance_eval(&)
         else
           yield self
         end
       elsif respond_to?(method_name)
-        public_send(method_name, *args, &b)
+        public_send(method_name, *, &b)
```
barmintor: there is a bug here, although it existed before in an untested edge case. The named block parameter doesn't exist. I can add a test that fixes this as my review.

barmintor (changes requested):
| change type | num files  |
| ----------- | --------- | 
| create sprockets manifest if needed | 1 |
| detect autoloaded module | 1 |
| disable OneClassPerFile to defer refactor | 1 |
| disambiguate hash and kwargs | 1 |
| do not fall through case statement for return value | 1 |
| have_content --> have_text | 19 |
| have_no_selector --> have_no_css | 1 |
| implied self.X | 2 |
| match --> include | 9 |
| merged hash literals | 1 |
| nil assets in production mode | 1 |
| prefer hash.slice or hash.grep over select | 2 |
| redirect_back_or_to | 1 |
| rubocop scaffolding | 2 |
| Ruby 3.4 hash serialization | 1 |
| solr scaffolding | 2 |
| sort keys as map | 1 |
| terminal varargs/closure style | 22 |
| test configuration | 1 |
| use File.dirname parameters over nested calls | 1 |
| ViewComponent 3 spec handling | 1 |
| whitespace | 6 |
| Grand Total | 78 |

Ok, so these changes look fine, but there's a minor bug that I think we should fix in `lib/blacklight/open_struct_with_hash_access.rb` that is more visible now with the anonymous block params.

barmintor (approved):

PR #3884, opened 2026-07-31
Stop testing on rails 7.2

jcoyne:
Rails 7.2 is EOL as of Aug 9, 2026 https://endoflife.date/rails

### DIFF
--- a/.github/matrix.json
+++ b/.github/matrix.json
@@ -44,12 +44,6 @@
       "api": "true",
       "additional_engine_cart_rails_options": "--api --skip-yarn",
       "additional_name": "| API"
-    },
-    {
-      "ruby": "3.4",
-      "rails_version": "7.2.3",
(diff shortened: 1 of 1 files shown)


seanaery (commented):
Thanks for keeping on top of this. As part of this removal -- or as a followup PR -- can we add the `-a propshaft` to one of the remaining test scenarios? If I understand correctly, I think this 7.2 scenario is the only one in the matrix currently testing the `PropshaftGenerator`.

jcoyne:
@seanaery propshaft is the default in Rails 8.

seanaery:
@jcoyne Oh right -- thanks! Now I see that all the remaining Rails 8 scenarios that specify `--js=esbuild` are indeed doing `generate  blacklight:assets:propshaft` and the rest are doing `generate  blacklight:assets:importmap`.

seanaery (approved):

PR #3886, opened 2026-08-10
Qualify ::String in JsonValue#cast so :json type parses JSON

### LINKED ISSUE #3885: json SolrDocument attribute type never parses JSON strings
`Blacklight::Types::JsonValue#cast `guards with `value.is_a?(String)`, but this method lives inside the `Blacklight::Types` module, where `String` resolves to `Blacklight::Types::String` (defined in the same file), not Ruby's `::String`. A JSON string is never an instance of `Blacklight::Types::String`, so the method always returns the raw string instead of parsed JSON.

Can be reproduced in the Rails console:
```
internal(dev)* class SolrDocument
internal(dev)*   include Blacklight::Solr::Document
internal(dev)*   attribute :metadata, :json, field: 'metadata_ssi'
internal(dev)> end
=> :metadata
internal(dev)> SolrDocument.new(metadata_ssi: '{"pages":288}').metadata
=> "{\"pages\":288}"

# expected: {"pages" => 288}
```

corylown:
fixes #3885

### DIFF
--- a/app/values/blacklight/types.rb
+++ b/app/values/blacklight/types.rb
@@ -88,7 +88,7 @@ module Blacklight
       def cast(input)
         value = super
 
-        return value unless value.is_a?(String)
+        return value unless value.is_a?(::String)
 
         JSON.parse(value)
       end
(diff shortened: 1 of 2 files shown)


seanaery (approved):
Nice find here!

jcoyne (approved):

PR #3887, opened 2026-08-19
Verify that OpenStructWithHashAccess.try correctly references block params

barmintor:


barmintor:
@seanaery I'll add this commit to the end of #fix-8x-ci if/when CI runs.

PR #3899, opened 2026-09-07
Add Solr request time and query count to the Rails request log line

maxkadel:
Rails' `Completed ... (ActiveRecord: ...)` log line has no equivalent for Solr, so query time is invisible at the default `info` log level. Adds a `Blacklight::LogSubscriber`/`Blacklight::ControllerRuntime` pair that mirrors how `ActiveRecord::Railties::ControllerRuntime` contributes to that same line. 

It now reads, for example:

```log
Completed 200 OK in 62ms (Views: 12.0ms | ActiveRecord: 1.2ms | Solr: 45.3ms (2 queries))
```

Solr-specific for now (event name, log label) rather than generalized to `AbstractRepository` for other backends (Elasticsearch/EDS/etc.) — happy to generalize if that's wanted, but didn't want to guess at the right abstraction without a concrete second backend to design against.

### DIFF
--- a/app/controllers/concerns/blacklight/controller_runtime.rb
+++ b/app/controllers/concerns/blacklight/controller_runtime.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Blacklight
+  # Adds Solr timing to the Rails request log line, mirroring how
+  # ActiveRecord::Railties::ControllerRuntime adds ActiveRecord timing, e.g.:
+  #   Completed 200 OK in 62ms (Views: 12.0ms | ActiveRecord: 1.2ms | Solr: 45.3ms (2 queries))
+  module ControllerRuntime
+    extend ActiveSupport::Concern
+
+    protected
+
+    def process_action(action, *args)
+      Blacklight::RuntimeRegistry.stats.reset
+      super
+    end
+
+    def append_info_to_payload(payload)
+      super
+      stats = Blacklight::RuntimeRegistry.stats
+      payload[:solr_runtime] = stats.solr_runtime
+      payload[:solr_query_count] = stats.solr_query_count
+      stats.reset
+    end
+
+    module ClassMethods
+      def log_process_action(payload)
+        messages = super
+        count = payload[:solr_query_count].to_i
+        return messages unless count.positive?
+
+        messages << solr_log_message(payload[:solr_runtime], count)
+        messages
+      end
+
+      private
+
+      def solr_log_message(runtime, count)
+        format("Solr: %.1fms (%d %s)", runtime.to_f, count, "query".pluralize(count))
+      end
+    end
+  end
+end
--- a/lib/blacklight/abstract_repository.rb
+++ b/lib/blacklight/abstract_repository.rb
@@ -10,9 +10,11 @@ module Blacklight
     attr_writer :connection
 
     # (Note: ActiveSupport::Benchmarkable requires a logger method)
+    # No longer used in core Blacklight, keeping for downstream consumers
     # @return [Logger]
     attr_writer :logger
 
+    # No longer used in core Blacklight, keeping for downstream consumers
     include ActiveSupport::Benchmarkable
 
     ##
--- a/lib/blacklight/log_subscriber.rb
+++ b/lib/blacklight/log_subscriber.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Blacklight
+  # Debug-only logging for solr_request.blacklight notifications
+  # see Blacklight::ControllerRuntime for the :info-level per-request summary.
+  class LogSubscriber < ActiveSupport::LogSubscriber
+    # ActiveSupport::LogSubscriber#debug/#info call the instance method
+    # `logger`, which the base class hardcodes to `LogSubscriber.logger`
+    # (itself) rather than `self.class.logger`. If that's ever fixed
+    # upstream, a `self.logger` override here would be worth implementing.
+    delegate :logger, to: :Blacklight
+
+    def solr_request(event)
+      payload = event.payload
+      debug { "Solr fetch (#{event.duration.round(1)}ms): #{payload[:method]} #{payload[:path]} #{payload[:params].to_hash.inspect}" }
+      debug { "Solr response: #{payload[:response].inspect}" } if verbose_logging?
+    end
+
+    private
+
+    def verbose_logging?
+      defined?(::BLACKLIGHT_VERBOSE_LOGGING) && ::BLACKLIGHT_VERBOSE_LOGGING
+    end
+  end
+end
--- a/lib/blacklight/solr/repository.rb
+++ b/lib/blacklight/solr/repository.rb
@@ -38,8 +38,10 @@ module Blacklight::Solr
     # @param [Hash] request_params
     # @return [Blacklight::Suggest::Response]
     def suggestions(request_params)
-      suggest_results = connection.send_and_receive(suggest_handler_path, params: request_params)
-      Blacklight::Suggest::Response.new suggest_results, request_params, suggest_handler_path, suggester_name
+      instrument_solr_request(suggest_handler_path, request_params) do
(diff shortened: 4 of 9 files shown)


app/controllers/concerns/blacklight/controller_runtime.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Blacklight
+  # Adds Solr timing to the Rails request log line, mirroring how
+  # ActiveRecord::Railties::ControllerRuntime adds ActiveRecord timing, e.g.:
+  #   Completed 200 OK in 62ms (Views: 12.0ms | ActiveRecord: 1.2ms | Solr: 45.3ms (2 queries))
+  module ControllerRuntime
+    extend ActiveSupport::Concern
+
+    def initialize(...)
```
jcoyne: Is this method necessary?

lib/blacklight/log_subscriber.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Blacklight
+  # Accumulates the runtime of "solr_request.blacklight" notifications
+  # (emitted by Blacklight::Solr::Repository#send_and_receive) so it can be
+  # surfaced on the Rails request log line by Blacklight::ControllerRuntime,
+  # the same way ActiveRecord::LogSubscriber tracks db_runtime.
+  class LogSubscriber < ActiveSupport::LogSubscriber
+    class << self
```
jcoyne: This pattern, where everything is a static method, seems rather unlike the patterns show in the documentation.  Can these all be instance methods?

jcoyne (changes requested):

lib/blacklight/log_subscriber.rb
maxkadel: Sorry, the pattern was based on the old Rails 7.0 pattern, not the newer Rails 8 / 9 patterns. That's my bad.

app/controllers/concerns/blacklight/controller_runtime.rb
maxkadel: No - was using the old pattern, not needed in the new one.

jcoyne (changes requested):
Should  lib/blacklight/solr/repository.rb:41 — #suggestions bypass the instrumentation?

app/controllers/concerns/blacklight/controller_runtime.rb
```diff
@@ -0,0 +1,37 @@
...
+      payload[:solr_runtime] = stats.solr_runtime
+      payload[:solr_query_count] = stats.solr_query_count
+      stats.reset
+    end
+
+    module ClassMethods
+      def log_process_action(payload)
+        messages = super
+        solr_runtime = payload[:solr_runtime]
+        return messages unless solr_runtime
```
jcoyne: Is there anyway this code could be triggered?

lib/blacklight/solr/repository.rb
```diff
@@ -70,7 +70,9 @@ def ping?
     # @return [Blacklight::Solr::Response] the solr response object
     def send_and_receive(path, solr_params = {})
       benchmark("Solr fetch", level: :debug) do
-        res = connection.send_and_receive(path, build_solr_request(solr_params))
+        res = ActiveSupport::Notifications.instrument("solr_request.blacklight", path: path, params: solr_params) do
+          connection.send_and_receive(path, build_solr_request(solr_params))
```
jcoyne: Should the `build_solr_request` be moved outside of the instrumentation block?

app/controllers/concerns/blacklight/controller_runtime.rb
maxkadel: Nope, changed it to a `#positive?` check.

lib/blacklight/solr/repository.rb
maxkadel: Yes - done.

maxkadel:
Thanks for the thorough review @jcoyne, I really appreciate it.

I did add `#suggestions` to the implementation, as well as `#ping?`, since I can see myself asking whether all of my Solr instance was slow at a given point, or just one specific query.

The other thing I was thinking was that we could refactor the debug lines in `lib/blacklight/solr/repository.rb` to use this pattern as well, but that seemed like a larger refactor, and it can wait.

app/controllers/concerns/blacklight/controller_runtime.rb
maxkadel: Nope - fixed for a `positive?` check rather than a "falsey" check, since `0.0` was getting evaluated as `truthy`.

lib/blacklight/solr/repository.rb
```diff
@@ -70,9 +70,8 @@ def ping?
     # @return [Blacklight::Solr::Response] the solr response object
     def send_and_receive(path, solr_params = {})
       benchmark("Solr fetch", level: :debug) do
-        res = ActiveSupport::Notifications.instrument("solr_request.blacklight", path: path, params: solr_params) do
-          connection.send_and_receive(path, build_solr_request(solr_params))
-        end
+        request = build_solr_request(solr_params)
```
jcoyne: This is still being done within the benchmark.  Don't we want this outside of the benchmark?

lib/blacklight/solr/repository.rb
```diff
@@ -70,7 +70,8 @@ def ping?
     # @return [Blacklight::Solr::Response] the solr response object
     def send_and_receive(path, solr_params = {})
       benchmark("Solr fetch", level: :debug) do
```
jcoyne: Isn't this benchmark() call redundant now?  Also, should we consider removing AcitiveSupport::Benchmarkable from Blacklight::AbstractRepository?

jcoyne (changes requested):

lib/blacklight/solr/repository.rb
maxkadel: @jcoyne So should we use the same pattern for the Debug lines do you think? I was worried that could make it too big of a change, but I do think it would make sense.

jcoyne: Yes, I think this had the same purpose as what you are trying to do here. It's just extra overhead to keep both.

maxkadel: I've replaced the benchmark here, I agree that otherwise they're duplicative (I don't think info-level logging wants the whole Solr query though).

I do think we should keep the `ActiveSupport::Benchmarkable` in the `Blacklight::AbstractRepository` - removing it would be a breaking change, which I try to avoid as much as possible, downstream apps might be using it, and from what I can tell it doesn't cost us much to include it.

I'm going to try to make a "suggestion" of what deprecating it might look like, then you can commit that if that's a blocker for approving this.

lib/blacklight/abstract_repository.rb
```diff
@@ -10,9 +10,11 @@ class AbstractRepository
     attr_writer :connection
 
     # (Note: ActiveSupport::Benchmarkable requires a logger method)
+    # No longer used in core Blacklight, keeping for downstream consumers
     # @return [Logger]
     attr_writer :logger
 
+    # No longer used in core Blacklight, keeping for downstream consumers
```
maxkadel:
```suggestion
    # (Note: ActiveSupport::Benchmarkable requires a logger method)
    # @return [Logger]
    attr_writer :logger

    include ActiveSupport::Benchmarkable

    Blacklight.deprecation.deprecate_methods(
      AbstractRepository,
      benchmark: 'no longer used internally; include ActiveSupport::Benchmarkable in your own Repository subclass if still needed',
      :logger= => 'no longer used internally; define your own accessor in your own Repository subclass if still needed (only existed to support #benchmark)'
    )
```

jcoyne (commented):
I think this is good now. Do you want to squash the 12 commits?

maxkadel:
> I think this is good now. Do you want to squash the 12 commits?

Done! Thanks Justin

jcoyne (approved):

PR #3900, opened 2026-09-07
Pin json < 3.0 to fix Rails < 8.1 CI

maxkadel:
`json` 3.0 dropped the `quirks_mode` kwarg that `ActiveSupport::JSON::Encoding` still passes on Rails < 8.1, raising `ArgumentError: unknown keyword: quirks_mode` on any ActiveRecord JSON serialization — this has been failing every `release-7.x` test job except `rails 8.1.3`.

**Tradeoff:** this is an unconditional gemspec pin, so it also forces `json < 3.0` on apps running Rails 8.1+ that don't need it.

### DIFF
--- a/blacklight.gemspec
+++ b/blacklight.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rails", '>= 6.1', '< 9'
   s.add_dependency "globalid"
   s.add_dependency "jbuilder", '~> 2.7'
+  s.add_dependency "json", '< 3.0' # 3.0 dropped the quirks_mode kwarg ActiveSupport::JSON still passes on Rails < 8.1 - remove once this gemspec requires rails >= 8.1
   s.add_dependency "kaminari", ">= 0.15" # the pagination (page 1,2,3, etc..) of our search results
   s.add_dependency "deprecation"
   s.add_dependency "i18n", '>= 1.7.0' # added named parameters


jcoyne (approved):

jrochkind:
Ah this is just in release-7.x, not 9.x or even 8.x. OK, restricting to old json in that old release seems legit -- I would not like it if it was restricting me in latest BL 9.x.

jcoyne:
@jrochkind there is no released version of Rails (it's fixed i main) that I know of that can use json 3.0. https://github.com/rails/rails/pull/58601

jrochkind:
Good to know, thanks!

In general I don't think it's blakclight gemspec's job to track restrictions that *should* have been in Rails. 

For given Blacklight release X, it's possible that a future Rails release compatible with BL X _will_ work with json 3.0, but the restriction we add to BL X gemspec is forever for that release.  It should be about what BL code is compatible with, not dependencies of dependencies.  

For all other rails apps, people need to add restriction to json < 3 to their app Gemfile themselves, they don't count on some other dependency restricting it for them, right? BL has this thing where it has to get tests to pass, which adds annoyance, but I'd suggest the way to resolve it could be with BL generating that into Gemfiles (where apps will have to put it anyway for now, and which can be changed at any time by the app), or specifying it in a Gemfile.extra only used for CI -- but not to put it in a gemspec requirement that is forever associated with that release. 

But for 7.x release its' fine for sure, that's an old release branch that can be locked to old dependencies forever, I don't mind.

PR #3901, opened 2026-09-08
Forward-port: Pin json < 3.0 to fix Rails < 8.1 CI

maxkadel:
json 3.0 dropped the quirks_mode kwarg that ActiveSupport::JSON::Encoding still passes on Rails < 8.1, raising ArgumentError: unknown keyword: quirks_mode on any ActiveRecord JSON serialization. Same issue/fix as Hyrax's fc79063ac.

### DIFF
--- a/blacklight.gemspec
+++ b/blacklight.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rails", '>= 7.2', '< 9'
   s.add_dependency "globalid"
   s.add_dependency "jbuilder", '~> 2.7'
+  s.add_dependency "json", '< 3.0' # 3.0 dropped the quirks_mode kwarg ActiveSupport::JSON still passes on Rails < 8.1 - remove once this gemspec requires rails >= 8.1
   s.add_dependency "kaminari", ">= 0.15" # the pagination (page 1,2,3, etc..) of our search results
   s.add_dependency "i18n", '>= 1.7.0' # added named parameters
   s.add_dependency "ostruct", '>= 0.3.2'


jrochkind:
Hi!

When Rails releases an 8.1.x that does work with json 3.x, the Blacklight 9.1 release that includes this PR would still restrict upgrading to it in an app using that BL release, even though it would work fine. 

BL would need to release a subsequent BL 9.1.x that allowed json 3 again.  It might end up including other changes that a developer might not want, if they just wanted json 3 with the same version of BL they had. 

Is there another way we could do this that would be preferable?  

Right now every single Rails 9.1 app that uses Activesupport json functions needs to avoid upgrading to json.  I don't think it's blacklight's job to fix this for the local app (if it were, would it be every other Rails-using dependency's job too, should they all lock to json 2 and then re-release unlocked again after Rails releases?)

I think the problem we do need to fix is getting CI to work again -- rather than keeping local apps from upgrading to json 3 that conflicts with the version of Rails they may have, as gemspec does. 

What if we just put this json version limit in a place used by CI (or even generated into installed app), via engine_cart mechanisms, so it doesn't actually go into the released gemspec that can't be changed without another release. Even if it's generate dinto local app's Gemfile, the local app can choose when to remove it and get json 3 with a future Rails 8.1.x that allows it. 

I think engine_cart has several possible ways to do this, I'd need to look into it to refresh my memory.

maxkadel:
@jrochkind  - I had the same concern - closing this one because the axe spec is also flaky and I need to bundle them together to get things green - see https://github.com/projectblacklight/blacklight/pull/3902

PR #3902, opened 2026-09-08
Fix axe spec

maxkadel:
This fixes a flaky spec - the axe spec was running into a race condition with Capybara and turbolinks. 

It also pins json to less than 3.0 via the rails generator and a test-specific Gemfile, rather than via the Gemspec, per @jrochkind astute comment on this other PR - https://github.com/projectblacklight/blacklight/pull/3901#issuecomment-5588477882. Bundling them together because neither is likely to get CI green on its own.

### DIFF
--- a/spec/support/features/turbo_helpers.rb
+++ b/spec/support/features/turbo_helpers.rb
@@ -7,6 +7,7 @@ module Features
     # clicks return before the visit completes, so wait for the busy state to clear
     # before making assertions that don't wait on their own (e.g. be_axe_clean).
     def wait_for_turbo
+      page.has_css?('html[aria-busy]', visible: :all, wait: 0.5)
       expect(page).to have_no_css('html[aria-busy]', visible: :all)
     end
   end
--- a/lib/generators/blacklight/install_generator.rb
+++ b/lib/generators/blacklight/install_generator.rb
@@ -41,6 +41,11 @@ module Blacklight
       generate "blacklight:assets", generated_options unless options[:'skip-assets']
     end
 
+    def pin_json_version
+      # json 3.0's JSON.parse rejects ActiveSupport::JSON.decode's positional options hash; unfixed as of Rails 8.1.3.
+      gem 'json', '< 3.0'
+    end
+
     def bundle_install
       inside destination_root do
         Bundler.with_unbundled_env do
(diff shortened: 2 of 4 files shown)


jrochkind:
Thanks!

I just figured out Gemfile.extra for blacklight_range_limit too. https://github.com/projectblacklight/blacklight_range_limit/pull/337

I'm surprised you need it in the Gemfile.extra _and_ install generator, but if this is intentional that's ok!  This avoids the problem i was worried about before, thanks!

maxkadel:
> I'm surprised you need it in the Gemfile.extra _and_ install generator, but if this is intentional that's ok! This avoids the problem i was worried about before, thanks!

Yes, the install generator puts it into the internal test app, the Gemfile.extra makes sure the testing harness / engine doesn't try to use it.

spec/support/features/turbo_helpers.rb
```diff
@@ -7,6 +7,7 @@ module TurboHelpers
     # clicks return before the visit completes, so wait for the busy state to clear
     # before making assertions that don't wait on their own (e.g. be_axe_clean).
     def wait_for_turbo
+      page.has_css?('html[aria-busy]', visible: :all, wait: 0.5)
```
jcoyne, line 10: Nothing is conditional on this. Is that intentional?

maxkadel: Yes - Capybara might hit the page *before* the turbo has rendered, then the axe spec hits it when turbo *has* rendered and fails it because of the aria-busy. This makes sure that Capybara waits for the aria-busy to show, then waits for it to clear, before passing it on to the axe specs. 

It was flaky before because it was a race condition wrt how quickly the axe specs ran.

jcoyne: What if it has already cleared before this function has been called. Doesn't this stick a 0.5s delay into each test?

maxkadel: It doesn't stick the delay on each test - only if it doesn't have the aria-busy label yet it'll wait up to half a second before it fails the test. Otherwise it will pass immediately and not wait.

jcoyne: It will also wait if the `[aria-busy]` has been set and removed before flow goes into the `wait_for_turbo` method.

maxkadel: Do you have an alternate suggestion? 

Should we just ignore-list that specific axe failure? 

Make it so turbo doesn't use that tag? (I'm not sure that would be appropriate)

jcoyne: I just benchmarked this, it's now ~12s compared to ~8s before, so probably not anything to obsess about.

PR #3910, opened 2026-09-18
Re-enable mounting the app for local development

maxkadel:
Re-enable mounting the app for local development

This no longer causes problems on macs, and is handy for development.


sandbergja (approved):
Nice, thanks @maxkadel ! 🦄

PR #3911, opened 2026-09-18
Allow really long solr values to be truncated

jcoyne:
This prevents thinks like vector search from completly filling the logs


sandbergja (approved):
Thanks, @jcoyne !

PR #3916, opened 2026-09-22
Fix class name in FacetFieldComponent deprecation message

mksndz:
When addressing deprecation warnings in our app, I noticed [a deprecation warning message with a typo](https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/facet_field_component.rb) in the suggested new class to use. The correct new class to use is [`Blacklight::Facets::FieldComponent`](https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/facets/field_component.rb).


sandbergja (approved):
Confirmed, thanks for catching and fixing this, @mksndz !

PR #3580, opened 2025-04-23
7.x fix ci

maxkadel:
- Fixes CI failures on newer versions of Ruby and Rails
- Uses the same format as the main and release-8.x branches to make it easier to create a weekly run of the release-7.x branch

PR #3917, opened 2026-09-23
Fix solr logging for Rails 6.x and propose patch version

maxkadel:
ActiveSupport::IsolatedExecutionState was added in Rails 7, and Blacklight 7 still supports Rails 6. This PR makes the feature safe for Rails 6, and proposes a patch release.


jrochkind:
Do y'all want to add CI testing of Rails 6 back to 7.x branch, if it is meant to be supported still? 

Looks like here is where CI testing of Rails < 7.1 it was removed from Blacklight 7.x branch, way back in April 2025. https://github.com/projectblacklight/blacklight/pull/3580/commits

